PR #7, opened 2019-11-27
Fix up SDL HAL code

ali1234:
Fixes:

* Now works on platforms other than windows.
* Doesn't use 100% CPU all the time.
* No mutex hassles.
* Better aspect ratio handling.

Btw, if you merge this, please don't squash it. There is useful information in the commit messages.

### DIFF
--- a/32blit-sdl/Main.cpp
+++ b/32blit-sdl/Main.cpp
@@ -43,7 +43,6 @@ std::map<int, int> keys = {
 };
 
 SDL_Thread *t_system_timer;
-SDL_mutex *mutex_window_resize;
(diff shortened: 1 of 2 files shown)


ali1234:
Virtual tilt/analog will still be relative to the window so this might make it behave slightly differently. I assume it will be replaced with something better later, anyway.

Gadgetoid:
Brilliant set of commits here, thank you, this fixes up a few of the bugbears that I've been stuck with on the SDL test environment for a while now. Commit detail read and noted, will duly preserve. Cheers.

PR #8, opened 2019-11-28
Run user code on a thread, Joypad support, more...

ali1234:
This allows us to monitor it and detect if it is too slow or hanging.
Inter-thread communication is done using semaphores and there is no
locking.

The timer thread waits with 20 msec timeout on a semaphore telling it
to exit. This is done instead of using bool running so that we can
tell it to exit after the system loop thread, and because it lets us
do the 20 msec delay at the same time. When the wait times out, it
posts to the update semaphore. If it detects this semaphore is not
being decremented, it posts slow/frozen events to the main loop.

The system loop thread blocking waits on the update semaphore and
does one update. Then it posts a redraw event to the main loop
and then blocking waits on the redraw semaphore.

When the main loop receives the redraw event it does the redraw
and then posts the redraw semaphore.

The system loop then continues and waits on the update semaphore
again.

When the main loop receives slow and frozen events it sets the window
title to show the message. This could be drawn in the window instead,
or handled as appropriate.

At shutdown, the system loop posts a semaphore and the main thread
waits with timeout for it. This is so we can detect frozen user
code and detach the thread, allowing clean exit.

### DIFF
--- a/32blit-sdl/Input.cpp
+++ b/32blit-sdl/Input.cpp
@@ -1,0 +1,48 @@
+#include <SDL.h>
+
+#include "Input.hpp"
+#include "engine/input.hpp"
+
+std::map<int, int> Input::keys = {
+	// arrow keys
+	{SDLK_DOWN,   blit::button::DPAD_DOWN},
+	{SDLK_UP,     blit::button::DPAD_UP},
+	{SDLK_LEFT,   blit::button::DPAD_LEFT},
+	{SDLK_RIGHT,  blit::button::DPAD_RIGHT},
+
+	// wasd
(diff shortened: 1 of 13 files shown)


ali1234:
Since this is thread code I advise a lot of testing. :)

ali1234:
Not sure if that will need a mutex.

Gadgetoid:
Excellent stuff- since this is a beta release I'm not averse to moving forward quickly. IE: merging this with minimal testing and letting other beta users feedback on how it works for them.

ali1234:
This is turning into a kitchen sink pull request.

PR #11, opened 2019-11-28
Include missing headers to fix compilation issues under OS X.

voidberg:
Due to some missing includes the engine code does not compile under OS X.

This PR fixes those issues.


Gadgetoid:
Thank you!

PR #25, opened 2019-12-16
CMake Build System

ali1234:
Adds support for building both targets with CMake. To use it, include the `32blit.cmake` and then use `blit_executable` to specify the executable, like this:

```
cmake_minimum_required(VERSION 2.8.9)
project (fire)
include (../../32blit.cmake)
blit_executable (fire fire.cpp)
```
A project can be located anywhere, as long as the path to `32blit.cmake` is correct.

To build it, create a new directory anywhere and run cmake inside it, supplying the path to the project. If you build in a subdirectory of the project it is just like this:

`mkdir build; cd build; cmake ..; make -j8`

For the 32blit repo you can build at the top level to build all examples, or you can build individual examples by specifying the path to the example directory.

If you want to include video capture support:

`cmake .. -DVIDEO_CAPTURE=true`

If you want to build for 32blit hardware supply the relative path to the `32blit.toolchain` at top of repo:

`cmake .. -DCMAKE_TOOLCHAIN_FILE=../../../32blit.toolchain`

Note: when you add source files to your project or change the CMakeLists.txt you need to re-run cmake. But you don't need to resupply the `VIDEO_CAPTURE` or `CMAKE_TOOLCHAIN_FILE` options because cmake remembers them. However, this means that if you want to change those options you have to make a fresh build dir. The best thing to do is make a build dir for every configuration you want to build.

If you open the top level of the repo in CLion you should immediately be able to select an example in the top right and build/run debug it:

image

And same goes for external projects. However, I have not yet figured out how to get CLion to build for ARM.


ali1234:
Note that this breaks the Makefile build because the ifdefs for video capture are reversed to make it easier to handle.

ali1234:
Also it is dependant on CMake being able to find SDL and FFMpeg, which is easy if your are on Linux - just install the distro packages. I have no idea how to make it work on Windows. Same goes for arm compilers. The whole thing works out of the box with distro packages on latest Ubuntu.

Gadgetoid:
Thank you!

Testing this on WSL with cross-compile to win32 executables is on my todo list. It does necessitate first cross-compiling SDL for the 32blit-sdl HAL, but that's relatively straight-forward and easy to automate. Building for win32 requires a glut of additional flags to avoid needing to ship a bunch of library files, but they're not too onerous: https://github.com/pimoroni/32blit-beta/blob/83ff405be76363fe4b1c9d7f692eed287e64d913/32blit-sdl/wsl-build.sh#L22

At the moment I also haven't investigated supporting Windows directly since I've been using WSL to compile/cross compile with the Makefiles and it's quite effective. I also haven't looked into Cygwin but it's my understanding that this should all work, at least with only minor modifications.

I'll aim to get this more thoroughly tested and merged ASAP since the Makefile build system is horribly, irredeemably broken and is more likely to confuse and confound people.

ali1234:
Those static flags should probably be used on Linux too, if you plan to ship an executable. CMake knows about static and dynamic configurations so this would probably be quite easy to set up.

Gadgetoid:
That's true- I've updated the README and merged. The mingw stuff still needs some TLC, but I'd rather the much-better CMake setup was available to everyone than let it languish while I look into things. Thank you again for this. Makes a world of difference!

PR #28, opened 2019-12-28
Fix dfu build command

johnmccabe:
Was using the already existing .hex file as the target output rather than a new .dfu file.


Gadgetoid:
Ooof, good catch, I'd been just running the raw command! Thank you.

PR #30, opened 2019-12-31
A couple of tweaks to enable building examples with SDL on OSX.

ntwyman:
I have only tested this on my mac which is currently running Catalina (10.15.2), but all the examples build and run using clang/llvm (standard Mac toolchain). 

All the code changes should be protected by suitable preprocessor #ifdef's but I have not verified that I didn't accidentally break builds on Linux or Mingw.


README.md
```diff
@@ -104,6 +104,39 @@ cd build.mingw
 cmake .. -DCMAKE_TOOLCHAIN_FILE=../../../mingww.toolchain -DSDL2_DIR=/usr/local/cross-tools/x86_64-w64-mingw32/lib/cmake/SDL2
 ```
 
+### Building & Running on OSX
+
+You will need build tools and CMAKE. Assuming you have [homebrew](https://docs.brew.sh/Installation) installed
+
+``` shell
+xcode-select --install
+brew install clang
```
johnmccabe: There isn't a `clang` formula in homebrew-core, clang is installed by the `llvm` formula - but the xcode-select will do so as well - I assume you've used that rather than brew here?

examples/fire/fire.cpp
```diff
@@ -33,9 +33,13 @@ const uint16_t screen_height = 240;
 
 /* define storage for the framebuffer, spritesheet, and mask */
 //rgb     __fb[screen_width * screen_height] __attribute__((section(".fb")));
+#ifdef __APPLE__
+uint8_t __m[screen_width * screen_height] __attribute__((section("__DATA,.m")));
+uint8_t __pb[screen_width * screen_height] __attribute__((section("__DATA,.fb")));
+#else
```
Gadgetoid: I think having `__APPLE__` specific defines in user code is somewhat messy- the user shouldn't really have to think about a platform they may not even use.

I think we should switch this around to detect a build for `STM32` and apply the appropriate `__attribute__((section(".xx")));` suffix. In all other cases, specifying a section is redundant on Linux, Windows and Mac because they don't have the varying size/performance non-continuous memory regions that the STM32 does.

We could possibly (ab)use `#ifdef __arm__` for this. Does the following alternative build okay for you?

```
#ifdef __arm__
uint8_t __m[screen_width * screen_height] __attribute__((section(".m")));
uint8_t __pb[screen_width * screen_height] __attribute__((section(".fb")));
#else
uint8_t __m[screen_width * screen_height];
uint8_t __pb[screen_width * screen_height];
#endif
```

Gadgetoid:
Thank you for this. Much appreciated!

I've reviewed the code and you should see a comment above.

examples/fire/fire.cpp
Gadgetoid: In fact `#ifdef __arm__` would be a really dirty hack that would not be compatible with other ARM targets. What if I want to build a 32Blit game in the SDL HAL for OpenPandora?

I don't think this compilation unit will see the defines set by the 32Blit STM32 HAL. Maybe we just need to set our own flag within the CMake toolchain file and rely on that?

README.md
ntwyman: Doh - that line was supposed to be `cmake`. Yup  I used xcode-select for the tools

examples/fire/fire.cpp
ntwyman: Totally agree. I have added a `__SECTION__(S)` macro in `32blit/32blit.hpp`.

Does the right thing based on a `TARGET_32BLIT_HW` define set in the `32blit.toolchain` and is a NOP if the toolchain doesn't define that.

Gadgetoid: Ah! What an elegant solution. We were closing on something like this in Discord chat, but not quite as nice. Thank you!

Gadgetoid:
@johnmccabe if you're satisfied that this is good to go, then I'll merge post-haste.

johnmccabe:
Missed this earlier, I'll run through the process on my Mojave based back at lunch to sanity check on the last OS release.

PR #34, opened 2020-01-02
Make tools executable and use LF line endings

johnmccabe:
Makes the tool scripts executable and changes the `dfu` and `README.md` line-endings to `LF`.


Gadgetoid:
Yeah- the whole STM32 codebase is one big CRLF nightmare due to how STM32Cube exports code on Windows. Should probably bite-the-bullet and do a bulk conversion on it all, then enable CRLF->LF conversion in Git.

But for now, this helps! Thank you.

PR #35, opened 2020-01-02
Remove alpha var from RGB getpixel

johnmccabe:
Asset building for an RGB sprite sheet was failing due to attempt to assign non-existent alpha var.

### DIFF
--- a/tools/asset-builder
+++ b/tools/asset-builder
@@ -278,7 +278,7 @@ if __name__ == '__main__':
                         r, g, b, a = source_image.getpixel((x, y))
                         c = color(r, g, b, a)
                     elif source_mode == 'RGB':
-                        r, g, b, a = source_image.getpixel((x, y))
+                        r, g, b = source_image.getpixel((x, y))
                         c = color(r, g, b)
 
                     try:
--- a/tools/sprite-builder
+++ b/tools/sprite-builder
@@ -348,7 +348,7 @@ if __name__ == '__main__':
                         r, g, b, a = source_image.getpixel((x, y))
                         c = color(r, g, b, a)
                     elif source_mode == 'RGB':
-                        r, g, b, a = source_image.getpixel((x, y))
+                        r, g, b = source_image.getpixel((x, y))
                         c = color(r, g, b)
 
                     try:


johnmccabe:
Hold off merging this, it needs another fix, the format is being set to `0x03` for `RGB` rather than `0x02`.

Gadgetoid:
Format `0x03` is paletted but doesn't have an opinion about palette type- the tool is probably building broken assets because the engine's idea of format changed and this didn't catch up.

Right now the tool can't diambiguate between RGB and RGBA palettes, I think it should always throw in a full alpha value and waste those bytes at the moment.

Pixel formats as understood by the engine are:

https://github.com/pimoroni/32blit-beta/blob/64eefb68736bb262fb51fab687d253bbad582ed6/32blit/types/pixel_format.hpp#L7-L13

Right now `load_from_packed` doesn't support anything but paletted assets because an RGB, RGBA or RGB565 asset is really just a regular surface. IE: There's nothing to unpack.

Here's the whole `load_from_packed` function:

https://github.com/pimoroni/32blit-beta/blob/64eefb68736bb262fb51fab687d253bbad582ed6/32blit/graphics/surface.cpp#L523-L580

As you can see from:

https://github.com/pimoroni/32blit-beta/blob/64eefb68736bb262fb51fab687d253bbad582ed6/32blit/graphics/surface.cpp#L525

it's hard-coded to 4-bytes per pixel, or RGBA palettes.

We should probably hard-code the `mode` to `0x03` since this tool will almost always be used to generate paletted, packed images.

Another code branch is probably needed to produce hpp-style output from RGB and RGBA raw pixel data- giving the option to compress into RGB565.

Gadgetoid:
I don't mind merging this as-is and copying the fix across to the new `sprite-builder`, but if you want to claim your bugfix foo in the immutable GitHub commit history then you could update this PR to include the same fix to `sprite-builder`. Let me know!

johnmccabe:
Will update it when I'm finished work later this evening.

johnmccabe:
@Gadgetoid pushed the change into sprite-builder as well (looks like its working well with both RGB and RGBA (tried with 8bit images thus far).

Gadgetoid:
Thank you!

PR #37, opened 2020-01-04
Store compilers in the CMake cache

Daft-Freak:
Fixes this error when using VS Code with the CMake Tools extension:

> The path to the compiler for one or more source files was not found in the CMake cache. If you are using a toolchain file, this probably means that you need to specify the CACHE option when you set your C and/or C++ compiler path

### DIFF
--- a/32blit.toolchain
+++ b/32blit.toolchain
@@ -1,8 +1,8 @@
 set(CMAKE_SYSTEM_NAME Generic)
 set(CMAKE_SYSTEM_PROCESSOR arm)
 
-set(CMAKE_C_COMPILER arm-none-eabi-gcc)
-set(CMAKE_CXX_COMPILER arm-none-eabi-g++)
+set(CMAKE_C_COMPILER arm-none-eabi-gcc CACHE PATH "Path to C compiler")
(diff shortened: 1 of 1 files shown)


32blit.toolchain
```diff
@@ -1,8 +1,8 @@
 set(CMAKE_SYSTEM_NAME Generic)
 set(CMAKE_SYSTEM_PROCESSOR arm)
 
-set(CMAKE_C_COMPILER arm-none-eabi-gcc)
-set(CMAKE_CXX_COMPILER arm-none-eabi-g++)
+set(CMAKE_C_COMPILER arm-none-eabi-gcc CACHE PATH "Path to C compiler")
```
Gadgetoid: You've got me at a loss here (my knowledge of CMake is still pretty limited). Is `"Path to C compiler"` a literal value understood by `CACHE PATH` or is this a value that the user needs to supply (by manually editing this file) before running `cmake ..`?

Daft-Freak: It's just documentation for the variable. I think it's used by the GUI.

Gadgetoid:
Thank you! Tested and merged.

PR #38, opened 2020-01-06
Cdc

AndrewCapon:
Added CDC USB device.
Added bit::debugf() for formatted printing over CDC.
Added 32blitgame code to show working.

Need to check this isn't stepping over the FatFS stuff as I am overriding _write() from stdlib.
Cant do this here as my FatFS won't startup.


Gadgetoid:
Thanks for putting this all into a PR. I'm slowly getting around to reviewing it all.

One issue with builds at the moment is that this PR is missing `Inc/usbd_def.h`:

```
[  1%] Building CXX object 32blit-stm32/CMakeFiles/BlitHalSTM32.dir/Src/main.c.obj
In file included from /32blit-beta/32blit-stm32/Src/main.c:34:0:
/32blit-beta/32blit-stm32/Inc/usb_device.h:33:22: fatal error: usbd_def.h: No such file or directory
```

AndrewCapon:
> Thanks for putting this all into a PR. I'm slowly getting around to reviewing it all.
> 
> One issue with builds at the moment is that this PR is missing `Inc/usbd_def.h`:
> 
> ```
> [  1%] Building CXX object 32blit-stm32/CMakeFiles/BlitHalSTM32.dir/Src/main.c.obj
> In file included from /32blit-beta/32blit-stm32/Src/main.c:34:0:
> /32blit-beta/32blit-stm32/Inc/usb_device.h:33:22: fatal error: usbd_def.h: No such file or directory
> ```

Right, I have been building straight from the Makefile in 32blit-stm32 folder, this Makefile should work fine.

I have not looked at the stuff from CMake, where are you building and what command line and I will check here.

Gadgetoid:
Ah this must be a CMake issue then- I'll give it a poke and see if I can get it to build.

Note: I plan to  deprecate the Makefiles because they're awful and broken in dozens of ways and an unecessary pain to work with alongside the CMake stuff.

Gadgetoid:
I've made a few tweaks before merging this- 

* fixed up the CMakeLists.txt so everything can be compiled with CMake and serial debug used with existing examples/projects.
* moved BlitGame.cpp into examples/serial-demo which can also  be built with CMake.
* moved OpenOCD into tools/ - I think these scripts/configs will need some work to target CMake outputs, maybe CMake could even generate the right config in each project for uploading/debugging? Both are new to me, so I'm not sure what the right approach is yet.

I've been having trouble with Windows 10 opening the CDC Serial device, despite it showing up perfectly fine. This may be a bug in the USB implementation that Windows is sensitive to. I usually run into issues with Windows vs Linux when doing USB stuff.

PuTTY isn't giving me anything more insightful than just:

image


Oh and- once again- thank you for this ace PR!

Gadgetoid:
Fixed the issue with Windows as of 8c47cae823a3d54f9c2b4f688d60d118fae452d4

PR #40, opened 2020-01-07
Require at least CMake 3.1

Daft-Freak:
Just something I noticed while setting up a project.


Gadgetoid:
@ali1234 any input on this? I have 3.10.2 (Ubuntu 18.04), CMake 3.1.0 shipped December 2014 and CMake 2.8.9 shipped August 2012. I don't see any harm merging since your average Linux user keeps up to date, and this may ward off future incompatibilities.

Gadgetoid:
Merging on the basis that this is trivial to revert if it goes awry.

Thank you!

PR #41, opened 2020-01-08
Build DFU files automatically as part of the build

ali1234:
Also fixes the renamed fire directory in examples and adds a new
variable for the size tool instead of hard-coding it.


ali1234:
Note that this will cause STM32 build to fail if DFU dependencies are not installed. SDL builds should not be affected.

Gadgetoid:
Aha thanks for the fix. I should have tested a full CMake build!

On-demand DFU files is a good add- I can now drop that step from the Windows flash script.

Will have to add the utility dependencies to the setup instructions.

mikerr:
The dfu python script adds a whole lot of  dependencies
- Linux & Mac users don't even need it at all, as dfu-util can flash the .bin directly

PR #43, opened 2020-01-08
Emscripten support for 32blit-sdl

Daft-Freak:
Don't know if you actually want this, but was working on SDL today and wondered how hard this would be...

[example](https://cupboard.daftgames.net/32blit/logo.html) (everything that gets built should be there)

### DIFF
--- a/32blit-sdl/CMakeLists.txt
+++ b/32blit-sdl/CMakeLists.txt
@@ -1,5 +1,3 @@
-find_package(SDL2 REQUIRED)
-
 add_library(BlitHalSDL STATIC
 	Input.cpp
 	Main.cpp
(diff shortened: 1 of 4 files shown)


Gadgetoid:
Yes.

Yes!

We had an early prototype of the SDL engine, Lua bindings and interpreter running in Emscripten and coupled with Visual Studio Code's editor. It was glorious. This gets us a step in that direction with the new codebase. Thank you!

Look forward to testing & PRing.

Gadgetoid:
It strikes me that this should probably come with a `.toolchain` file to set the correct compiler, etc for the Emscripten build in much the same way `32blit.toolchain` currently does for the ARM build.

This would make the `cmake` invocation something like `cmake .. -DCMAKE_TOOLCHAIN_FILE=../../../emscripten.toolchain`

Is that something you want to put together? Otherwise I can have a crack at it.

Gadgetoid:
Oh in fact it looks like the `cmake` invocation should just be:

```
cmake .. -DCMAKE_TOOLCHAIN_FILE=$EMSDK/upstream/emscripten/cmake/Modules/Platform/Emscripten.cmake -G "Unix Makefiles"
```

Daft-Freak:
Or even `emcmake cmake ..` for maximum shortness.

PR #44, opened 2020-01-08
Updated README.md

lenardg:
Added some more details based on my setup experience.

- Added link to DFuse tool on st.com
- Added some details about WSL - although the readme did refer to it, I found the reference a little vague
- Changed the package name for the arm compiler to be what you actually need to apt get.


Gadgetoid:
Looks good, thank you! The more the merrier with these- it's really hard to see the wood for the trees when writing instructions.

PR #45, opened 2020-01-08
Added info about XMing and Python requirements

lenardg:
Some more updates to the README.md file to help people get started

- Added info about Python3 and PIP3 and Python modules the user needs. I needed to install these modules into WSL myself.
- Added info about XMing


Gadgetoid:
Thank you!

Worth keeping a note somewhere to bump the XMing setup version if it changes? Or would this even matter? The direct link is super handy either way, but subject to bitrot.

lenardg:
It seems to be a very old version that is not updated any longer. So I guess we could keep it this way? I might add a link to the ”official” page as well.

Issue #47, opened 2020-01-08
Proposal: Split README.md into OS-specific and general setup


Gadgetoid:
Right now our README.md is getting bigger by the minute with all the necessary steps to get working.

While the Wiki is great for FAQs and more esoteric issues I think it might be worth creating a `documentation` folder and having a `WINDOWS.md`, `LINUX.md` and `OSX.md` for noise-free instructions.

Thoughts?

Gadgetoid:
I see @ntwyman @lenardg @LordEidi @johnmccabe and @ali1234 have contributed to the readme so far (which is awesome in and of itself) so your input on this would be the most valuable. Thank you!

lenardg:
This same thought crossed my mind while writing the WSL content. So I am in favor of this!

But I would call the folder just simply `docs` :)

Gadgetoid:
As of https://github.com/pimoroni/32blit-beta/commit/691a68b6f3aaa94f9430a224a5f5c94ce4df4766 I've also poured Emscripten onto this doc-fire, so it's getting out of hand!

LordEidi:
I'd prefer a /doc folder with multiple .md files in it over the wiki for a simple reason. You can download (well git clone) the whole /doc folder and have things locally. Doing so with the wiki is not as easy. And yes I'd go with /docs or /doc vs /documents.

Which also shows my preference regarding a single vs multiple files. Anything we could help there with?

Gadgetoid:
Any suggestions for making the documentation clearer and more complete are always welcome. I think each new person who sets up and runs into issues helps iron on the problems.

Should probably - and I'm leaving this note to myself more than anyone - get some details of the Windows 10 DFU Mode driver issue in there too- whereupon 32blit seems to appear as "STM32 In FS Mode" rather than "STM32 In DFU Mode" and doesn't show up in DfuSe.

PR #48, opened 2020-01-09
Moved documentation bits into the docs folder #47

lenardg:
As a result, README.md became much more simpler.


Gadgetoid:
Thank you! It's all looking much more manageable now.

I wonder if we can list and link the individual docs- I've never been quite clear on how to link to markdown files but apparently GitHub supports relative links (as of forever ago): https://github.blog/2013-01-31-relative-links-in-markup-files/

PR #49, opened 2020-01-09
Changed the window title to something more appropriate

lenardg:
When using the SDL version to run locally, changed the window title to show more dedication to the 32blit :-)


Gadgetoid:
Oh wow, TinyDebug harks back a while! Good catch, thank you!

PR #50, opened 2020-01-09
Visual Studio 2019 support

lenardg:
I added support for Visual Studio 2019 (Visual C++). The examples can be compiled and run (from an .EXE) to test the code on Windows. Nothing is required just VS2019, no mingw or other tools. This can be good for development or testing. (The ARM toolchain and WSL or a Linux VM is still required to build DFU files for the device). 

I needed to do two small code changes for this to compile with VC++. I tried to keep these changes to the minimum possible. My goal was to add VS2019 support on top of everything else.

As a result, most of my work is in the `vs` folder. Here is the solution file and all the C++ project files. I also added a file to the docs folder.

The projects reference the source code from where it is, it is not copied over, neither is the VS support sprinkled into he current files. If you want to "officially" support VS2019 going forward, this method might need revisiting. 

As per the doc file, SDL development libraries need to be downloaded separately.


lenardg:
I should note: one example is not included in the Visual Studio solution, and another one will not run. The not included one had some problems compiling, and the other just failed at runtime. I will look into these later.

lenardg:
I also had to relax VS2019 default security checks on C++ code because of some sprintf calls (in some of the examples).

lenardg:
Also, because I now referened the source files from the C++ projects, it means going forward these projects need to be maintained to have the VS2019 compilation work. It's a bit annoying, but that is how it is now. I will of course try to maintain it going forward :)

Sorry for the multiple comments.

Daft-Freak:
Just a contributor, not the devs but the extra project shouldn't be needed since [VS can open CMake projects](https://docs.microsoft.com/en-us/cpp/build/cmake-projects-in-visual-studio?view=vs-2019). (Or, ignoring that `cmake -G "Visual Studio 16 2019"`).

Gadgetoid:
@Daft-Freak is there anything else to getting cmake to work with SDL2 packages installed by vpkg?

I'm invoking CMake like so:

```
cmake .. -G "Visual Studio 16 2019" -DSDL2_DIR=C:\vcpkg\installed\x86-windows\share\sdl2\
```

And running into:

```
CMake Error at 32blit-sdl/CMakeLists.txt:16 (find_package):
  Could not find a configuration file for package "SDL2" that is compatible
  with requested version "".
```

Gadgetoid:
My bad- I have Win32 only SDL2 installed it appears, and the Generator defaults to x64 arch.

Gadgetoid:
@lenardg looks like your project configuration has the classic catches-me-out-every-single-darn-time issue of being set up correctly for x86 builds but not x64.

Also I think the build depencency heirachy is odd because if I "Build Solution" I'll get 2 succeeded and 21 failed the first time and 21 succeeded the second time. I'd supply a patch for fixing the build dependencies but apparently solution files change *a lot* when you tweak something. Grumble!

Still, I managed to get things to build and run with fewer steps than manually generating a solution using CMake, and your hand-made solution and project files are considerably nicer than the absolute garbage churned out by CMake.

Here's quick comparison of the project heirachy with your lovely hand-tailored filters and projects on the left (nice folder for examples, 32blit has engine/graphics etc) versus the flat disaster output by CMake:

image

I think we have three options moving forward, either:

1. Fix up this PR (x64 support, build dependencies and any other bugs we find), merge and hope people are motivated to maintain it (it shouldn't change that often in future)
2. Try to get CMake to output something nicer that resembles this PR
3. Use a combination of both approaches to automate some of the maintenance

I don't think I'd bet on 2 or 3 in this case, but Visual Studio doesn't half make a mess of its configuration files to the point where they're almost immune to sensible source control.

Looks like we could stand to lose the "Header Files", "Source Files" and "Resource Files", but broadly the TLDR takeaway from this is that I currently prefer the method in this PR to a CMake generated project. I'm happy to hear others' input on this too!

Daft-Freak:
Have you tried File > Open > Folder? That worked immediately for me. It also doesn't require generating a project and just uses the directory structure as-is.

Gadgetoid:
I couldn’t get it to find the vcpkg installed SDL- in retrospect that might be because I’d not installed the x64 version. Will try again tomorrow!

lenardg:
I will submit with the fixes you mention - build dependency and x64 support.

lenardg:
Ok, so now dependencies and configurations should be fixed, and it compiles for both x86 and x64. It has been a while since I have been actively working with shared C++ projects, so excuse my n00b mistakes :) 

Regarding maintaining the solution and project files going forward: I will probably be doing that if this gets merged into the pimoroni/32blit repo. I at least prefer developing this way now that I have this working.

Q-Division:
For what it's worth I've got a version (that needs cleaning up) that is building SDL2 as part of the build in Windows which works I'll see if I can incorperate Daft-Freak's tweak as well given multiple options

Daft-Freak:
Guessing something like:

```cmake
add_subdirectory(sdl2)
set(SDL2_LIBRARIES SDL2)
```

32blit-sdl/Main.cpp
```diff
@@ -124,6 +124,9 @@ void em_loop() {
 }
 #endif
 
+#ifdef __cplusplus
+extern "C"
+#endif
```
Daft-Freak: I think you can avoid this one by linking to SDL2main.

Q-Division:
I'm using ExternalProject_Add but yes the result is similiar

32blit/graphics/surface.hpp
```diff
@@ -5,7 +5,10 @@
 #include <array>
 #include <stdint.h>
 
-#ifdef WIN32
+#ifdef WIN32 
+#define __attribute__(A)
+#endif
+#ifdef _MSC_VER
```
Daft-Freak: Define WIN32 in the VS project instead?

32blit-sdl/Main.cpp
lenardg: Whoops, that was left there by accident, good catch. Thx! :)

Gadgetoid:
@lenardg I suspect @lowfatcode might also appreciate being able to use the VS project and step debugging.

I'm all for merging this, will test later today or tomorrow. Having a VS project doesn't prevent anyone from opening the CMake project directly or generating their own, and is a more comfortable jumping off point for VS users. (our original codebases were all in VS)

Gadgetoid:
Oh I just remembered either last night or this morning I was thinking the `00placeholder.txt` file might just as well be renamed to `README.md` with the same SDL download message and a link. That way at least it serves a purpose and can be viewed via the GitHub web UI.

Probably the best thing about using Visual Studio right now is that it - on Windows - easily gives a console output for stdout alongside the SDL window.

Daft-Freak:
Any complaints if I submit the vcpkg patch anyway and document it as an alternative?

Gadgetoid:
@Daft-Freak no complaints at all- I'm all for options! PR away. Wanted to ask you for the patch actually- didn't want to commit the manually patched file because it lacks attribution.

Daft-Freak:
PR sent, BTW might want to `set_target_properties(... PROPERTIES CXX_EXTENSIONS OFF)` on most of the CMake targets to avoid any non-standard code sneaking in and breaking VS.

Gadgetoid:
Agreed- the VS projects should be C++14. In the interest of moving forward I've merged this PR and we can address other issues as they arise.

Probably a good thing for the ROADMAP is to have a tool that generates blank 32Blit projects and creates the relevant files for Visual Studio against a template.

PR #52, opened 2020-01-10
New option added to spritebuilder

ahnlak:
Added arrayname command line option to allow for the naming of the generated packed_data array; useful for projects that aren't working from a single spritesheet for all their graphics data.

### DIFF
--- a/tools/sprite-builder
+++ b/tools/sprite-builder
@@ -264,6 +264,9 @@ if __name__ == '__main__':
     parser.add_argument('--tilesizex', type=int, help='tile size x', default=8)
     parser.add_argument('--tilesizey', type=int, help='tile size y', default=8)
 
+    # Allow for naming of the resulting packed_data array
+    parser.add_argument('--arrayname', type=str, help='set output array name (default <packed_data>)', default='packed_data')
+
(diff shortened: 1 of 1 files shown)


Gadgetoid:
This should probably be added to `parser_packed` rather than the main parser, since it only applies to packed assets. But otherwise, great idea!

Because argparse is a pain and subparsers trump all, adding to `parser_packed` will change the command format to something like:

```
sprite-builder packed --arrayname something
```

Rather than (what I guess it must be currently):

```
sprite-builder --arrayname something packed
```

ahnlak:
Ok, I'll switch that around (and try and work out if I can modify a PR or have to reissue it!) - I'm not clear what output it expected with non-packed assets as it doesn't seem to do much with raw yet :)

ahnlak:
Aha, it looks like it picks up changes right away (hmm, note to self, use proper branching in future!). And yes, you're right it changes the order of the arguments.

Gadgetoid:
Oh yes indeed- PRs are great for the picking-up-changes immediately, and showing the commits that comprise the PR. They really make this sort of back-and-forth easy!

PR #53, opened 2020-01-10
Handle SDL2 built with CMake

Daft-Freak:
This gets Visual Studio builds working using the CMake integration + vcpkg and theoretically anyone else that builds SDL with CMake instead of autotools...

(I'll attempt to write some docs for using this with VS after the other VS support gets merged.)

### DIFF
--- a/32blit-sdl/CMakeLists.txt
+++ b/32blit-sdl/CMakeLists.txt
@@ -15,9 +15,14 @@ if(EMSCRIPTEN)
 else()
 	find_package(SDL2 REQUIRED)
 
-	target_include_directories(BlitHalSDL
-		PRIVATE	${SDL2_INCLUDE_DIRS}
-	)
+	# If SDL2 was built using CMake, the generated configuration files define SDL2::* targets instead of the SDL2_* variables
+	if(TARGET SDL2::SDL2)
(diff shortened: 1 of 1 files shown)


32blit-sdl/CMakeLists.txt
```diff
@@ -15,9 +15,14 @@ if(EMSCRIPTEN)
 else()
 	find_package(SDL2 REQUIRED)
 
-	target_include_directories(BlitHalSDL
-		PRIVATE	${SDL2_INCLUDE_DIRS}
-	)
+	# if using cmake configuration files SDL_* is not defined but the SDL2::SDL2 target is
```
Daft-Freak: Would be more accurate to say cmake generated configuration files here...

Gadgetoid: Worth changing? Possibly also expanding the comment to say that this specifically addresses Visual Studio? Or does this apply to other scenarios too?

Daft-Freak: It would apply to any system that has SDL built using its CMake build. SDL has two build systems and they generate different (and incompatible) CMake configurations...

ali1234:
I think this fixed building on Arch too, because their SDL package is built with cmake.

PR #55, opened 2020-01-11
Audio Engine

Gadgetoid:
Audio currently clashes with USB support and isn't ready to be merged yet.

Actual sound output sounds like pain- perhaps missing a low-pass filter to take the gritty edge off mixed waveforms?

Audio example needs cleaning up and audio data dumps removed.

Possible other tweaks to rationalise what some of the maybe-not-audio-related changes are in this diff, and commit them to Master via a more descriptive means.

A crude SID-playing example might make a nice additional demo, with appropriate instructions. (Unless anyone wants to write a CPU emulator so it can just run SID files right off the SD card? :D)

Will need a Squash & Merge once complete.


Gadgetoid:
Closing in favour of rebased version on #101

PR #57, opened 2020-01-13
Created an "empty" game for beta testers to base their work upon

lenardg:
I created a new Visual Studio solution and project with source files called game.sln, game.vcxproj and game.cpp and game.hpp. 

Although not a real "template" in the Visual Studio sense, it is still an empty skeleton to base new projects upon. Current beta testers probably do not need this anymore :) But I thought it would be nice to include something like this in the package for new testers joining. I would have liked something like this in the beginning.


dstretch:
I appreciate this, thanks!

Daft-Freak:
BTW, ~equivalent CMake template
```CMake
cmake_minimum_required(VERSION 3.1)
project (game)
include (${32BLIT_PATH}/32blit.cmake)
blit_executable (game game.cpp game.hpp)
```

(Use with `cmake -D32BLIT_PATH=[path to 32blit-beta]` from wherever)

Gadgetoid:
Thank you!

PR #58, opened 2020-01-13
Linked to other docs from the main README

lenardg:
Added links to the main README.md file to point to the other doc files directly using the linking mechanism supported by Github.

Also fixed a small style thingy in the Linux file.

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,7 +22,19 @@ On Windows 10 you can either build for win32 or use XMing and run your builds in
 
 For more information about how to build on the various systems, refer to the platform specific docs in the `docs` folder!
 
-# Overview
+# More docs!
+
+Refer to the OS specific docs:
+
+* [Linux users](docs/Linux.md)
+* [Windows users](docs/Windows.md) or if you are [Visual Studio user](docs/VisualStudio.md)
+* [MacOSX users](docs/MaxOSX.md)
+
+If you want to run the examples in a browser, refer to the [Emscripten docs](docs/Emscripten.md)
+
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -22,7 +22,19 @@ On Windows 10 you can either build for win32 or use XMing and run your builds in
 
 For more information about how to build on the various systems, refer to the platform specific docs in the `docs` folder!
 
-# Overview
+# More docs!
+
+Refer to the OS specific docs:
+
+* [Linux users](docs/Linux.md)
+* [Windows users](docs/Windows.md) or if you are [Visual Studio user](docs/VisualStudio.md)
```
Daft-Freak: Should mention that the first option is WSL/MinGW. Also if you replace 'OS specific' with 'platform specific', then you can list Emscripten and 32blit with the others.

Gadgetoid: I wonder if- faced with good options for using Visual Studio- anyone is actually likely to use WSL/MinGW or WSL/XMing (the latter also doesn't support audio with the current mainline WSL)? We should maybe also move that documentation into a section specifically for Windows uers.

Gadgetoid:
Thank you! Will merge this since the Windows docs questions could probably use some wider discussion elsewhere.

README.md
```diff
@@ -22,7 +22,19 @@ On Windows 10 you can either build for win32 or use XMing and run your builds in
...
+* [MacOSX users](docs/MaxOSX.md)
```
Daft-Freak: A bit late, but there's a typo here: "MaxOSX". (Also I think it's called "macOS" now...)

PR #60, opened 2020-01-13
VS cleanups

Daft-Freak:
Changes for consistency with the CMake build. Mentioned in #50.


Gadgetoid:
Ah! The joys of Visual Studio project files! Thanks for making the tweaks.

PR #61, opened 2020-01-13
Document alternative VS support using CMake+vcpkg

Daft-Freak:
Promised docs from #53. Also adjusted first sentence as it looked like it was saying you could build for the device with VS (at least to me).


Daft-Freak:
This will probably conflict with #57...

Gadgetoid:
Brilliant stuff, thank you!

PR #65, opened 2020-01-15
Add recent examples to the list

Daft-Freak:

Issue #18, opened 2019-12-08
32blit - Namespace pollution


ali1234:
The following engine headers include `using namespace blit;`.

- [ ] types/map.hpp
- [ ] graphics/surface.hpp
- [ ] graphics/tilemap.hpp

This causes every file that includes them to also use the namespace, leading to random name collisions in user code. This filters through to any file that includes `32blit.hpp`.

Note that fixing this will probably break all the examples that don't already contain `using namespace blit;`.

Gadgetoid:
Let it be said that I broadly hate the use of `using namespace blit;` and would probably prefer everything to use the longform, namespace-qualified names. It leaves much less opportunity for name conflicts, or shadowing? Is that even a thing in C++. Although I can't say something if like `colour = blit::colour(0,0,0);` is more or less desirable.

I guess I can't shake that it reminds me of one of my most-hated Python code-smells: `import numpy as np`.

ali1234:
The issue here is using it inside headers - it's fine if you do it inside the .cpp because it will never affect user code.

And yes name shadowing most certainly is a thing. I have a variable called buttons and the IDE complains that it shadows blit::buttons everywhere.

PR #66, opened 2020-01-15
Remove using namepace blit from headers

Daft-Freak:
Got some name collisions in my project, discovered #18, decided to fix it. Tested with #65 applied, so should cover everything.


Gadgetoid:
Thank you!

PR #68, opened 2020-01-16
Apply VS cleanups to template

Daft-Freak:
Or, apply #60 to #57. Not at a Windows machine right now, but `surface.hpp` is probably generating an error about `__attribute__`.


Gadgetoid:
Thanks!

PR #69, opened 2020-01-17
Implement blit::debugf for SDL

Daft-Freak:
Was going to use `SDL_LogMessageV(SDL_LOG_CATEGORY_APPLICATION, SDL_LOG_PRIORITY_DEBUG, psFormatString, args);` but that doesn't return the characters written. So currently both implementations are the same.

### DIFF
--- a/32blit-sdl/System.cpp
+++ b/32blit-sdl/System.cpp
@@ -18,6 +18,13 @@ void debug(std::string message) {
 	std::cout << message << std::endl;
 }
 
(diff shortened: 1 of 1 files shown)


Gadgetoid:
Ooof - yes the null pointer error anyone currently trying to use this in SDL would encounter is not fun!

Thanks!

Daft-Freak:
And then I realise that both are missing a `va_end`...

PR #72, opened 2020-01-17
Call va_end in blit_debug

Daft-Freak:
Let's fix the code I just copy/pasted in #69...


Gadgetoid:
Agreed! Thank you for taking the time to turn this into a PR!

PR #73, opened 2020-01-19
Handle controller connect/disconnect

Daft-Freak:
This fixes controllers in Emscripen builds (not detected until a button is pressed) and also helps with my slightly broken cables...


Gadgetoid:
Thank you!

PR #76, opened 2020-01-23
Update Docs

shane-powell:
Added  a couple paragraphs for troubleshooting gotcha's I stumbled across when setting my environment up.

This should help new beta testers.


Gadgetoid:
Thank you! Actually Jon ran into the SDL issue- which is another thing I think the placeholder file could mention, too.

PR #77, opened 2020-01-24
Add CMakeLists.txt for Empty Visual Studio Game Solution.

shane-powell:
Noticed that the empty game solution for Visual Studio was missing a CMakeLists.txt so have added one.


Gadgetoid:
Oh- good catch! Thank you.

PR #78, opened 2020-01-24
Fix some warnings

Daft-Freak:
Now less than 100!

I was tempted to create a `blit::pi` constant instead of messing with `M_PI`.

(BTW, something I noticed while doing this is that `mat4::rotation` takes degrees and `mat3::rotation` takes radians)

### DIFF
--- a/32blit/32blit.hpp
+++ b/32blit/32blit.hpp
@@ -20,9 +20,8 @@
 #include "types/rect.hpp"
 #include "types/size.hpp"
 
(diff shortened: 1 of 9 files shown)


Gadgetoid:
Ah attack of the unintentional promotion to double-precision :D There seems to be an infinite supply of these, and I just can't seem to form the habit of writing `0.0f` instead of `0.0`.

Thanks for making a PR for this- it's super appreciated.

PR #79, opened 2020-01-24
Misc small changes

Daft-Freak:
Some of this is stuff I mentioned in other PRs, the rest is just things I noticed.

### DIFF
--- a/32blit-stm32/CMakeLists.txt
+++ b/32blit-stm32/CMakeLists.txt
@@ -104,15 +104,15 @@ set(MCU_LINKER_SCRIPT "${CMAKE_CURRENT_LIST_DIR}/${MCU_LINKER_SCRIPT}" PARENT_SC
 set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} -specs=nano.specs -T ${CMAKE_CURRENT_LIST_DIR}/${MCU_LINKER_SCRIPT}" PARENT_SCOPE)
 
 function(blit_executable NAME SOURCES)
-	add_executable(${NAME}.elf $<TARGET_OBJECTS:BlitHalSTM32> ${SOURCES} ${ARGN})
-  	target_link_libraries(${NAME}.elf BlitEngine)
-	set_target_properties(${NAME}.elf PROPERTIES LINK_DEPENDS ${MCU_LINKER_SCRIPT})
-	set_property(TARGET ${NAME}.elf APPEND_STRING PROPERTY LINK_FLAGS " -Wl,-Map=${NAME}.map,--cref")
-  	add_custom_command(TARGET ${NAME}.elf POST_BUILD
+	add_executable(${NAME} $<TARGET_OBJECTS:BlitHalSTM32> ${SOURCES} ${ARGN})
+	target_link_libraries(${NAME} BlitEngine)
+	set_target_properties(${NAME} PROPERTIES LINK_DEPENDS ${MCU_LINKER_SCRIPT} SUFFIX ".elf")
+	set_property(TARGET ${NAME} APPEND_STRING PROPERTY LINK_FLAGS " -Wl,-Map=${NAME}.map,--cref")
+	add_custom_command(TARGET ${NAME} POST_BUILD
 		COMMENT "Building ${NAME}.bin\nBuilding ${NAME}.dfu"
-		COMMAND ${CMAKE_OBJCOPY} -O ihex $<TARGET_FILE:${NAME}.elf> ${NAME}.hex
-		COMMAND ${CMAKE_OBJCOPY} -O binary -S $<TARGET_FILE:${NAME}.elf> ${NAME}.bin
+		COMMAND ${CMAKE_OBJCOPY} -O ihex $<TARGET_FILE:${NAME}> ${NAME}.hex
+		COMMAND ${CMAKE_OBJCOPY} -O binary -S $<TARGET_FILE:${NAME}> ${NAME}.bin
 		COMMAND ${CMAKE_DFU} build --force --out ${NAME}.dfu ${NAME}.bin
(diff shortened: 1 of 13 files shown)


docs/macOS.md
```diff
@@ -23,10 +23,8 @@ sudo make install
 Finally, from the root directory of this repository:
 
 ``` shell
-mkdir build
-cd build
-cmake .. -DCMAKE_TOOLCHAIN_FILE=../darwin.toolchain
-make
+cmake . -B build -DCMAKE_TOOLCHAIN_FILE=./darwin.toolchain
```
Daft-Freak: I'm not sure that a toolchain is actually needed here (at least, I've never needed one before on a mac)

Gadgetoid:
Oh wow! There's a lot to unpack here, but a brief look over everything seems to suggest it's generally all sensible tweaks/improvements. I'm not sure if I'm entirely qualified to test this all, or if I should just hit merge and let nature take it's course (chaotic neutral? :laugh:)

The switch to `cmake . -B build` is an interesting one. I've trained myself to make the build directory first and do the `cmake ..` dance, since it's the pattern I've encountered before with cmake driven projects. This wraps it all up pretty nicely into a one-liner. Just pray nobody does `cmake .. -B build` :laugh:

32blit-stm32/CMakeLists.txt
```diff
@@ -100,19 +100,24 @@ target_compile_definitions(BlitHalSTM32
...
-  	target_link_libraries(${NAME}.elf BlitEngine)
-	set_target_properties(${NAME}.elf PROPERTIES LINK_DEPENDS ${MCU_LINKER_SCRIPT})
-	set_property(TARGET ${NAME}.elf APPEND_STRING PROPERTY LINK_FLAGS " -Wl,-Map=${NAME}.map,--cref")
-  	add_custom_command(TARGET ${NAME}.elf POST_BUILD
+	add_executable(${NAME} $<TARGET_OBJECTS:BlitHalSTM32> ${SOURCES} ${ARGN})
+	target_link_libraries(${NAME} BlitEngine)
+	set_target_properties(${NAME} PROPERTIES LINK_DEPENDS ${MCU_LINKER_SCRIPT} SUFFIX ".elf")
+	set_property(TARGET ${NAME} APPEND_STRING PROPERTY LINK_FLAGS " -Wl,-Map=${NAME}.map,--cref")
+	add_custom_command(TARGET ${NAME} POST_BUILD
 		COMMENT "Building ${NAME}.bin\nBuilding ${NAME}.dfu"
```
ali1234: +1 for getting rid of the .elf prefix. I was meaning to do this some time. This makes it much easier for user to override settings on the target because it always has the same name.

docs/VideoCapture.md
```diff
@@ -13,9 +13,7 @@
 Then configure your 32blit project with:
 
 ```
-mkdir build
-cd build
-cmake .. -DVIDEO_CAPTURE=true
+cmake . -B build -DVIDEO_CAPTURE=true
```
ali1234: Does this work if we already have a build in `build`? If not how do we fix that?

Daft-Freak: It should work the same as before. Looks like I broke the build with VIDEO_CAPTURE in a previous PR though.

32blit-stm32/CMakeLists.txt
Daft-Freak: Yep, exactly what I was thinking.

docs/VideoCapture.md
ali1234: The way it worked before is... it didn't. You can't change variables after you've run cmake the first time. I'm not sure why. It just says "variable was not used".

Daft-Freak: Hmm, think it depends on the variable. This one seems to work:
```
~/programs/32blit-beta$ cmake -B build .
-- The C compiler identification is GNU 9.2.1
-- The CXX compiler identification is GNU 9.2.1
-- Check for working C compiler: /usr/bin/cc
-- Check for working C compiler: /usr/bin/cc -- works
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Detecting C compile features
-- Detecting C compile features - done
-- Check for working CXX compiler: /usr/bin/c++
-- Check for working CXX compiler: /usr/bin/c++ -- works
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Configuring done
-- Generating done
-- Build files have been written to: /home/daftfreak/programs/32blit-beta/build
~/programs/32blit-beta$ cmake -B build -DVIDEO_CAPTURE=1 .
-- Configuring done
-- Generating done
-- Build files have been written to: /home/daftfreak/programs/32blit-beta/build
```

I think you get that error from trying to change something like the toolchain file.

ali1234: Hmm... okay then. This is something that I wish was clearer with cmake, but I don't understand the mechanism well enough. I just delete the whole build when I see those errors.

Gadgetoid: I'm in the "delete the whole build" camp here, too. CMake makes that pretty painless to do, since everything is segued into one directory.

Thanks for your feedback here @ali1234 - I'm still learning the ropes with CMake so someone a bit further down that road vetting these changes is much appreciated.

Gadgetoid:
I have opted to merge this, since we're still in the "move fast and break things" beta phase.

docs/VideoCapture.md
Daft-Freak: You can usually get away with deleting `CMakeCache.txt` and `CMakeFiles/`, but if you're changing something like the toolchain, everything will be getting rebuilt anyway. I'm using the CMake Tools extension in VS Code which handles it automatically.

PR #80, opened 2020-01-25
Update Linux Docs

mikerr:
Add a line to install required tools


Gadgetoid:
Good catch, thank you!

PR #81, opened 2020-01-25
Fix building with video capture

Daft-Freak:
Caused by the Emscripten support, oops.


Gadgetoid:
I may have been contributory to this- might have done some merging... and not enough testing ;)

Thanks for the fix! I think @ali1234 was probably right that nobody is going to use video capture support. I think it'd be great if it snapped short "screenshots" as loopable gifs though. Maybe something for the TODO pile!

PR #84, opened 2020-01-28
Two audio fixes

ali1234:


Gadgetoid:
Ooof- really good catch, thank you! I'm guessing this is the cause of the "farty attack".

PR #85, opened 2020-01-28
A few SDL audio fixes

Daft-Freak:
Stuff I mentioned yesterday and a tiny cleanup. There's still no audio in Chrome without a workaround. (See https://github.com/emscripten-ports/SDL2/issues/57, which I should probably fix in SDL at some point.)


Gadgetoid:
Thank you! This had driven me bonkers. I'm familiar with the issues in Chrome from prior non-emscripten brushes with web audio. We should probably replace the default HTML with a custom one that includes the keyboard controls and an "Unmute" button.

PR #86, opened 2020-01-28
Fix non-high capacity V2 SD cards

Daft-Freak:
This gets the 2GB card I have to work (mostly, I had to change the prescaler).


Gadgetoid:
Thank you- I've encountered one larger (8 or 16GB) card that requires a prescalar change too, but have not tried a small enough card to run into this issue.

Looking over the change the switch from checking for card type to checking if block addressing support is enabled seems sensible enough.

I've chased it through the code, managed to make totally the wrong assumption about what's happening, and then figured out the right one. I think this change can be loosely summed up as "enable byte addressing for SD1 and MMC card types" ? Is that correct?

Daft-Freak:
It's changing the condition from "not SD2" to "not BLOCK". So it's enabling byte addressing for SDv2 without the HCS flag.

PR #87, opened 2020-01-29
Unshorten CMake instructions

Daft-Freak:
-B isn't supported until CMake 3.13, which is newer than the version in Ubuntu 18.04 which is used in WSL. :cry:

### DIFF
--- a/docs/32blit.md
+++ b/docs/32blit.md
@@ -5,9 +5,12 @@ To build your project for 32blit using arm-none-eabi-gcc you should prepare the 
 From the root of your project:
 
 ```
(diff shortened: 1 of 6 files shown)


Gadgetoid:
D'oh!

I believe the `emcmake` command is still valid as a short-hand for including the full path to the toolchain file. I used it in WSL recently and it worked for me.

Daft-Freak:
Ah, yes that can stay.

Daft-Freak:
There, it's now a "mostly revert"

Gadgetoid:
Brilliant! Thank you.

PR #90, opened 2020-01-29
Fix scrolly-tile Emscripten build

Daft-Freak:
There's a conflicting `random` in stdlib.h and an exception is being thrown for attempting to convert a negative float to an unsigned integer.


Gadgetoid:
Thank you! This repository could probably use continuous-integration when we start to move toward a release.

PR #92, opened 2020-01-29
Fix #91, build fails

### LINKED ISSUE #91: STM32 build fails
STM32 build fails (at least on Linux). There is the following error:

```bash
rnentjes@rnentjes-MS-7B79:~/Development/32blit/32blit-beta/examples/palette-cycle/build.stm32$ make
-- Configuring done
-- Generating done
-- Build files have been written to: /home/rnentjes/Development/32blit/32blit-beta/examples/palette-cycle/build.stm32
[ 73%] Built target BlitHalSTM32
[ 98%] Built target BlitEngine
[ 99%] Linking CXX executable palette-cycle.elf
Building palette-cycle.bin
Building palette-cycle.dfu
/usr/bin/objcopy: Unable to recognise the format of the input file `/home/rnentjes/Development/32blit/32blit-beta/examples/palette-cycle/build.stm32/palette-cycle.elf'
make[2]: *** [CMakeFiles/palette-cycle.dir/build.make:236: palette-cycle.elf] Error 1
make[2]: *** Deleting file 'palette-cycle.elf'
make[1]: *** [CMakeFiles/Makefile2:74: CMakeFiles/palette-cycle.dir/all] Error 2
make: *** [Makefile:84: all] Error 2
```

There is a semicolon in an assignment in 32blit-stm32/Makefile:91 which shouldn’t be there. This causes the PREFIX to not be set and then it will pick up the command from the path.

perses-games:
I think this fixes issue #91 where the build fails because the wrong objcopy is picked up.

### DIFF
--- a/32blit-stm32/Makefile
+++ b/32blit-stm32/Makefile
@@ -88,7 +88,7 @@ ASM_SOURCES =  \
 startup_stm32h750xx.s
 
 
-PREFIX := arm-none-eabi-
+PREFIX = arm-none-eabi-
 # The gcc compiler bin path can be either defined in make command via GCC_PATH variable (> make GCC_PATH=xxx)
 # either it can be added to the PATH environment variable.
 ifdef GCC_PATH


perses-games:
Seems like this didn't fix my problem it just looked that way because of some cached files. Never mind.

PR #94, opened 2020-01-30
More Travis!

Daft-Freak:
Should cover the most likely to break builds. Windows is a bit weird, but seems to work (Just don't look at the build output.)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,7 @@ notifications:
 
 language: cpp
 
+os: linux
 dist: bionic
 
 jobs:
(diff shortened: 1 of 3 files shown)


Daft-Freak:
And because I submitted against the travis branch, it's running for the PR.

Gadgetoid:
Nice work! I've been looking into a few things we could do with this automated build:

1. Scrape all the build output and report Errors/Warnings somewhere readable- there's got to be some sort of tool for this?
2. Deploy the various 32blit binaries to a service- would be great if all the examples were just built and zipped up so someone could prep their new SD card with stuff to browse/play with.
3. Deploy the Emscripten builds to a website like the one you set up

This is great stuff for moving out of beta phase, where we have to be more careful about keeping Master stable. Thank you!

PR #96, opened 2020-01-31
More warning cleanup

Daft-Freak:
https://travis-ci.com/Daft-Freak/32blit-beta/builds/146918275

Linux build has no warnings, Emscripten has one (nested comments in platformer), STM32 still has a pile in the HAL code (and the wchar thing) and Windows... is Windows (most of the warnings are about float/int conversions).

(Yes, the branch name is slightly backwards.)


Gadgetoid:
The more warnings the merrier... oh wait :D

Thank you!

PR #95, opened 2020-01-31
External flash

AndrewCapon:
Notes
=====

Sorry, I tried to get the flash-loader going as an example via CMake and failed miserably.
So currently it is being built by the makefile in 32blit-stm32

Overview
========

Consists of three parts:

1. Ability to switch execution between internal and external flash

2. Generalised command handling via CDC streaming.
	There is a single CDCCommandStream hanging off of the CDC receive via a FIFO.
	There can be many CDCCommandHandlers registered with the CDCCommandStream with a FourCC key.
	A CDCCommandHandler can register more that 1 FourCC key.
	The CDCCommandStream parses the incoming data and dispatches a CDCDataStream to the registered CDCCommandHandler

	User code needs placing in a subclass of CDCCommandHandler and implements StreamData() and StreamInit()
	In StreamInit() return true if more data is expected, otherwise false for non streaming commands.
	In StreamData() return a StreamResult, currently implemented:
		srError 	: Somethings gone wrong
		srContinue	: Everything fine but need more data
		srFinish	: Everything fine and finished with stream.

	For examples of simple non streaming commands see CDCInfoHandler and CDCResetHandler.
	For an example of streaming commands see the flash-loader.

3. flash-loader example
	Simple example flash loader
		Display list of bin files on SDCard (just one page), program external flash and run.
		Switch between running loader and app from "System Menu"
		Save bin files to SDCard over CDC
		Program external flash over CDC.

Changes
=======

Changed 32blit.toolchain
	Added variables needed for switching execution and CDC Fifo Buffering

Added 	32blitEXT.toolchain
	Use this toolchain to build for external flash

Changed Engine.hpp and Engine.cpp
	Added code to switch execution

Changed 32blit-stm32/CMakeLists.txt
	Added ccsbcs.c for unicode/ascii
	Added CDC source files
	Added defines for toolchain variables
	Added more Elf info displayed when building

Changed 32blit-stm32/Makefile
	flash-Loader is being built here, sorry I tried looking
	at getting this going via Make but after a couple of hours gave up
	someone that knows about make will need to work this out.

Changed 32blit-stm32/STM32H750VBTx_FLASH.ld
	Increase heap slightly

Added	32blit-stm32/STM32H750VBTxEXT_FLASH.ld
	ld for linking against EXT flash

Changed 32blit.h and 32blit.c
	added blit_switch_executable
	added code to disable ADC, used when transferring data via CDC
	there is an issue in the ADC polling code where it can fail on the second channel read
[...]

PR #98, opened 2020-01-31
External flash (again)

Daft-Freak:
This is #95 with a little extra rebasing and some hopefully reasonable CMake support.


Daft-Freak:
Okay, CMake shouldn't break anything compared to before. (Check the last two commits for details). Also travis is happy: https://travis-ci.com/Daft-Freak/32blit-beta/builds/146981631

Gadgetoid:
I've attempted to wrangle CMake into letting us store firmware projects outside of `32blit-stm32`. See: https://github.com/pimoroni/32blit-beta/commit/3e734ff2291a8f2a1a3a6a7b7b8e2e5ed5a59eb8

Daft-Freak:
Pushed a slightly modified version that avoids duplicating the include dir list or relying on the directory structure. (Though, this is roughly the same as switching the `target_include_directories(BlitHalSTM32 ...` to PUBLIC.)

PR #100, opened 2020-01-31
Update VisualStudio.md

mylogon341:
slight typo


Gadgetoid:
Good catch, thank you!

Daft-Freak:
Uh, the "ore" was wrong, but i think the case was right...

mylogon341:
Yes I noticed this earlier. The case is different from the drop down menu vs the launch screen. My bad.

PR #103, opened 2020-02-01
Add CMakeLists.txt for timer-test

johnmccabe:
Adds the missing `CMakeLists.txt` file for the `timer-test` example.


Gadgetoid:
Well that slipped through the cracks! Good catch. Thank you.

PR #104, opened 2020-02-01
Chrome os doc

shane-powell:
Added a simple document for getting started on Chrome OS.


Gadgetoid:
One day I'll get a Chrome OS device. One day!

Thank you! I don't doubt this will be useful to people.

PR #106, opened 2020-02-01
Improve tga functions

Pharap:
Two simple changes that should help to avoid dynamic allocation and string copying when using the TGA-related functions.

### DIFF
--- a/32blit/engine/utility.cpp
+++ b/32blit/engine/utility.cpp
@@ -20,7 +20,7 @@ namespace utility {
     FILE_READ_ERROR = -3,
   };
 
-  tga tga_header(std::string file) {
+  tga tga_header(const std::string & file) {
     FILE *fp = fopen(file.c_str(), "rb");
 
     tga tga;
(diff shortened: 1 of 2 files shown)


Gadgetoid:
It's with a heavy heart that I must admit these functions are deprecated since we now use packed assets in lieu of images. (Tga is a little obscure to be fair). I'll merge before we kill them off, though, credit where credit is due and I think we can likely learn something from this. Thank you!

Pharap:
> these functions are deprecated since we now use packed assets in lieu of images.

If C++14 features are the deprecated functions could be marked with [the `deprecated` attribute](https://en.cppreference.com/w/cpp/language/attributes/deprecated).
Otherwise a few `// Deprecated` comments might help to prevent others attempting similar changes to other deprecated functions.

Is a large percentage of the publicly visible code deprecated or was I just unlucky here?
If the former is the case I probably won't attempt to make any more suggestions,
but if the latter is the case then I may have a few more suggestions/optimisations to propose.

> Tga is a little obscure to be fair

It might not be as well known as `.png` and the like, but `.tga` files do find a fair bit of use in the game industry.

Notably they're used in Valve's Source Engine, so all Source Engine games use them (Portal, Garry's Mod, Half Life).
Looking at my Steam folder I also have a number of non-Source games that use them,
e.g. Scrap Mechanic, Awesomenauts, Borderlands, Dungeon Defenders, A Hat in Time - not exactly obscure titles (for the most part).

I don't dispute there are probably more suitable formats, I'm merely pointing out that `.tga` is a bit more commonly used than it gets credit for.

PR #109, opened 2020-02-02
Even more Travis

Daft-Freak:
Fixes the VS solution build, then adds it and MinGW to Travis. Also labels the builds.
Screenshot_2020-02-02 Daft-Freak 32blit-beta - Travis CI


Gadgetoid:
Thank you! The labels are a thing of beauty and a joy forever, too!

PR #110, opened 2020-02-02
Flash targets

Daft-Freak:
Not quite finished, but opening a PR to more easily track progress. The current state is that I can run:
```
cmake -B build.stm32 -G Ninja -DCMAKE_TOOLCHAIN_FILE=./32blit.toolchain -DFLASH_PORT=/dev/serial/by-id/usb-STMicroelectronics_32Blit_CDC_305E39803437-if00
ninja -C build.stm32 raycaster.flash
```

And have an example on the device. 

The code shuffling commit is going to be/cause rebase pain for anyone working on the 32Blit tool, sorry. (If anyone wants to merge something, I'll rebase.)

### DIFF
--- a/32blit-stm32/CMakeLists.txt
+++ b/32blit-stm32/CMakeLists.txt
@@ -168,6 +168,10 @@ function(blit_executable NAME SOURCES)
 	set_target_properties(${NAME} PROPERTIES LINK_DEPENDS ${MCU_LINKER_SCRIPT_EXT} SUFFIX ".elf")
 
 	blit_executable_common(${NAME})
+
+	if(32BLIT_TOOL)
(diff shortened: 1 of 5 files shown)


Daft-Freak:
I can also select a target and hit F7 in VS Code, which is very nice.

Daft-Freak:
Okay, this should be good now. You can get from nothing to an example like this: (assuming Linux)
```
mkdir build.stm32
cd build.stm32
cmake .. -DCMAKE_TOOLCHAIN_FILE=../32blit.toolchain
make flash-loader
sudo dfu-util -a 0 -s 0x08000000 -D build/firmware/flash-loader/flash-loader.bin
make logo.flash
```

And then flash another example
```
make raycaster.flash
```

(Hopefully the docs merge makes sense)

Gadgetoid:
Thanks for rebasing this! Truly appreciated.

PR #111, opened 2020-02-03
All types changed to CamelCase. Examples fixed.

lowfatcode:
Sorry Phil.


Daft-Freak:
Hopefully the conflicts from us both fixing the same thing aren't so bad...

Gadgetoid:
Oof from the latest failure it looks like `serial-debug/serial-debug.cpp` isn't in the VS project. Adding this as a reminder to myself to add it!

Gadgetoid:
Same issue again with `text.cpp`. Right now the Visual Studio solution is building, but `text.cpp` is not.

Daft-Freak:
I think I wrote that example before the VS solution was merged.

Gadgetoid:
It's a shame Visual Studio needs a project file for every example- hard to catch if you're on another platform (usually Code and WSL in my case). There's probably some linting we can use to catch that, so we keep on top of maintaining them. Having the VS project is super handy for debugging/cleaning up errors/warnings.

PR #112, opened 2020-02-03
Add dependencies for text VS project

Daft-Freak:


Gadgetoid:
Thank you! Is this related to the failed VS build for #113 ?

Daft-Freak:
Yes, that PR fails because master was failing.

PR #113, opened 2020-02-03
Avoid unwanted bleeping

Daft-Freak:
First one is every 65535ms in scrolly-tile after jumping at least once, second is when audio-test starts.


Gadgetoid:
I made the argument to @lowfatcode that `channel.time_ms += frame_ms;` should *always* happen, otherwise the channels/waveforms will de-sync over time as notes are triggered/stopped. I'm not 100% sure if our current codebase even satisfies this condition. You're clearly right that an overflow will re-trigger the envelope, though, so I've added him as a reviewer to see how we can fix this.

PR #115, opened 2020-02-04
Audio Improvements

Gadgetoid:
This set of changes includes several improvements to the audio engine, and some breaking name changes:

* Now uses a state-machine to handle the envelope generation
* The envelope generator now uses frame timing, rather than trying to work in milliseconds
* attack/decay/sustain/release are converted into frame timings as the state-machine progresses
* The last ADSR value is observed when progressing to the next phase, avoiding pops/clicks
* `AudioVoice` is now `Waveform`

This should fix the issue raised in #113 in addition to preventing any unwanted, jarring transitions between envelope states.

### DIFF
--- a/32blit/engine/audio.cpp
+++ b/32blit/engine/audio.cpp
@@ -9,7 +9,6 @@
 namespace blit {
 
   uint32_t sample_rate = 22050;
-  uint32_t frame_ms = (1000 << 16) / 22050;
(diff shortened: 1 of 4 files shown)


Daft-Freak:
The bleeping is gone, but audio-test is sounding a bit distorted.

PR #116, opened 2020-02-04
Delete unused variables to avoid wasting flash space

Daft-Freak:
Fixes the build if targeting the internal flash


Gadgetoid:
I'm genuinely surprised how many of these I'd missed! Thanks for tidying this up.

PR #117, opened 2020-02-04
Text improvements 2

Daft-Freak:
Follow up to #75. I tried to make the clip rect default to something sensible. Also fixes some wrapping bugs.

Also indirectly fixes flash-loader blowing up when you have too many files.


Gadgetoid:
Brilliant, thank you!

PR #120, opened 2020-02-05
Fix macOS build of 32Blit tool

trollied:


tools/src/32Blit.cpp
```diff
@@ -299,7 +299,7 @@ std::string GuessPortName()
     return "";
 
   io_object_t portService;
-  char devicePath[1024]{};
+  char devicePath[1024];
```
Daft-Freak: `= {}`? or at least zero it somehow, otherwise the return value will be uninitialised when no port is found.

PR #121, opened 2020-02-05
Arbitrary waveform implementation and example.

tinwhisker:
Allows waveforms to be pumped in.


Daft-Freak:
Probably want to add the example to the list in examples/CMakeLists.txt

32blit/engine/audio.cpp
```diff
@@ -94,6 +94,16 @@ namespace blit {
...
         }
 
+        if(channel.waveforms & Waveform::WAVE) {
+
+          channel.waveform_offset = channel.wave_buf_pos;
+          channel_sample += channel.wave_buffer[channel.wave_buf_pos] << 8;
+          if (channel.wave_buf_pos++ == 64) { //
+            channel.wave_buf_pos = 0;
+            (*channel.callback_waveBufferRefresh)();
+          }
```
Daft-Freak: `waveform_count++;` to avoid / 0 fun.

Gadgetoid:
Thanks for putting this together! I see you'd spotted the user waveform type and ran with it :D

We might tweak this slightly to add a half-complete callback or shuffle things around, but it looks like a solid start!

I'm still not totally sure that dividing the output by the number of active waveforms is even valid, but it certainly sounds good for SID emulation.

PR #122, opened 2020-02-05
Added missing ../ to flashloader cmake.

shane-powell:


Daft-Freak:
That's... one level outside the 32blit repo? Are you running inside the firmware folder? I meant for that to be run at the root.

Gadgetoid:
I think the docs are missing the step for building the Flash loader tool (for Windows/Linux/Mac)

There might be a clearer way to say that the build should be run in the root- but then I suspect it's also possible for CMake to find these directories no matter where you run the build from? I'll have a tinker!

Daft-Freak:
I put a "Run the build for your platform in the top level" to avoid duplicating the instructions for every platform.

Gadgetoid:
I think the docs on the whole differ between "build in root" and "build in subfolder" so we probably need to canonicalise on the latter so people don't get into bad habits.

Stuff like the "32Blit" programmer almost feels like it belongs completely outside of this source tree- since it's not valid to build it with anything but a platform specific toolchain, but you'd only be using it when building with the device toolchain. I wonder if it's feasible to stick it into another repository and just ship binaries here?

That said- with all the attention on this repository it's the best place for it to be maintained!

Daft-Freak:
Maybe split out all the tools into their own repo when things are more stable. You can get CMake to find the tool anywhere if it's in your PATH.

Something else that should probably be covered in the docs is building an example vs building your own project, which should probably happen outside the repo entirely. I'm using something like the CMakeLists in vs/game and then doing `cmake -D32BLIT_PATH=path/to/this repo` for my stuff.

I'm going a bit off topic now though. Might go do adjust some docs...

shane-powell:
> That's... one level outside the 32blit repo? Are you running inside the firmware folder? I meant for that to be run at the root.

I was just thinking if people are following the other build examples they would probably go to the files with the cpp file and make a build folder from there.

shane-powell:
As aside note feel free to chuck any documentation my way to test. Should give you a more realistic idea of how its likely to be interpreted as you guys clearly are much more knowledgeable.

Gadgetoid:
I've started reshuffling some of the 32Blit documentation to hopefully isolate the different stages and make them clearer: https://github.com/pimoroni/32blit-beta/commit/f5ef27c75a3ca5d8dc45f7720fe926d0699d394c

Your input on these changes would be greatly appreciated!

PR #123, opened 2020-02-05
macOS travis

Daft-Freak:
Might as well add the last one. I needed the first commit to build locally as well, it's a bit of a mess though.


Daft-Freak:
Emscripten is randomly failing to download SDL...

Gadgetoid:
Re-ran and it's fine. Weird! Glitch in the matrix, perhaps. Thank you!

PR #126, opened 2020-02-05
mac doesnt have wget

mylogon341:
macos doesn't have wget


Gadgetoid:
Thank you! I haven't got a Mac set up for testing in-house, so your help working out the wrinkles in the documentation is appreciated!

PR #127, opened 2020-02-05
path change

mylogon341:
make flash-loader creates path so either we need to cd into it or prefix the path with the directory

### DIFF
--- a/docs/32blit.md
+++ b/docs/32blit.md
@@ -78,7 +78,7 @@ make flash-loader
 Install `dfu-util` from your package manager then enter:
 
 ```
-sudo dfu-util -a 0 -s 0x08000000 -D flash-loader.bin
+sudo dfu-util -a 0 -s 0x08000000 -D flash-loader/flash-loader.bin
 ```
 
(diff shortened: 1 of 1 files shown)


Gadgetoid:
This is a step in the right direction- thanks! 

I think this relates to #122 in that we have some confusion about where a build should be triggered from. I may have spread my (not necessarily bad) habit of compiling from individual example directories via the documentation.

I'm told it's better to compile from the root, since it shares the HAL (32blit-stm32) and API (32blit) across all builds.

Gadgetoid:
Okay I'm doing some significant re-writing of the docs off the back of this. Stay tuned!

mylogon341:
Happy to help out :)

Gadgetoid:
I've started reshuffling some of the 32Blit documentation to hopefully isolate the different stages and make them clearer: https://github.com/pimoroni/32blit-beta/commit/f5ef27c75a3ca5d8dc45f7720fe926d0699d394c

Your input on these changes would be greatly appreciated!

PR #128, opened 2020-02-06
Blue LED is pin 5 of port B, not port C

ali1234:
Pin 5 of port C is the joystick button and setting no pull-up on it
causes it to float.


ali1234:
(Forced rebase on master.)

Gadgetoid:
Oof, I must have missed the GPIO code being refactored. That'll do it. Good catch, thank you.

PR #129, opened 2020-02-06
Add CMake gitignore.

shane-powell:
Adds the github provided CMake ignore list to the bottom of .gitignore.


Gadgetoid:
Pretty sure all `cmake` output should be confined to its `build.*` directories, but it can't hurt to have these anyway to guard against this stuff creeping in! Thank you!

PR #132, opened 2020-02-06
Fix macOS SDL find fallback

Daft-Freak:
Should allow find_package to work properly when the framework isn't used.


Gadgetoid:
Just going to leave one of those drinking birds on my enter key to merge your PRs!

Thanks again!

PR #133, opened 2020-02-06
Tidy up links to OS-specific documentation

Ozzard:


Gadgetoid:
Apparently Markdown's link syntax baffles me too much I decided to just use ] everywhere?

Thanks for spotting and fixing this!

PR #135, opened 2020-02-06
Spelling in flash tool

shane-powell:


Gadgetoid:
Good catch, thank you!

PR #136, opened 2020-02-06
Fix reconnecting with AUTO port

Daft-Freak:
It would try to reconnect to AUTO instead of the detected port...


Gadgetoid:
Whoops! Thank you. I hadn't been using the debugging so I didn't notice!

PR #138, opened 2020-02-07
some python3 install instructions 

mylogon341:
...with pip3 dependency list missing as i don't know what they are 🤷‍♂️

I added a sentence in there about checking if you already have it installed as (i think) it's fairly common for people to already have it, but possibly through some other method (gem?), so ideally this will reduce the issue of having a conflicting install

### DIFF
--- a/docs/macOS.md
+++ b/docs/macOS.md
@@ -7,9 +7,45 @@ xcode-select --install
 brew install cmake
 ```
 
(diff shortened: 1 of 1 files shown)


Gadgetoid:
Thank you , I haven't got my hands on a Mac to try the instructions myself so any improvements/testing is much appreciated!

PR #140, opened 2020-02-07
mac shortcuts

mylogon341:
added /cmd to idiot-proof for us mac types


Gadgetoid:
Good spot! Thank you both!

Issue #139, opened 2020-02-07
VSCode instructions not working


mylogon341:
@Daft-Freak Probably one for you.
I'm following the `VSCode.md` file and got to change the configuration provider and the list is empty
[Seen here](https://imgur.com/a/j4jcIIv)
As you can see I do have both plugins installed, the list is empty, but I did eventually have a notification to 'configure the project'.
Any troubleshooting suggestions you might have?

Daft-Freak:
It looks like you need to configure the build before that option is available. There's also a "CMake Tools would like to configure IntelliSense for this folder." notification that usually pops up after configuring, which does the same thing as changing the configuration provider manually.

Daft-Freak:
Do the updated docs help?

Gadgetoid:
It's worth noting that these instructions are not easily transferable to Windows since, well, Windows is Windows.

It actually gets as far as preparing the build with my locally installed Windows CMake, but understandably fails on finding SDL2 libraries. I'll investigate passing the required argument to `cmake.EXE` and see if I can make it work.

mylogon341:
Yep. I got a little further. How do I populate the `launch.json` file? It's complaining (understandably) that the default values are not configured.
<img width="1035" alt="Screenshot 2020-02-08 at 17 17 44" src="https://user-images.githubusercontent.com/5493303/74089239-30f37480-4a97-11ea-9b2a-64ca367c6702.png">

mylogon341:
Ok. I got it "working" with hard coded value `"${workspaceFolder}/build/examples/raycaster/raycaster",`?
Does anyone know if there is an environment variable so i don't need to put in the project name there.

Daft-Freak:
@mylogon341 If you use Ctrl+F5 or the little debug button at the bottom, you don't need a launch.json. Though, according to [this](https://vector-of-bool.github.io/docs/vscode-cmake-tools/debugging.html#debugging-with-cmake-tools-and-launch-json) you should be able to use `${command:cmake.launchTargetPath}`. 

@Gadgetoid I knew windows would be a pain, looking at ways to get it to find SDL...

mylogon341:
It wouldn't stop referring to the contents of the launch.json file, even if I deleted it. Replacing it with that variable has done the trick, by the looks of it 👍
It might be worth adding that to the docs. I assume its a vsc generated file so can't check it in

Gadgetoid:
[...]

PR #143, opened 2020-02-07
Expand VS Code docs

Daft-Freak:
An initial configure is needed for the IntelliSense setup, should help #139. Also adds some instructions for building and running.

(I checked and debug is `Ctrl` + `F5` even on macOS.)


Gadgetoid:
Still need to give these a try! Thank you.

PR #146, opened 2020-02-07
Update and move game template

Daft-Freak:
It has a CMakeLists and should be a good base for out-of-tree projects. The VS solution is a pain to move however (You need to adjust the project paths in the sln and the include/library paths in the vcxproj). Still working on those docs...


Gadgetoid:
Thank you! Visual Studio is wholly terrible in this respect- hand-editing the config files is just painful. I think it's possible to create a template in its native format, but I haven't played with that since waaay back in early 32Blit pre-alpha days.

PR #147, opened 2020-02-07
Add the loader tool to the VS solution

Daft-Freak:
Since I was on Windows anyway... The project is called "32Blit-loader" as there is already a project called "32blit" but the output is still "32Blit.exe".


Gadgetoid:
This was on my TODO list, thank you!

PR #148, opened 2020-02-07
Use snprintf

Daft-Freak:
Another "I was on Windows" PR. Gives Visual Studio one less thing to complain about and is probably a good idea anyway.


Gadgetoid:
I think some VS Projects have a define to supress this, but probably not all. Actually fixing it is the sensible approach! Thank you.

PR #149, opened 2020-02-07
Fix audio-wave example sample rate

tinwhisker:
'cos I can't read and the dog can't count.


Daft-Freak:
I was going to say this needed some tidying, but went and did it instead: https://github.com/Daft-Freak/32blit-beta/commit/2915a0fedb0965678cbf582281930c875e191120

Gadgetoid:
`Pen` isn't in master yet, but I can leave this open until `graphics-rework` is merged and it should, in theory, build and merge fine.

PR #150, opened 2020-02-07
graphics-rework: update remaining example

Daft-Freak:
And make travis pass


Gadgetoid:
I'm at a loss for how I missed this! I was running `make all` from the project root. Thank you!

Daft-Freak:
It seem to only be built with VS at the moment...

PR #153, opened 2020-02-07
Misc docs updates

Daft-Freak:
Probably not done with this, but it's getting late and I'll probably start writing complete nonsense soon... Also, opening this to hopefully avoid anyone changing the same things.


Gadgetoid:
Ah! Good show. You caught a lot of the little straggly bits I'd missed, too. Thank you!

PR #154, opened 2020-02-08
32 blit sender fix

AndrewCapon:
Fix the 32Blit SAVE problems, this seems to be to do with the app closeing before all USB data has been sent by the OS, simple fix to sleep a second after sending sorts it out


Daft-Freak:
Strange, would have expected the OS to handle that. Especially since we're calling tcdrain.

AndrewCapon:
My thoughts exactly.

Daft-Freak:
Is there an easy way to reproduce this? I haven't had it happen yet.

AndrewCapon:
On OSX here it happens 100% of the time on SAVE and 100% of the time on PROG without the reconnect flag.

AndrewCapon:
I noticed that if I added a breakpoint on the final file close in the firmware then it always worked, if not I had data missing at the end of the stream. So I added the sleep and voila 100% working.

Gadgetoid:
Thanks for figuring this out- I hadn't run into the issue at all here on Windows, despite spamming a whole folder full of files onto my device. I think this is one of those "merge it for now, and hopefully we'll understand the problem better later for a more idiomatic fix". I'll add a comment explaining why this delay is here.

Daft-Freak:
That might have been a good time to use the "rebase and merge" button...

Gadgetoid:
Oooof, those merge commits. WHAT HAVE I DOOONNEE!

AndrewCapon:
Am I doing this git stuff wrong?

I used on a new clone of my fork:
```
 git remote add upstream https://github.com/pimoroni/32blit-beta.git
git fetch upstream
git checkout master
git merge upstream/master
git checkout -B "32BlitSenderFix"
```
make changes, then:
```
git commit -m "blah blah"
git push
```

Daft-Freak:
Replace the `git merge master` with a `git reset --hard upstream/master` and your master will be replaced with upstream (throwing away any changes), or if you have never touched master on your fork `git pull upstream` should also work.

AndrewCapon:
I never touch Master now :)

Daft-Freak:
Good idea, avoids a lot of pain. :smile:

PR #158, opened 2020-02-08
Added a troubleshooting tip as it caused me a head scratch :/

mylogon341:


Gadgetoid:
Ah good addition, it's these little things that we take for granted and don't document properly. Thank you.

mylogon341:
always happy to help :)

PR #159, opened 2020-02-08
Windows SDL cmake config improvements

Daft-Freak:
This removes the need to rename folders for the x86 build and also makes it usable in VS Code with Ninja. Should help with windows problems from #139.


Gadgetoid:
Will give this a test this morning hopefully. Thank you!

PR #160, opened 2020-02-08
Further VS Code docs

Daft-Freak:
More work for #139. Turns out writing documentation is hard. (Someone give me some code to write... please?).

Anyway, this requires #159 for Windows to work. I haven't tested if the arm toolchain file works with an appropriate (windows) toolchain in the PATH, but it does work if you use the "Remote - WSL" extension (which is to be expected since it's effectively running on Linux then...)


Daft-Freak:
Wait, I'm telling the user to edit the config before opening the repo... :man_facepalming:

mylogon341:
>  Turns out writing documentation is hard. (Someone give me some code to write... please?).
NEVER!!! 😂

Gadgetoid:
The Remote WSL estension is also known locally to me as "that bloody extension that breaks all my expectations when I open my Windows filesystem path in VSCode from WSL and aaargh"

But, actually, good point. Using the extension and WSL should make all of the CMake stuff a lot easier.

Thanks for doing the legwork on these docs- documentation is *definitely* not easy. A good programmer leaves it until the very end of their project- which works out well because middle-damagement always add stuff to the project continuously so the programmers never have to write any documentation.

PR #161, opened 2020-02-08
Update 32blit.md

mylogon341:
I just went over this again and realised it reads like the second sentence is referring to the first. Which it isn't. My bad.


Gadgetoid:
Thank you! And no worries- the great thing about a "beta" repo is that we can just say "It's a beta" whenever something like this crops up.

PR #162, opened 2020-02-09
Update Tools.md

mylogon341:
I might be wrong with pillow, but for me it kicked off with `ERROR: No matching distribution found for PIL` which was resolved with installing pillow. 
In my very limited experience with python though, I seem to remember pillow being deprecated.. ? Either way, this fixed my error.


ali1234:
PIL is the one that is deprecated. Pillow provides a PIL package for backwards compatibility.

Gadgetoid:
Good catch! Thank you.

PR #163, opened 2020-02-09
Profiling

AndrewCapon:
Profiler module

A Profiler can have multiple ProfilerProbes

A ProfilerProbe has a set of metrics measures in us: Min Value, Max Value, Current value, Average value

The average value is calculated from a running average which can be set in the ProfilerProbe constructor along with the span.

Values can be logged to CDC via LogProbes().

Values can be displayed as an overlay when called at the end of Render() using DisplayProbeOverlay()

For examples of use and setup please see the profiler-test example:

Example showing how to use the profiler

Button			Function
=====================================================
A					Enable/Disable Bar Graphs
B					Enable/Disable Text Values
X					Enable/Disable History Graph
Y					Switch graph time between global, set by maximum render time
					and individual times, set by each probles maximum time
UP					Reduce rows displayed on page
DOWN				Increase rows displayed on page
LEFT				Back Page
RIGHT			        Next Page
HOME                           Log metrics via CDC


Gadgetoid:
@AndrewCapon knocks it out of the park again. Thank you! Oh Deacon-of-Debugging, Patron-saint of Profiling, Lord of Logging, S... you get the point!

This feature is something we discussed a loooooonnng time ago in the 32Blit formulative stages, and we did have some crude profile graphs in the very early Lua-running Emscripten web UI. It's great to see it back in full force and usable for C++ profiling. We will probably also be able to revive the Emscripten web graph functionality again using this as a basis.

I think this is probably one of those times where a Squash & Merge makes sense - @Daft-Freak would you agree?

AndrewCapon:
On thing that is still missing is the profiling working from anything but the STM32.

There are three functions in system.cpp that need implementing for the other platforms: EnableUsTimer(), GetUsTimer() and GetMaxUsTimer(). I have code here that will work for WIN32/Linux/OSX,  but I was unsure about Emscription as I know nothing about it!

I can add that code in after the STM32 version is merged, or would you prefer it before?

Daft-Freak:
Hmm,  I'm seeing double commits so, yeah sqash. Probably want to add a commit that un-comments out flight/flash-loader first. I might get SDL working if no-one beats me to it...

Daft-Freak:
...And we commented at the same time. Theoretically, Linux timing code should work in emscripten. I was probably going to use SDLs performance counter functions to avoid anything platform specific.

AndrewCapon:
Ah, SDL thingy sounds like a better idea, I know nothing about that either!

All it needs is a timer in us in a uint32_t, and the maximum us that can be used without wrapping in a uint32_t.

Daft-Freak:
Okay, SDL performance counter: https://github.com/AndrewCapon/32blit-beta/compare/Profiling...Daft-Freak:AndrewCapon-Profiling

I put the max us as UINT32_MAX since the performance counter is a 64bit value. Also, I noticed that `EnableUsTimer` isn't called anywhere.

Screenshot from 2020-02-09 17-05-59

(You'll only get 1ms resolution in Emscripten unless you fiddle with browser settings and have an Emscripten patch I haven't even committed yet...)

AndrewCapon:
Nice job. 

The graphs do look a  little off maybe unless the example has been changed, I would expect something like:

IMG_0071


The text values will be different but the historical data graph should look pretty much the same I would have thought?

Also I don't understand that if you have a 1ms resolution how can we have timings displayed that are not divisible by 1000, so 0 ms= 0, 1ms =1000 etc?

EnableUsTimer() should be called in the Profiler constructor, not sure how I missed that!

Daft-Freak:
The timings in the browser are extremely noisy, this is what a local build looks like:

Screenshot from 2020-02-09 18-37-32

The Emscripten screenshot was taken with the timer precision increased, without that it looks like this:
Screenshot from 2020-02-09 18-40-32

Which is... not so helpful.

AndrewCapon:
AH, got you.

I was  a bit confused that the data was half size but it's because it is running at twice the frame rate, doh!

AndrewCapon:
Hi DF,

I have checked and you seem you have all the profiling code in your branch, I am going to delete my repository so can you put a pull request in from your branch into the pimoroni guys for me.

Cheers

Andy

Daft-Freak:
Okay, #168.

Gadgetoid:
Awesome stuff, thank you both. I will close this PR and refer to #168

PR #165, opened 2020-02-09
Fix off-by-one in audio wave handling

Daft-Freak:


Gadgetoid:
Good catch, thank you!

PR #166, opened 2020-02-09
Find python and use it to execute the dfu script

Daft-Freak:
This is pretty much the only thing preventing doing an arm build on windows. The PythonInterp module is deprecated, but the replacement isn't available until CMake 3.12.


Gadgetoid:
Ooh, interesting! I shall get the ARM GCC toolchain installed to Windows and see what happens. I've got some not-32blit-things to attend to first however, but thanks for this!

Daft-Freak:
The only other thing you should need to do is download make/ninja. (Since apparently the one bundled with VS is only detected if using a VS generator...)

Gadgetoid:
I tried a build with GnuWin32 make.exe (not including the changes in this PR) and it wasn't a happy bunny. As near as I can tell the compile was failing to find input paths, presumably because it was generating Unix Makefiles and attempting to run in cmd.exe. Will merge this and try ninja instead.

PR #167, opened 2020-02-09
Implement open/read/close_file for SDL

Daft-Freak:
I was attempting to stream audio from the SD card, and this was helpful for debugging.


Gadgetoid:
Thank you!

PR #168, opened 2020-02-10
Profiling+

Daft-Freak:
See #163 for details. This just adds SDL support on top, plus a few tidy-ups. (Also rebased again as I was using it earlier and needed some of my recent patches.)

I did notice while using this that there is a bit of a consistency issue that might need sorted at some point (snake_case vs CamelCase method names).


Gadgetoid:
Thanks for carrying the torch with this one.

I've been having consistency issues with my own code, so I think we'll probably have to pick a style and set up C++ linting to keep everything in check.

PR #174, opened 2020-02-11
Wrap fonts in a struct

Daft-Freak:
Replaces `&minimal_font[0][0]` with `blit::minimal_font`, which is much nicer to type. Also a step towards different size fonts. I think I've covered everything from #21 now.


Gadgetoid:
This particular wrinkle had been irritating me for months! Thanks again :D

PR #178, opened 2020-02-11
Improve VS "Open Folder" build

Daft-Freak:
It now uses the same SDL as the solution (and using VS Code). It also adds the Debug/Release, x86/x64 build configurations as default.

### DIFF
--- a/CMakeSettings.json
+++ b/CMakeSettings.json
@@ -1,0 +1,76 @@
+﻿{
+  "configurations": [
+    {
+      "name": "x64-Debug",
+      "generator": "Ninja",
+      "configurationType": "Debug",
+      "inheritEnvironments": [ "msvc_x64_x64" ],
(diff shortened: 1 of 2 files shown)


Gadgetoid:
Thoughts on just dropping SDL2 into the folder in this repository? I guess it's a little weird to distribute platform-specific binaries when there's a VS-compatible source project for SDL2.

The license (zlib) would seem to permit this if we went that route.

At the very least, since https://github.com/pimoroni/32blit-beta/blob/master/vs/sdl/00placeholder.txt is now redundant I'm going to rename it to "README.md" and re-iterate the install steps for clarity.

Daft-Freak:
Not a big fan of storing binaries in the repo, especially when this is possibly the easiest to install (download zip, drop files in folder, done.)

Gadgetoid:
Yeah I can't say I am either- it's unweildy and sort of counter the whole concept of git. I've updated the docs.

PR #181, opened 2020-02-12
Rebuild stale assets, move to 32blit.cmake to avoid duplication

Gadgetoid:
This change combines the asset build functions from SDL and STM32 into a single function in 32blit.cmake.

Additionally the library target is renamed to `asset_${ASSET}` to avoid a circular dependency being triggered when the custom command depends upon `${ASSET}`.

### DIFF
--- a/32blit-sdl/CMakeLists.txt
+++ b/32blit-sdl/CMakeLists.txt
@@ -79,32 +79,6 @@ if(DEFINED VIDEO_CAPTURE AND VIDEO_CAPTURE)
 	)
 endif()
 
-function(blit_asset NAME ASSET)
-	add_custom_command(OUTPUT ${ASSET}.o
-		COMMAND cd ${CMAKE_CURRENT_SOURCE_DIR} && ${CMAKE_LINKER} -r -b binary -o ${CMAKE_CURRENT_BINARY_DIR}/${ASSET}.o ${ASSET}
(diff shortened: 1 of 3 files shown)


Daft-Freak:
I've got a bit of a simplification here: https://github.com/Daft-Freak/32blit-beta/commit/f6bbb7d88a45a1d2cc54cda570d498d4dbcfdfd6. Basically it adds the asset as a source of the original target instead of creating a new library. (Still doesn't work on non-gcc though)

Gadgetoid:
Oh wow. That is, indeed, a simplification and I can see exactly why it works. I had a suspicion there was a simpler way, but I just didn't have enough grasp of CMake to find it. Totally happy to use your setup over mine.

Is the `SET_SOURCE_FILES_PROPERTIES` not needed in this approach?

Daft-Freak:
I think `GENERATED` is automatically added if the file is an `OUTPUT`, `EXTERNAL_OBJECT` is probably not needed due passing the full path. (At least, it builds...)

Daft-Freak:
I've updated my branch: https://github.com/Daft-Freak/32blit-beta/commits/blit-link-assets. Surprisingly the Emscripten build doesn't fail (it does warn that the object file is invalid, which is not an error apparently). This kind of reminds me of how files work in Emscripten, they're all embedded in a big blob which it pretends is a filesystem...

PR #182, opened 2020-02-12
Add an argument to the wave callback

Daft-Freak:
Helpful when your callback is a class member...


Gadgetoid:
I'm actually surprised- and I'm presuming here- that your MP3 player hooks this at all. I guess it's about as good a proof/test of the wave callback you could ever want.

I believe this closely mirrors what SDL does with `userdata` for audio sample callbacks, so it makes sense. Thank you!

Daft-Freak:
Yeah, I'm streaming the decoded MP3 buffer into channel 0, so the other channels are technically free... Anyway, I've pushed the source here: https://github.com/Daft-Freak/32blit-music-player.

PR #185, opened 2020-02-12
MAX_FILELEN was wrong, should be 10 characters for string representat…

AndrewCapon:
…ion of uint32_t

### DIFF
--- a/firmware/flash-loader/flash-loader.hpp
+++ b/firmware/flash-loader/flash-loader.hpp
@@ -10,7 +10,7 @@
 #define ROW_HEIGHT 10
 #define ROW(x) Point(0,x * ROW_HEIGHT)
 #define MAX_FILENAME 256+1
-#define MAX_FILELEN 5+1
+#define MAX_FILELEN 10+1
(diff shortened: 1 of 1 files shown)


Gadgetoid:
Caught that this morning, see: https://github.com/pimoroni/32blit-beta/pull/183/files

Although we've gone with an overkill 16 characters for now.

I think we might rewrite this to use a packed, uint32_t as a constant 4 sequential bytes since it's much simpler to parse into a number (although a little more opaque for users reading the raw protocol). The file size can never be any larger than 33554432 anyway since it would exceed the size of the QSPI flash. Well, I guess 33554432 + Blit Header perhaps.

AndrewCapon:
As we are using a tool now a binary format would be much better, I did it originally as text for sending from shell scripts but that isn't really needed anymore.

Gadgetoid:
Oh that's a good point- I forgot you'd started off with a shell script.

We'll look into moving this over to a binary format, since it'll simplify the protocol and make things like this a non-issue (it can just fetch a header and cast it to a struct).

PR #187, opened 2020-02-12
Example fixes

ali1234:


Gadgetoid:
Doing the lord's work there, thank you!

PR #190, opened 2020-02-13
Improvement/menu refactoring

mylogon341:
- Created `MenuItem` enum
- Created helper functions for locations etc
- Menu now wraps
- Brought it together by writing a loop that creates menu items in a reusable way.
So
- Now, if you want to reorganise menu items, just reorganise the enum cases (less the last one).
- If you want to add a new item, 
    1. add an enum case, 
    2. add a case for the string, 
    3. add a case for the action when pressing A. 
    4. "Press A" is default. If you want something else, add a case for it in the for loop in `blit_menu_render`

### DIFF
--- a/32blit-stm32/Src/32blit.c
+++ b/32blit-stm32/Src/32blit.c
@@ -157,29 +157,73 @@ void blit_init() {
 
 }
 
+/*
 
-int menu_item = 0;
+    Menu Items
+
+*/
+
+enum MenuItem {
+    BACKLIGHT,
+    VOLUME,
+    DFU,
+    SHIPPING,
+    SWITCH_EXE,
+    LAST_COUNT // leave me last pls
+};
+
+//pinched from http://www.cplusplus.com/forum/beginner/41790/
+inline MenuItem& operator++(MenuItem& eDOW, int) {
+  const int i = static_cast<int>(eDOW) + 1;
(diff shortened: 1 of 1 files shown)


mylogon341:
ps: sorry for any syntactical/ pointer / style guide errors - c++ is far from my first language

mylogon341:
mind, now that the rows are "typed" the compiler is now warning that switch statements are missing cases. If the warnings are too annoying, I can stick the explicit cases back in, where applicable.

Gadgetoid:
I think we'll squash the warnings over time, but in the interim it looks like a good set of refinements.

Your overloading of `--` and `++` on the enum is... actually pretty cool.

Thank you!

I'm going to dive in and rename "Switch Exe" now :D

mylogon341:
Thanks! After coming from Swift there are a lot things that I'm sitting there going "WHY DONT YOU EXIST!?!" so have to find workarounds like that :'D
Ahh, what it is to be able to 
``` swift 
enum Test: Int, CaseIterable {
    case blah,meh,foo
    var triple: Int { rawValue * 3 }
}

Test.allCases.forEach({
    print($0) // blah meh foo
    print($0.rawValue * 2) // 0 2 4
    print($0.triple) // 0 3 6
})
```

PR #193, opened 2020-02-13
List files for the BlitEngine target

Daft-Freak:
Using a glob prevents CMake from reconfiguring automatically when a file is added or removed.


Gadgetoid:
Thank you!

PR #196, opened 2020-02-14
Move _estack to end of user RAM section

Gadgetoid:
This prevents the stack from stepping all over the heap. Or something like that. I've really lost the plot.

Thanks @trollied and @andrewcapon for pointing this out!

### DIFF
--- a/32blit-stm32/STM32H750VBTxEXT_FLASH.ld
+++ b/32blit-stm32/STM32H750VBTxEXT_FLASH.ld
@@ -33,7 +33,7 @@
 ENTRY(Reset_Handler)
 
 /* Highest address of the user mode stack */
-_estack = 0x20020000;    /* end of RAM */
+_estack = 0x2405A800;    /* end of RAM */
 /* Generate a link error if heap and stack don't fit into RAM */
 _Min_Heap_Size = 0x400;      /* required amount of heap  */
 _Min_Stack_Size = 0x400; /* required amount of stack */
--- a/32blit-stm32/STM32H750VBTx_FLASH.ld
+++ b/32blit-stm32/STM32H750VBTx_FLASH.ld
@@ -33,7 +33,7 @@
 ENTRY(Reset_Handler)
 
 /* Highest address of the user mode stack */
-_estack = 0x20020000;    /* end of RAM */
+_estack = 0x2405A800;    /* end of RAM */
 /* Generate a link error if heap and stack don't fit into RAM */
 _Min_Heap_Size = 0x400;      /* required amount of heap  */
 _Min_Stack_Size = 0x400; /* required amount of stack */


Daft-Freak:
I was looking here yesterday, I think that _estack was pointing to the end of DTCMRAM, so the stack and heap should never meet since they are in different memory regions? (If you do run out of heap, it steps over LDTC, which gives you a nice indicator of what happened...)

ali1234:
That is correct. This patch actually moves stack into the same region as heap, so makes it possible for them to collide.

Gadgetoid:
Yeah I'm not really sure what was going on here- 

Is it possible that previously (and this seems intuitively unlikely to me) the stack was growing to >128K and running out of DTCMRAM whereas the larger RAM_D1 section just coincidentally mitigates this one particular issue?

Or, perhaps, that some things we don't expect to be there were clashing with the stack in DTCMRAM?

Looking over the linker we have:

```ld
  /* Initialized data sections goes into RAM, load LMA copy after code */
  .data :
  {
    . = ALIGN(4);
    _sdata = .;        /* create a global symbol at data start */
    *(.data)           /* .data sections */
    *(.data*)          /* .data* sections */

    . = ALIGN(4);
    _edata = .;        /* define a global symbol at data end */
  } >DTCMRAM AT> FLASH
```

Which would suggest we have both initialised data sections (non-const variables with pre-set values, or that aren't set `NOLOAD` by the linker script?) and the stack in the same 128K memory region?

Daft-Freak:
Opposite ends of the same region, but yeah. According to the build output, my music player has 580 bytes in there. So the result of this change is (potentially) more stack, less heap, most of DTCMRAM unused.

Semi-related, I added this to my code yesterday when debugging my memory issues:
```cpp
#ifdef TARGET_32BLIT_HW
extern "C" void *_sbrk(ptrdiff_t incr)
{
    extern char end, __ltdc_start;
    static char *heap_end;

    if(!heap_end)
        heap_end = &end;

    // ltdc is at the end of the heap
    if(heap_end + incr > &__ltdc_start)
        return (void *)-1;

    char *ret = heap_end;
    heap_end += incr;

    return (void *)ret;
}
#endif
```

Makes allocation actually fail when you run out.

Gadgetoid:
I guess it boils down to whether it's more sensible to give the stack its own memory region: DTCMRAM and move initialised globals etc into RAM_D1 with the heap, or keep initialised globals in DTCMRAM and move the stack into RAM_D1.

Is this a sensible approach?

```patch
diff --git a/32blit-stm32/STM32H750VBTxEXT_FLASH.ld b/32blit-stm32/STM32H750VBTxEXT_FLASH.ld
index 583e9f1..1a34d39 100644
--- a/32blit-stm32/STM32H750VBTxEXT_FLASH.ld
+++ b/32blit-stm32/STM32H750VBTxEXT_FLASH.ld
@@ -33,10 +33,10 @@
 ENTRY(Reset_Handler)

 /* Highest address of the user mode stack */
-_estack = 0x2405A800;    /* end of RAM */
+_estack = 0x20020000;    /* end of RAM */
 /* Generate a link error if heap and stack don't fit into RAM */
 _Min_Heap_Size = 0x400;      /* required amount of heap  */
-_Min_Stack_Size = 0x400; /* required amount of stack */
+_Min_Stack_Size = 0x20000; /* required amount of stack */

 /* Specify the memory areas */
 MEMORY
@@ -131,7 +131,7 @@ SECTIONS

     . = ALIGN(4);
     _edata = .;        /* define a global symbol at data end */
-  } >DTCMRAM AT> FLASH
+  } >RAM_D1 AT> FLASH


   /* Uninitialized data section */
@@ -150,15 +150,24 @@ SECTIONS
     __bss_end__ = _ebss;
   } >RAM_D1

-  /* User_heap_stack section, used to check that there is enough RAM left */
-  ._user_heap_stack :
+  /* User_stack section, used to check that there is enough RAM left */
+  ._user_stack :
   {
     . = ALIGN(8);
     PROVIDE ( end = . );
     PROVIDE ( _end = . );
-    . = . + _Min_Heap_Size;
     . = . + _Min_Stack_Size;
     . = ALIGN(8);
+  } >DTCMRAM
+
+  /* User_heap section, used to check that there is enough RAM left */
+  ._user_heap :
+  {
+    . = ALIGN(8);
+    PROVIDE ( end = . );
+    PROVIDE ( _end = . );
+    . = . + _Min_Heap_Size;
+    . = ALIGN(8);
   } >RAM_D1

   itcm_data = LOADADDR(.itcm);
```

Do we *really* need 128K for the stack, or can it be allocated into the 63K remainder of `RAM_D2` leaving `DTCMRAM` for initialised globals and `RAM_D1` for the heap?

I think the real answer to this question is that it varies depending on the application, but since we're supposed to be supplying sensible defaults for beginners we should do that and anyone who wants to rewrite their linker scripts is welcome to do so.

Daft-Freak:
What's broken with the current approach? Or, what are we trying to fix here? Most of the options here seem to involve having less heap...

AndrewCapon:
What is the issue with just storing heap and stack in RAM_D1?

Daft-Freak:
Some additions:
Stack/heap were originally both in DTCMRAM, heap moved in https://github.com/pimoroni/32blit-beta/commit/df6e76a750a58cdb2cea5329e87a16fb660a9b20#diff-f50c6bc87caa7e67d16ec5001ad22a1bR153-R161. The reservation wasn't. So, we need at least:
```diff
diff --git a/32blit-stm32/STM32H750VBTxEXT_FLASH.ld b/32blit-stm32/STM32H750VBTxEXT_FLASH.ld
index bee44a4..60dc852 100644
--- a/32blit-stm32/STM32H750VBTxEXT_FLASH.ld
+++ b/32blit-stm32/STM32H750VBTxEXT_FLASH.ld
@@ -131,6 +131,7 @@ SECTIONS
 
     . = ALIGN(4);
     _edata = .;        /* define a global symbol at data end */
+    . = . + _Min_Stack_Size;
   } >DTCMRAM AT> FLASH
 
 
@@ -157,7 +158,6 @@ SECTIONS
     PROVIDE ( end = . );
     PROVIDE ( _end = . );
     . = . + _Min_Heap_Size;
-    . = . + _Min_Stack_Size;
     . = ALIGN(8);
   } >RAM_D1
 ```
To make `_Min_Stack_Size` do anything useful in the current script.

My main objection to this change is (apart from that it isn't fixing what it says it's fixing) that if the stack is moved from DTCMRAM, there will be almost nothing using it (most of the examples and music-player have ~500-600 BYTES of .data). Of course, it's entirely possible that I'm completely wrong about all of this...

AndrewCapon:
One of the things that DTCMRAM might be useful for (as it has twice the clock speed I think?) is to run certain code from rather than the flash. I did have an attempt at this to do some benchmarking but the stdlib initialisation barfed when initialising statics and I decided to revisit it at another time. Im guessing this was to do with my misunderstanding of setting it up correctly in the ld file.

I don't know if the pimoroni guys have looked at this, I seem to remember something being mentioned somewhere.

Of course this doesn't mean we can't run the stack from there as well.

AndrewCapon:
ps. I only said "What is the issue with just storing heap and stack in RAM_D1?" because I thought that was how it was working already!

Daft-Freak:
Is that not ITCMRAM? Which I think you can put functions in with `__attribute__ ((section (".itcm")))`. (At least it looks like that in the linker script: https://github.com/pimoroni/32blit-beta/blob/master/32blit-stm32/STM32H750VBTxEXT_FLASH.ld#L164-L172)

lowfatcode:
As yet we haven't worried too much about specifically where to allocate space for certain code/data. We have done some profiling of DTCM vs D1 vs D2 etc and found that for our workloads there hasn't seemed to be a substantial difference (perhaps because most of our tighter loops also need to do a lot of data push/pull and that vastly overshadows any potential performance gain.).

Our main focus at the moment has been giving the user the largest single contiguous block of memory to work in though depending on the allocator we end up using that may not matter so much.

If people do want to fiddle around and try stuff out we're eager to hear the results. :-)

AndrewCapon:
> Is that not ITCMRAM? Which I think you can put functions in with `__attribute__ ((section (".itcm")))`. (At least it looks like that in the linker script: https://github.com/pimoroni/32blit-beta/blob/master/32blit-stm32/STM32H750VBTxEXT_FLASH.ld#L164-L172)

ITCMRAM and DTCMRAM both run at 480 I think, I could be wrong though!

AndrewCapon:
> Our main focus at the moment has been giving the user the largest single contiguous block of memory to work in though depending on the allocator we end up using that may not matter so much.

So as Daft-Freak says probably best to keep the stack as is and separated from the heap.

Daft-Freak:
> > Is that not ITCMRAM? Which I think you can put functions in with `__attribute__ ((section (".itcm")))`. (At least it looks like that in the linker script: https://github.com/pimoroni/32blit-beta/blob/master/32blit-stm32/STM32H750VBTxEXT_FLASH.ld#L164-L172)
> 
> ITCMRAM and DTCMRAM both run at 480 I think, I could be wrong though!

But one of them is set up for running code from 😄

AndrewCapon:
True, but when I was trying to test it though I linking all of the code section in flash to mirror in DTCMRAM and run from there.

lowfatcode:
In theory we probably want to place the audio processing and rendering code into ITCM as it should be faster from there - but with some early testing we didn't see the improvements we hoped for. It's as good a place for it as any though so might as well have the system use it! There are considerations related to jump length and caching if i recall which may be why we didn't see much improvement. I'll try to experiment again and see what happens...

Issue #176, opened 2020-02-11
Investigate asset pipeline


Gadgetoid:
#175 introduced the ability to add binary assets to a project, but currently doesn’t allow any transformation from source.

It would be great if assets could be added by type and converted for use- IE a sprite calls sprite builder and then compiles the resulting binary data into the project ala https://github.com/Daft-Freak/32blit-beta/commit/a373fa4c096a86dff6d65fd3298db1a176b0a4db

PR #197, opened 2020-02-14
Rework asset embedding

Daft-Freak:
This main changes here are:
 - a single function call for all assets
 - assets are combined into as single binary before embedding
 - there is now a fallback for non-gcc compilers (embeds an array)
 - pulls in the sprite-packer stuff from https://github.com/Daft-Freak/32blit-beta/commit/a373fa4c096a86dff6d65fd3298db1a176b0a4db (#176)

The new syntax for assets is:
```
blit_assets(super-game
    SPRITE_PACKED
    amazing-sprites.png
    RAW
    a.bin b.bin
)
```

This generates an `assets.hpp` file that can be included to access all of the assets, it looks a bit like this:

```cpp
#pragma once

#include "engine/asset.hpp"

extern const blit::Asset asset_amazing_sprites;
extern const blit::Asset asset_a;
extern const blit::Asset asset_b;
```

Each asset has a `data` and `length`.


Daft-Freak:
It's currently a bit inconsistent with subdirs. `dir/file` ends up as `asset_dir_file` for a RAW asset, but `asset_file` for a sprite. Listing sprites `dirA/file.png` and `dirB/file.png` probably explodes (same output path).

Gadgetoid:
You're a magician!

Tested in VSCode using CMake tools both via WSL/MinGW and regular Windows/MSVC.

Builds and runs flawlessly in each case- granted I had some Python dependencies to install (I neglect my native Windows in favour of WSL) but that was as simple as using the terminal in VSCode to `python -m pip install bitstring construct pillow` (and pillow has a binary wheel so easy)

In both cases the `blit::Asset` approach is really elegant and porting over my Rocks & Diamonds code was also simple:

```patch
diff --git a/CMakeLists.txt b/CMakeLists.txt
index ee99663..345dd99 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2,5 +2,6 @@ cmake_minimum_required(VERSION 3.1)
 project (rocks-and-diamonds)
 include (../32blit-beta/32blit.cmake)
 blit_executable (rocks-and-diamonds rocks-and-diamonds.cpp)
-blit_asset (rocks-and-diamonds level)
-blit_asset (rocks-and-diamonds sprites)
\ No newline at end of file
+blit_assets (rocks-and-diamonds
+    RAW level
+    SPRITE_PACKED sprites.png)
\ No newline at end of file
diff --git a/rocks-and-diamonds.cpp b/rocks-and-diamonds.cpp
index d40fb07..83dd35e 100644
--- a/rocks-and-diamonds.cpp
+++ b/rocks-and-diamonds.cpp
@@ -37,11 +37,11 @@ void init() {
   set_screen_mode(ScreenMode::lores);

   // Load the spritesheet from the linked binary blob
-  screen.sprites = SpriteSheet::load((const uint8_t *)_binary_sprites_start);
+  screen.sprites = SpriteSheet::load((const uint8_t *)asset_sprites_png.data);

   // Load the level data from the linked binary blob
   // Note: This will be const! Should probably copy the level data to a local array first
-  level = new TileMap((uint8_t *)_binary_level_start, nullptr, Size(level_width, level_height), screen.sprites);
+  level = new TileMap((uint8_t *)asset_level.data, nullptr, Size(level_width, level_height), screen.sprites);
 }

 void render(uint32_t time_ms) {
diff --git a/rocks-and-diamonds.hpp b/rocks-and-diamonds.hpp
index a4bf598..25a7040 100644
--- a/rocks-and-diamonds.hpp
+++ b/rocks-and-diamonds.hpp
@@ -3,12 +3,10 @@
 #include <stdint.h>

 #include "32blit.hpp"
+#include "assets.hpp"

 void init();
 void update(uint32_t time);
 void render(uint32_t time);

-extern const char _binary_level_start[];
-extern const char _binary_sprites_start[];
-
 void update_camera();
\ No newline at end of file
```

I feel keeping `COMMAND ${CMAKE_LINKER} -r -b binary -o assets.bin.o assets.bin` is probably a little redundant if the "hacky" (actually quite standard and elegant if you look at other solutions to this problem out there) approach will work on GNU too. It seems almost arbitrary to support two ways of doing something. What do you think?

Daft-Freak:
Hmm the linker approach should be more efficient that just stuffing the data in an array, it's just a question of if it's worth the extra code. (Also, didn't want to delete ALL your code :smile: )

Added a little fix so that the asset pointers are const.

Daft-Freak:
Small thing I forgot to mention:
I made the custom commands depend on the scripts, which means that the assets get rebuilt when the script is updated.

Gadgetoid:
This is most awesome, thank you!

PR #198, opened 2020-02-14
Remove configure and install steps for SDL2.dll

Gadgetoid:
These steps don't belong in mingw.toolchain.

Since duplicate commands already exist in 32blit-sdl's CMakeLists.txt it makes sense to invoke them by setting `SDL2_DLL`

Right now they still make potentially incorrect assumptions about where SDL2.dll is located, but this assumption *should* be correct for all versions of SDL2 that have been cross-compiled as per the docs.


Daft-Freak:
I think I suggested roughly this at one point... https://github.com/pimoroni/32blit-beta/issues/139#issuecomment-583775525. Duplication was my fault I guess.

Gadgetoid:
So you did! Consider your suggestion implemented and tested then! (Clearly my reading comprehension wasn't hot there)

PR #204, opened 2020-02-15
Mass storage pr

AndrewCapon:
Changes to allow bits sdcard to be mounted on a pc over USB.

Added a MSC USB device.

Added USBManager class to handle switching between CDC and MSC and to detect mount status from host.

Updated firmware example so pressing home mounts the blit sdcard on the host and waits for an eject to go back to normal operation.


32blit-stm32/STM32H750VBTx_FLASH.ld
```diff
@@ -33,7 +33,7 @@
 ENTRY(Reset_Handler)
 
 /* Highest address of the user mode stack */
-_estack = 0x20020000;    /* end of RAM */
+_estack = 0x2405A800;    /* end of RAM */
```
Daft-Freak: Linker change...

32blit-stm32/Src/fatfs_sd.c
```diff
@@ -559,3 +540,577 @@ DRESULT SD_disk_ioctl(BYTE drv, BYTE ctrl, void *buff)
 
 	return res;
 }
+
```
Daft-Freak: Probably don't want the entire old file as a comment

32blit-stm32/Src/fatfs_sd.c
```diff
@@ -377,7 +357,7 @@ DRESULT SD_disk_read(BYTE pdrv, BYTE* buff, DWORD sector, UINT count)
 	if (Stat & STA_NOINIT) return RES_NOTRDY;
 
 	/* convert to byte address */
-	if (!(CardType & CT_BLOCK)) sector *= 512;
+	if (!(CardType & CT_SD2)) sector *= 512;
```
Daft-Freak: Revert of my fix.

32blit-stm32/Src/fatfs_sd.c
```diff
@@ -422,7 +402,7 @@ DRESULT SD_disk_write(BYTE pdrv, const BYTE* buff, DWORD sector, UINT count)
 	if (Stat & STA_PROTECT) return RES_WRPRT;
 
 	/* convert to byte address */
-	if (!(CardType & CT_BLOCK)) sector *= 512;
+	if (!(CardType & CT_SD2)) sector *= 512;
```
Daft-Freak: Same.

32blit-stm32/Src/fatfs_sd.c
```diff
@@ -468,8 +448,8 @@ DRESULT SD_disk_write(BYTE pdrv, const BYTE* buff, DWORD sector, UINT count)
 DRESULT SD_disk_ioctl(BYTE drv, BYTE ctrl, void *buff) 
 {
 	DRESULT res;
-	uint8_t n, csd[16], *ptr = (uint8_t*)buff;
-	WORD csize;
+	uint8_t n, csd[16], *ptr = buff;
```
Daft-Freak: Keep the cast?

32blit-stm32/Src/fatfs_sd.c
AndrewCapon: Whoops!

32blit-stm32/STM32H750VBTx_FLASH.ld
AndrewCapon: So the idea is to have .data and stack in DTCMRAM?

Daft-Freak: I was mainly pointing out that whatever is decided, it probably doesn't belong here. (In a commit about MSC)...

AndrewCapon: Ah ok.

Daft-Freak:
Hmm, something seems to have broken, I'm getting the device detected but "No Medium". (tried a few cards)

Gadgetoid:
Will test this also, as soon as I can.

If it doesn't completely hose the builds and flash-loader then it might be worth merging early and working out the kinks in smaller PRs.

Looking over the file list, it looks like some Eclipse project configuration has leaked into the repository. Namely:

* `.cproject`
* `.project`
* `.settings/language.settings.xml`
* `.settings/org.eclipse.cdt.codan.core.prefs`
* `language.settings.xml`

These should be removed from the PR, too.

Daft-Freak:
Should just need to drop the first commit for the eclipse stuff.

AndrewCapon:
How do you avoid adding commits when doing a PR, that eclipse commit is from my master?

Daft-Freak:
You can drop a commit when rebasing by using git rebase -i and deleting the commit from the list. (you can also reorder commits and squash (fixup) them)

AndrewCapon:
Thanks for the info DF, I will look into that...

Daft-Freak:
Something weird is happening. Debug builds work fine, Release (or MinSizeRel/RelWithDebInfo) don't. Somehow the sector size isn't getting set (I'm getting strange values in dmesg).

If I adjust `STORAGE_GetCapacity_HS` like this: 
```c
	*block_size = 0;
	if (disk_ioctl(lun, GET_SECTOR_SIZE, &sectorSize) == RES_OK)
	{
		if (disk_ioctl(lun, GET_SECTOR_COUNT, &sectorCount) == RES_OK)
		{
			*block_num  = sectorCount;
			*block_size = 512;//sectorSize;
			return (USBD_OK);
		}
	}
```

I consistently get something like:
```
[657255.134744] usb-storage 1-4:1.0: USB Mass Storage device detected
[657255.135049] scsi host10: usb-storage 1-4:1.0
[657256.153337] scsi 10:0:0:0: Direct-Access     STM      Product          0.01 PQ: 0 ANSI: 2
[657256.153697] sd 10:0:0:0: Attached scsi generic sg4 type 0
[657256.154096] sd 10:0:0:0: [sde] Sector size 0 reported, assuming 512.
[657256.154100] sd 10:0:0:0: [sde] 249737216 512-byte logical blocks: (128 GB/119 GiB)
[657256.154101] sd 10:0:0:0: [sde] 0-byte physical blocks
```

in dmesg, as if the = 512 somehow isn't happening. (Doesn't look like it's optimised out either.)

However, if I hardcode 512 a level higher up:
```diff
diff --git a/32blit-stm32/Src/usbd_msc_scsi.c b/32blit-stm32/Src/usbd_msc_scsi.c
index fb35f54..a199366 100644
--- a/32blit-stm32/Src/usbd_msc_scsi.c
+++ b/32blit-stm32/Src/usbd_msc_scsi.c
@@ -272,6 +272,7 @@ static int8_t SCSI_ReadCapacity10(USBD_HandleTypeDef  *pdev, uint8_t lun, uint8_
   }
   else
   {
+    hmsc->scsi_blk_size = 512;
 
     hmsc->bot_data[0] = (uint8_t)((hmsc->scsi_blk_nbr - 1U) >> 24);
     hmsc->bot_data[1] = (uint8_t)((hmsc->scsi_blk_nbr - 1U) >> 16);
```

It works, despite the fact that the same value should have been set through a pointer a few lines above. At this point, I probably need a debugger...

AndrewCapon:
Looks weird, I will have a look with a debugger in the morning...

AndrewCapon:
Well I have found out what is going on.

In ```STORAGE_GetCapacity_HS *block_size`` if getting changed when we call ```(disk_ioctl(lun, GET_SECTOR_COUNT, &sectorCount)```.

This is stored in r4 and r4 is being stepped on in ```disk_ioctl``` when -O2 or -O3 is used.

So ```*block_size = sectorSize;``` fails as it is now pointing to the wrong memory.

So a bit of a code generation is all I can guess

AndrewCapon:
so, this fixes it:
```
int8_t STORAGE_GetCapacity_HS(uint8_t lun, uint32_t *block_num, uint16_t *block_size)
{
  /* USER CODE BEGIN 10 */
	// get sector size and count
	volatile uint32_t uP = block_size;
	DWORD sectorCount = 0;
	WORD sectorSize = 0;
	if (disk_ioctl(lun, GET_SECTOR_SIZE, &sectorSize) == RES_OK)
	{
		if (disk_ioctl(lun, GET_SECTOR_COUNT, &sectorCount) == RES_OK)
		{
			*block_num  = sectorCount;
			uint32_t *p = (uint32_t)uP;
			*p = sectorSize;
			return (USBD_OK);
		}
	}
	return (USBD_FAIL);

  /* USER CODE END 10 */
}
```
I did try saving a volatile pointer instead but that also changed!

So I'm guessing this isn't a great solution, as the Fatfs we have only supports 512 size maybe we should just not call this and hard code it higher up?

AndrewCapon:
Actually thinking about it maybe we are looking at some stack corruption, I'm sure r4 would have been pushed and popped off the stack so maybe the popped stack value has been corrupted. We have a big thunder storm here at the moment so the computers disconnected, once it passes over I will have a look at the stack...

AndrewCapon:
Yep it is stack corruption.

I went for the most obvious target which was ```csd[16]``` in ```SD_disk_ioctl()``` and changed it to ```csd[32]``` and it springs into life.

AndrewCapon:
So here is the real problem, in ```SD_RxDataBlock()``` we have:

```
	do {
		SPI_RxBytePtr(buff++);
	} while(len--);
```

Where it should be:
```
	while(len--)
		SPI_RxBytePtr(buff++);
```

1 byte too many!

Daft-Freak:
Yep, fixed. On my branch I'm finally getting my 2GB card detected. (Good old off-by one errors...)

Daft-Freak:
New (minor) issue: after mounting/unmounting storage flashing over serial (PROG) is slower.

AndrewCapon:
And SAVE is as well, half the speed here!

So as they are both effected it has to be something to do with the USB transfer, or I'm thinking more likely the FIFO stuff.

I will need to get the debugger on it to check...

Daft-Freak:
Since I just caused a giant conflict, here's a rebase: https://github.com/Daft-Freak/32blit-beta/tree/MassStoragePR

Gadgetoid:
I'm for pivoting this PR over to the rebase if everyone's okay with that? I've been somewhat hesitant to merge what with @lowfatcode doing some hefty rewriting of the firmware and me possibly not understanding both components well enough to successfully integrate them.

AndrewCapon:
I searched to see what "pivoting this PR over to the rebase" means but am none the wiser?

Gadgetoid:
Sorry I was overcomplicating a simple process with fancy words, rather than referring to a specific git/github feature- I basically mean "close this PR and open a new PR for @Daft-Freak's rebase"

Daft-Freak:
Soo, should I open a PR for the (now slightly more rebased) rebase?

Gadgetoid:
Merged #242 thank you for this epic effort @AndrewCapon and for your continued support @Daft-Freak

PR #207, opened 2020-02-15
Allow custom name for assets

Gadgetoid:
At the request of @lowfatcode this improves the asset pipeline to allow for custom asset names. This is preferable to the name-mangling default since you can call your `sprite.png` or `level.tmx` by whatever name you so choose instead of `asset_sprite_png` and `asset_level_tmx`.

May not yet fix the pack_sprites function TODO, but allows the assets to be given names independent of their filename

It does this by popping 3 args at a time from blit_assets- type, file and name

Type and file are processed as usual, but names are collated and passed to asset-packer via a new argument: `--names`

This is a comma-separated list of asset names that must match the order of files supplied.

You can now use (for example):

```
blit_assets (rocks-and-diamonds
    MAP_TILED2BIN level.tmx asset_level
    SPRITE_PACKED sprites.png asset_sprites)
```

Include `assets.hpp`

And then access `asset_level.data` and `asset_sprites.data` in your code.

### DIFF
--- a/tools/asset-packer
+++ b/tools/asset-packer
@@ -11,8 +11,11 @@ if __name__ == '__main__':
     parser = argparse.ArgumentParser()
     parser.add_argument('--base-path', action='append', type=str)
     parser.add_argument('--inline-data', action='store_true', help='embed data as an array in assets.cpp')
+    parser.add_argument('--names', type=str, help='list of names for asset files (default is to mangle filename)', default='')
     parser.add_argument('files', nargs='+', type=str, help='input files')
     args = parser.parse_args()
+
+    args.names = args.names.split(',')
 
     bin_out = open('assets.bin', 'wb')
     cpp_out = open('assets.cpp', 'w')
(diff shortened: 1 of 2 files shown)


Daft-Freak:
I'll miss the ability to do `blit_assets(RAW a.bin b.bin c.bin)` and I liked not having to repeat the filename, but I guess this does have advantages. (Maybe I'm just prioritising typing less too much...)

tools/asset-packer
```diff
@@ -25,32 +28,41 @@ if __name__ == '__main__':
...
 
-        cpp_out.write('const blit::Asset asset_{}(_binary_assets_bin_start + {}, {});\n'.format(name, offset, length))
+        cpp_out.write('const blit::Asset {}(_binary_assets_bin_start + {}, {});\n'.format(name, offset, length))
 
 
     if args.inline_data:
         bin_out.close()
         bin_out = open('assets.bin', 'rb')
 
-        data = bin_out.read();
```
Daft-Freak: Ugh, I thought I'd learned to not do semicolons in python code by now.

Gadgetoid: I get that habit back if I write C enough- it never goes away!

Doesn't help that Python ignores them!

Gadgetoid:
Found a bug and added a polyfill for CmMake < 3.12.0 where there's no `list(JOIN ... )`

I was thinking about your shorthand- it's still mostly supported by other commands and it might be possible to still support it.

Maybe `blit_assets` and `blit_assets_named`?

Or, just `blit_assets` and the option to supply `AUTO` as a name.

Gadgetoid:
OH I just realised what you meant about `blit_assets(RAW a.bin c.bin d.bin)` - that's actually quite graceful and I've totally broken support for it.

Maybe we should reintroduce your whole code path under a different alias? There's got to be a best-of-both-worlds fix here.

Gadgetoid:
Since we now have a `while` loop that checks the length of `ARGN` we can consume arbitrary lengths of arguments for each command.

Maybe we coud have `RAW` and `RAW_NAMED` that just consume arguments until the next valid `type` is hit?

Daft-Freak:
Possible compromise loop:
```cmake

set(ASSET_TYPE "")

while(ASSETS_LENGTH GREATER 0)
	list(GET ASSETS 0 ARG)

	if(ARG STREQUAL "RAW" OR ARG STREQUAL "SPRITE_PACKED" OR ARG STREQUAL "SPRITE_RAW" OR ARG STREQUAL "MAP_TILED2BIN")
		set(ASSET_TYPE ${ARG})
		list(REMOVE_AT ASSETS 0)
		list(LENGTH ASSETS ASSETS_LENGTH)
		continue()
	elseif(ASSET_TYPE STREQUAL "")
		message(FATAL_ERROR "No asset type specified")
	endif()

	set(ASSET_PATH ${ARG})
	list(GET ASSETS 1 ASSET_NAME)
	list(REMOVE_AT ASSETS 0 1)
	list(LENGTH ASSETS ASSETS_LENGTH)

	set(ASSET_PATH ${CMAKE_CURRENT_SOURCE_DIR}/${ASSET_PATH})

	if(ASSET_TYPE STREQUAL "SPRITE_PACKED")
		pack_sprites(${ASSET_PATH} packed ASSET_PATH)
	elseif(ASSET_TYPE STREQUAL "SPRITE_RAW")
		pack_sprites(${ASSET_PATH} raw ASSET_PATH)
	elseif(ASSET_TYPE STREQUAL "MAP_TILED2BIN")
		pack_map(${ASSET_PATH} tiled2bin ASSET_PATH)
	endif()

	list(APPEND ASSET_FILES "${ASSET_PATH}")
	list(APPEND ASSET_NAMES "${ASSET_NAME}")
endwhile()
```

(Maybe. not actually tested)

Should still allow omitting type
```cmake
bilt_assets(
    target
    RAW
    a.bin a
    b.bin b
    c.bin c
    SPRITE_PACKED
    sprite-a.png sprite-a
    sprite-b.png sprite-b
)
```

Gadgetoid:
I think coming back around to this, and with feedback from @lowfatcode, and with Visual Studio in consideration we should try to move as much of the logic behind asset configuration out of CMakeLists.txt and into a config-file driven asset builder that can handle all the edge cases and be as simple or complex as the developer needs.

I've raised https://github.com/pimoroni/32blit-beta/issues/212 to track progress on this.

PR #208, opened 2020-02-16
Unify linker scripts

ali1234:
Replaces the internal and external flash linker scripts with a single
script that is configured on the ld command line. This mitigates the
risk of them getting out of sync. The script defaults to external
flash. Defining the symbol "FLASH_TARGET_INT" causes it to use
internal flash instead.


ali1234:
This will make it easier to work on persistence and memory layout as there is now only one file to edit.

It is very likely to break the visual studio projects.

Gadgetoid:
Ooh, this is graceful! I was hacking on the linker script recently and came across exactly this problem. Thank you!

ali1234:
It's not clear to me why the visual studio project builds passed with this change. Do they not build ARM binaries?

PR #209, opened 2020-02-16
Add ITCM init to startup

Daft-Freak:
Mostly a copy/paste of the .data init. My assembly skills don't extend much past (barely) reading it... But this does work.


Gadgetoid:
Can't say my assembly comprehension is much more advanced, but it looks sensible to me. Thank you!

PR #210, opened 2020-02-16
Fix discarding unused fonts

Daft-Freak:
The global constructor would reference the font data, preventing it from being discarded. Doing it at compile time avoids this.


Gadgetoid:
I clearly need to research constexpr because I've been splashing it around possibly without understanding what it does. Thank you!

Daft-Freak:
It basically means "this function can be evaluated at compile time". Quite useful for this kind of thing.

PR #211, opened 2020-02-16
Fix measure text for multiple consecutive newlines

Daft-Freak:
...and I find another font/text issue. The old code would check if the next character was a newline and skip it, which would fail if there were consecutive newlines (it would treat the additional newlines as spaces).


Gadgetoid:
Whoops, good catch! Thank you.

PR #213, opened 2020-02-17
JPEG decoding

Daft-Freak:
Using the hardware decoder (+SDL support).

Adds two functions:
```c++
  JPEGImage decode_jpeg_buffer(uint8_t *ptr, uint32_t len);
  JPEGImage decode_jpeg_file(std::string filename);
```

Also adds SDL_image as a dependency for 32blit-sdl. It's a bit annoying as it has no CMake config and there is only a find module for v1, had to do some semi-manual detection.

Draft for a few reasons:
 - I don't know if anyone else has a use for this
 - ~It includes the mass storage/SD changes (was testing while working on this, haven't rebased back to master)~
 - Doesn't update the docs for the new dependency (SDL_image)
 - ~Also haven't updated VS, still builds as nothing is using the new stuff~
 - ~This is going to break Travis when merged until the cache is cleared (no SDL_image in cache).~

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,6 +34,7 @@ jobs:
       addons:
         apt:
           libsdl2-dev
+          libsdl2-image-dev
           python3-pip
           python3-setuptools
       before_deploy:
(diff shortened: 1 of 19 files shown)


Daft-Freak:
Now based on the right branch.

Daft-Freak:
Okay, nothing is broken now, just needs some docs.

Gadgetoid:
Looks like you've pretty comprehensively solved any issues locating `SDL_image`- to be honest I'm surprised we've got away without that as a dependency up to this point.

I also don't know if anyone will use this, but I also don't think that should stop us supporting it. It could be the lynchpin for a port of a game that has existing jpeg assets, or just a handy way to load backdrops for a point-and-click adventure.

Daft-Freak:
Merging this also means I can release the MJPEG player source :smile: . Anyway, took a break from DOOM to update some of the docs.

Gadgetoid:
Well, we can but merge and see who gets dependency issues! It would be a shame not to exploit a perfectly good hardware feature. Thanks!

PR #214, opened 2020-02-18
Fix list_files for symlinked directories

Daft-Freak:
Bit of a corner case, but managed to hit it...


Gadgetoid:
Wowser- how'd you end up trying to read from a symlinked dir? On an SD card?

I suspect others may do the same and be confused, so thanks for the fix!

Daft-Freak:
I have a lot of symlinks in my Music dir :smile: (SD cards shouldn't be affected as FAT doesn't have symlinks.)

PR #215, opened 2020-02-19
Class based file API

Daft-Freak:
Wraps the existing API in a class. I think this is a bit nicer (okay, I mainly did this because I kept forgetting to close files...).

```c++
int32_t file = blit::open_file("some.thing")
blit::read_file(file, ...);
return; // oops
```
->
```c++
blit::File file("some.thing")
file.read(...);
return; // yay, destructors
```

Guess the next step would be to make the old API internal...


Gadgetoid:
I prefer this to the old method, since keeping track of file-descriptors feels really clunky and- of course- exposes that "forgot to close file" footgun. Will merge and hope it sticks ;) Thank you.

PR #216, opened 2020-02-19
WIP: Improvements/menu v2

mylogon341:
I've made some decent headway on making this a much more dynamic + neat affair.
So far, I have effectively mirrored the look and functionality of the existing menu.
However, for this refactor, I have planned in to be able to drill down into submenus ("About", for example) which will navigate much like a mobile phone navigation system, for example.

I have decided to open a wip mr at this point in case I am missing some really terrible optimisation points that someone is able to chime in on.

Thanks :)


mylogon341:
Some things to note here - I have yet to add 'About' bits, so the mock text is in there. If you agree that is a nice addition, I imagine there are just some bits to add to the Travis file + fetching in code. Sadly, I don't know how to do that. Otherwise, just comment out (to leave as an example) the option from the data source file.

mylogon341:
video of it in action https://youtu.be/gJIp_N_HN6A

Gadgetoid:
Nobody wants to write menu screens (except you, apparently 😆), which is why I wonder if we should implement this into the 32Blit core API, rather than relegating this concept to a neglected corner of the firmware where it wont ever get to truly shine.

I'd *much* prefer to use this to build an in game menu, with some overloads for handling drawing style, than faff around doing it manually.

mylogon341:
That'd be pretty cool. Happy for anyone to take over this one if putting it into the core api is agreed upon. Otherwise, i'll wait a bit until i can bear to look at it again XD.

Daft-Freak:
Hmm, writing a menu system was actually on my TODO list...

Gadgetoid:
How does the code in this PR align with what you'd have put together @Daft-Freak? It all seems pretty sensible to me, although- due to being written for the firmware menu- it's a little opinionated toward volume bars.

For example- Individual settings might also include lists of values, like `Difficulty: [ Easy, Normal, Hard, Insane ]` selected by pressing left/right.

Daft-Freak:
Hmm, I'm not entirely sure where my would've went, I think my main idea was more of a grid than a list. I'll probably have to have to try this out at some point and see if I have any ideas.

(I suppose the closest I got to working on this was my file list...)
Screenshot from 2020-02-21 16-58-59
)

mylogon341:
> For example- Individual settings might also include lists of values, like Difficulty: [ Easy, Normal, Hard, Insane ] selected by pressing left/right.  

Hold my beer..

Daft-Freak:
Small recommendation: you don't want to ever put a `using namespace` in a header. It tends to cause name collisions. (Like the one you're getting on Traivs.)

mylogon341:
Ah good to know. I’ll get it working and finished when I get home. I’ll accept and tips and critique at this stage - I’m very rusty at this C stuff so appreciate it all, so don’t worry you won’t hurt my feelings.

mylogon341:
*SIGH*
Ok. I think I've just about nailed this now. 
There are separate game and firmware menus and I've added a confirm button. 
I've added the game menu to flight only, but added the confirm button to the 'exit game' button, which is a duplicate of the button in the system menu. I would have changed this to 'reset' or something but don't know how to do that.
The game menu should now be pretty straight forward to make for a game. Just create a 'Menu' object and pass it to the MenuController singleton. The rest should be taken care of after that.

Daft-Freak:
I had a look through the code and it looks like there's a lot of vector copying/resizing going on, for example:

```c++
about.push_back(MenuItem("Version", "number"));
```

Could be:
```c++
about.emplace_back("Version", "number");
```

to avoid a temporary. In fact the entire function could be simplified using an initialiser list:

```c++

vector<MenuItem>aboutItems () {
    return {
        {"Version", "number"},
        {"Build number", "number"},
        {"Build time", "date"}
    };
}
```

Or since the function returns the exact same values every time, just make it a static list:
``` c++
static std::vector<MenuItem> about_items {
    {"Version", "number"},
    {"Build number", "number"},
    {"Build time", "date"}
};
```

(Note: that was from memory, so may be slightly off)

Gadgetoid:
Mr. Feedback 😆 

Sorry I've just realised that I have conflicted the heck out of this. It might need to be a squash-and-merge or mega-rebase jobby to get this into Master anyway. Your commits make for entertaining reading here (they're at least more coherent than the ones in our old 32Blit internal repo) but would probably mess up Master.

Anwyay I'm going to give this a try!

mylogon341:
Oh, I like Mr. Feedback more as it was actually 'merge request'. I think i'm going to start calling people by their contributions now :) 
I was sat going "make for entertaining reading? like what?"
This tickled me
Screenshot 2020-03-05 at 23 55 54
That cross 🤣

mylogon341:
I think the main conflict is that I was trying to intercept all menu and home buttons in the MenuController, whereas that approach just wasn't working for me for some reason (as nice as I thought it would be to contain it all) as there was some "if menu or home then switch the render & update functions" so it was kind of split in two places, which I couldn't seem to iron out for love or money..

Gadgetoid:
Frustratingly we've gone and conflicted this again, which is annoying because every time I pick up the current firmware with its... "menu"... I want to scream.

I'm tremendously out of the loop with 32blit,  but I'll have a look over this and see what might need doing.

Gadgetoid:
Daft's menu refactor trumped this, somewhat! Closing for now.

PR #217, opened 2020-02-19
WIP: Examples/flight

mylogon341:
This is kind of a "nice to have". I have been toying around with this example and (hopefully) fixed a few little issues I had with it + added some additional functionality to it. Really, I think the rendering code can be cherry-picked from this, maybe, and the "make it an actual plane" code can optionally be discarded on account of it being for fun.

The flying code has received fairly little testing, but the rendering improvements have received more, hence with wip.

Feedback welcome :)


mylogon341:
and i just noticed that i left a custom lerp function in there  🤦‍♂

Gadgetoid:
Oooh, nic! We clearly need to get the Mode7 stuff working again, it's rather anti-climactic without it. I also need to get around to flashing this on a blit and seeing how it feels with the analog stick.

Gadgetoid:
Thank you :D

PR #219, opened 2020-02-20
Class based file API - part 2

Daft-Freak:
The part that breaks everyone's code. Ports the one example using files, shuffles code around to hide the old API then does some cleanup. Also fixes a small leak. (I can drop the last commit if the pointers are important for something.)

Edit: best viewed with whitespace changes hidden: https://github.com/pimoroni/32blit-beta/pull/219/files?w=1

### DIFF
--- a/32blit-sdl/File.cpp
+++ b/32blit-sdl/File.cpp
@@ -13,8 +13,6 @@
 #include "File.hpp"
 
 static std::string basePath;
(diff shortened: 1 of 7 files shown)


Daft-Freak:
Removed the function pointer change since I know what those are for now.

Gadgetoid:
Thing of beauty and a joy forever! Thank you.

PR #223, opened 2020-02-21
Make asset packing also depend on the yaml

Daft-Freak:
Makes sure a rebuild is triggered when editing the config for an asset.

(I always seem to run into the weird corner cases...)


Gadgetoid:
Whoops, yes, this is important! Totally overlooked it despite - IIRC - you mentioning it before. Thanks.

PR #224, opened 2020-02-21
Fix the symlink fix

Daft-Freak:
Helps if you pass the path to stat instead of the filename... Must have tested on a dir in the working dir.


Gadgetoid:
Whoops :D Thanks for the fix to the fix. We need to find a way to fix this now.

Daft-Freak:
Haha, I fixed a fix to a feature by myself. Third time lucky?

PR #225, opened 2020-02-22
Remove old asset packing

Daft-Freak:
I know this is coming, might as well delete it myself.


Gadgetoid:
I'm remiss to bite the bullet and kill the old way- but onwards and upwards! Thank you.

PR #226, opened 2020-02-24
Generate version number defines

Daft-Freak:
A bit of a WIP as I'm not sure where the defines/header should go (probably somewhere internal). Also, left a test commit in, and VS is broken. Inspired by #216. 

Test release here: https://github.com/Daft-Freak/32blit-beta/releases/tag/1.2.3

### DIFF
--- a/32blit/CMakeLists.txt
+++ b/32blit/CMakeLists.txt
@@ -9,6 +9,7 @@ set(SOURCES
 	engine/running_average.cpp
 	engine/timer.cpp
 	engine/tweening.cpp
+	engine/version.cpp
 	graphics/blend.cpp
 	graphics/color.cpp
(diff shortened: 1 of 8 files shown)


Gadgetoid:
Ah, of course. Helpful Visual Studio does not invoke CMake to generate the version number include.

Daft-Freak:
I'll probably just add a placeholder file for the VS solution since it isn't used to build releases.

Daft-Freak:
Builds everywhere now. Changed to a pair of getters to avoid include path fun.

Gadgetoid:
This looks entirely sensible! Thank you.

Daft-Freak:
Oh, I left the test commit in! :man_facepalming:

PR #229, opened 2020-02-26
Clean up some unused declarations in engine.hpp

Daft-Freak:
Including two for the same thing. Neither of these are defined anywhere.


Gadgetoid:
Oh wow, these are *old*. IIRC they were related to the Lua bindings.

PR #235, opened 2020-02-28
Small VS code docs changes

Daft-Freak:
Forgot about these ones. I don't know if we need anything else for #139?

### DIFF
--- a/docs/VSCode.md
+++ b/docs/VSCode.md
@@ -12,7 +12,7 @@ Then open the cloned repository with "Open folder...".
 
 Windows needs a little help to find the required dependencies for local builds.
 
- - First download SDL as described in "Option 1" of the [Visual Studio](Windows-VisualStudio.md) docs.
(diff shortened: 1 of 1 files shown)


Gadgetoid:
Thank you.

I'm not sure- the trouble with having an already working setup is it's hard to unravel everything you changed to get you there. I might try these instructions on another computer to see what we're missing.

One thing that pops out- if I'm not totally missing the plot- is that these instructions don't mention installing MSVC which I think can be installed standalone and doesn't have to involve the whole Visual Studio suite. It's the MSVC kit that I use in Visual Studio Code to run and debug local SDL builds.

Daft-Freak:
They also don't describe installing the compiler on other platforms, I was trying to leave platform specific stuff to the platform setup docs. Hmm...

Gadgetoid:
That's a good point- we don't want to repeat ourselves necessarily!

PR #236, opened 2020-02-28
Added basic code for storing values in backup sram. When changeing be…

AndrewCapon:
…tween firmware/game via menu this setting will be remembered on reset.

Just the basics really as I am not sure how you guys want to go about this in the future.

Firmware/Game state is now remembered and it will reset to the same state.

"PROG" will reset to firmware, program and then stay in Game state.
"SAVE" will reset to firmware, save and then stay in Firmware state.

### DIFF
--- a/32blit-stm32/Inc/32blit.h
+++ b/32blit-stm32/Inc/32blit.h
@@ -43,8 +43,3 @@ void blit_menu();
 
 extern void blit_enable_ADC();
 extern void blit_disable_ADC();
-#define BACKUP_START_STATE_INDEX    (0)
-#define BACKUP_START_STATE_FIRMWARE (0)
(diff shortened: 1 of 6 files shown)


AndrewCapon:
Now moved the persist struct into battery backed up ram.

Holding down "home" when booting will force into firmware, otherwise it will boot to last used firmware or game.

Daft-Freak:
This is so much better than my hacky attempt at the same thing. If only I could press the merge button :smile:

Gadgetoid:
Merged! Excellent stuff. I don't know why but I couldn't get the backup SRAM to be persistent for toffee. This should, in theory, keep persistence ... persistent... even in the lowest power states.

PR #238, opened 2020-02-29
Update macOS documentation to source SDL2 and SDL2_Image from homebrew

ntwyman:
Recent Jpeg code requires SDL IMage. This describes how to install it on macOS


Gadgetoid:
Thank you! We knew the SDL Image requirement would play havoc with the docs/setup. Cheers for taking the time to PR a fix!

PR #239, opened 2020-02-29
Target C++17

Daft-Freak:
I've seen this mentioned a few times. I was a bit concerned that the arm-gcc in Ubuntu 18.04 (Traivs, WSL) is a bit old for full support, so I created this branch to check. It didn't throw any errors anywhere so at least enabling it works. I guess Travis should catch any incompatible stuff in future...

If we do end up needing a newer gcc/libstdc++ there's always this: https://github.com/pimoroni/32blit-beta/compare/master...Daft-Freak:19.10-toolchain


Gadgetoid:
I think C++17 would let me use `static` in header-only asset includes, too. Yay! Thank you.

Daft-Freak:
Still a bit mean on the compiler though... Now to C++17-ify some stuff...

PR #240, opened 2020-02-29
Add a new font constructor

Daft-Freak:
Not very exiting on its own... Takes an array like this:
```
 "FONT" (4 bytes)
number of characters (1 byte)
fixed character width (1 byte)
character height (1 byte)
vertical spacing (1 byte)
  
variable character widths (1 byte per character)

character data (fixed w * ((height + 7) / 8) bytes per character)
```


Gadgetoid:
Sometimes the not exciting things are the most important! Thank you.

Daft-Freak:
The exciting bit is in the other repo :smile:

PR #241, opened 2020-02-29
Update 32Blit-Firmware.md

mylogon341:
The section is already called 'flashing the firmware'. Naming the sections that too seems ott


Gadgetoid:
Agreed- this is much more skimmable! less is more. Thank you!

PR #242, opened 2020-02-29
Rebased Mass Storage

Daft-Freak:
#204. Rebased since I've caused conflicts for it at least twice. Also drops the eclipse stuff.


Gadgetoid:
May Git have mercy on my soul!

Also thank you immensely for rebasing this- it was top of my list to look into.

PR #243, opened 2020-03-01
Fixed template

ymauray:
I copied the template and it did not compile. I figured this was out of date ? So I - hopefully - fixed it.


Gadgetoid:
Whoops, yes, very out of date! Good catch! Thank you.

PR #246, opened 2020-03-01
Filesize next to filename, for files on SD card

ymauray:
Following the upload problems we had lately, I thought it could be usefull to have the filesize next to the filename, when listing files on the SD card.
My C++ skills are a bit rusted, and string manipulation has always been a bit of a mystery so I'd be happy to fix anything if needed !


Daft-Freak:
Won't this break the flashing as it will try to open a file called `thing.bin    [some number]`?

ymauray:
Yep... my bad. Let me rethink that !

PR #247, opened 2020-03-01
Enable float formatting for ext flash builds

Daft-Freak:
This adds slightly less than 10k to the .bin and makes `printf("%f")` and `std::to_string(1.0f)` work.

### DIFF
--- a/32blit-stm32/CMakeLists.txt
+++ b/32blit-stm32/CMakeLists.txt
@@ -180,7 +180,7 @@ function(blit_executable NAME SOURCES)
 		DESTINATION bin
 	)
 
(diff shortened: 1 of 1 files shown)


mylogon341:
👌

Daft-Freak:
Well, that was fast :smile:

mylogon341:
That's what sh..

Gadgetoid:
Awesome- a small size cost for implementing the "principle of least surprise." Thank you.

Daft-Freak:
I think for another ~8k we can have a working `scanf("%f")`...

PR #248, opened 2020-03-01
Filesize

ymauray:
Following the problems we have currently with the upload (PROG and SAVE), I thought it would be usefull to show the size of each file stored on the SD card. 
Using a second array of string for the labels that are actually displayed, it is still possible to flash the file from the SD card to the internal flash.


Gadgetoid:
Agreed! Any quality-of-life stuff in the firmware makes sense for now.

PR #252, opened 2020-03-02
Interrupt on GPIO_PIN_5 for system menu

Gadgetoid:
This changes the home button behaviour from polling in the main loop to triggering on an interrupt.

It defines two behaviours:

1. Short press = Open/Close system menu
2. Long press (500ms) = Immediately reset to firmware (only happens in app)

The latter is to replace the force-of-habit "press reset to return to system menu" with a more sensibly defined behaviour, since the reset button will not be so easily accesible on final units.

The former fixes the horrible bounce issue on the menu button.

TODO: Disable button interrupt when switching exe, because it breaks horribly (read: drops to `Default_Handler`) when apps don't define the interrupt vector but the interrupt is still enabled.

### DIFF
--- a/32blit-stm32/Inc/stm32h7xx_it.h
+++ b/32blit-stm32/Inc/stm32h7xx_it.h
@@ -67,6 +67,7 @@ void TIM3_IRQHandler(void);
 void TIM4_IRQHandler(void);
 //void TIM6_DAC_IRQHandler(void);
 void OTG_HS_IRQHandler(void);
+void EXTI9_5_IRQHandler(void);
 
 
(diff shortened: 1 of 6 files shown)


mylogon341:
> The former fixes the horrible bounce issue on the menu button.

Hallelujah

Gadgetoid:
Known issues:

1. It's... kinda unresponsive? Frustratingly so. I'm not quite sure why, either. Perhaps I've just sped up debounce issues.
2. Sometimes the menu will pop up by default when using press-and-hold to return to firmware, this shouldn't happen.

PR #254, opened 2020-03-02
Compile C files with the C compiler

Daft-Freak:
Doesn't fix the new -Wregister warning (it's used in a header included nearly everywhere), but is slightly more sane.


Gadgetoid:
Good grief, that's a few changes and no mistake. I think our intent in future is to migrate all non-STM32 HAL code into nice tidy .cpp files, so it should get... less chaotic. Thank you!

PR #255, opened 2020-03-02
Use string_view for text functions

Daft-Freak:
Avoids string copying, especially when using horizontal alignment.


Daft-Freak:
Looks like this is a (small) code size improvement, -808 bytes on the firmware.

Gadgetoid:
Thank you! I've run out of words to say how awesome you are, so have a balloon instead 🎈

PR #257, opened 2020-03-02
No wrap around

ymauray:
In the firmware :
- having the first file selected and pressing "up" resulted in having the last line selected at the bottom of the screen ;
- having the last file selected and pressing "down" resulted in having the line past the last file selected ;
- wrap around is evil [(c) @AndyCap ^^, and I agree]

This PR prevents any wrapping. Pressing up on the first file will just keep the selection there. Pressing down on the last file will also keep the selection in place.


Gadgetoid:
I don't understand the hate for menu wrap-around- what's the practical case against it?

It creates a shortcut from items at the front/back of the list, where the user can "wrap around" instead of scrolling through every single item to get there.

The top/bottom list items have no special priority, so the studied UX benefits of being guaranteed that you can hit the top/bottom item without wrapping around don't really apply.

Is there some angle I'm missing? Searching for any study of UX concerns around menu navigation just gives me lots of HTML/CSS noise.

ymauray:
It's not that I _hate_ it, it's more that I think it's not really practical.
But I guess it's like `vim` vs `nano` or `Windows` vs `Linux` : it's a question of personal choice.

Come to think of it, I shouldn't have changed the current behaviour, but simply fix the problem. I apologize for that, and will update the PR as soon as possible to bring the wrap back.

Gadgetoid:
You might need to bump `uint8_t m_uCurrentFile = 0;` to an `int16_t` to handle the negative cases, since at the moment the `uint8_t` will just wrap around and never go `< 0`.

ymauray:
> You might need to bump `uint8_t m_uCurrentFile = 0;` to an `int16_t` to handle the negative cases, since at the moment the `uint8_t` will just wrap around and never go `< 0`.

I don't think it's an issue because when there is no file on the SD card, a message is rendered instead of the list, and the value of `m_uCurrentFile` is not used.

Gadgetoid:
Sorry you're right, I totally misread the code!

PR #258, opened 2020-03-02
Code cleanup, double promo fixes

Gadgetoid:
Avoid some unecessary float to double promotion issues - perhaps it would be more idiomatic of C++ to use `std::` prefixed functions? But this does mean littering `std::` everywhere.


Gadgetoid:
Splitting this into separate PRs.

PR #259, opened 2020-03-03
Fix JPEG decoding after the compiler changes

Daft-Freak:
Wrapping the include in `extern "C"` isn't the nicest way to handle this, but jpeg_utils.h is a generated file. Missed this since there are no users in the repo.


Gadgetoid:
Should probably add a jpeg example at some point to keep this in check. Thanks!

PR #260, opened 2020-03-03
Shared API data

Daft-Freak:
Moves the data/functions to interface with the firmware to a special section. This is the first step towards not building the firmware into every app and eventually, multiple apps/PIC. One thing that isn't covered here is the JPEG API, which needs some extra work.

Sending this first since it is a pain to rebase and is probably the most finished part of the branch. 

(This was @Gadgetoid's idea, which I extended while getting PIC going.)


Gadgetoid:
This will probably cause havoc with my spaghetti mess of changes in #258 but it's got to the point where I need to split that branch into individual PRs now, I think.

Particularly in my branch:

1. I've rewritten the button handling- again- to user a timer in addition to the button interrupt so it can resolve a long press and/or debounce without needing to mess about with HAL_GetTick. - https://github.com/pimoroni/32blit-beta/pull/258/commits/ef192dcdf580c24e17de4710eee790b9294d856e
2. I have removed `battery_vbus_status` and `battery_charge_status` completely and just grab the relevant bits from `battery_status` when needed. (Note: I've totally broken Hardware Test in SDL here, and should just accept that advanced battery information will have to be added to the system menu) - https://github.com/pimoroni/32blit-beta/pull/258/commits/09174182546299c5de04b01e0bb3ed8c3e0b5313

I'll try to look over/merge the PIC API tomorrow and cherry pick/rebase my changes into some more succinct PRs.

Thanks again!

Daft-Freak:
Hmm, yeah I just deleted the battery stuff from hardware-test: https://github.com/pimoroni/32blit-beta/pull/260/commits/5f9498766fe2bb728e8291850121b6da0f8a666d. There's hopefully not much else here that should conflict.

I need to do a some tidying on the rest of the PIC stuff (And make it possible to build without needing an entirely custom toolchain).

Daft-Freak:
Okay, the JPEG API is sorted, but i can't add the commit here yet as it depends on #259.

Issue #220, opened 2020-02-20
Investigate "Position Independent" external flash applications


Gadgetoid:
Investigate `-fPIC` or position-independent linkage that would allow a program to be stored at and run from an arbitrary location in external flash.

Can an external flash program be compiled position-independent and launched from the QSPI flash no matter where it's located?

Gadgetoid:
As an extension to this, there's no reason why the user binary should ever include any of the STM32 initialisation routines. It should only really define `init`, `update` and `render`.

Should look into whether or not we can:

1. Define our own structured API that works like persistence and allows values such as button states, accelerometer data and LED output to be shared between the firmware and the user code via clearly defined linkage
2. Define our own vector table format that puts the offsets of `init`, `update` and `render` into a region of memory that the firmware knows about, like:

```ld
g_blitVectors:
  .word init
  .word update
  .word render
```

PR #269, opened 2020-03-04
Don't include the firmware in every app

Daft-Freak:
Part two of the quest for PIC (#220)...

This probably needs a lot of testing. Known issues:
 - ~If you don't call `set_screen_mode` in `init`, you don't get a screen.~
 - ~"Exit Game" is mostly broken (it does return to the file list, but with whatever screen mode the app set and the app's sound still going)~
 - The memory allocations might need adjusting (user code is similar to before, firmware heap/bss is in RAM_D2)
 - ~Timers are broken~


Daft-Freak:
https://github.com/pimoroni/32blit-beta/pull/269/commits/7ca49a3263747b454fdee6c10e9a3526903ad76a / https://github.com/pimoroni/32blit-beta/pull/269/commits/04b0e7321352c4ac332878b04fe818de54841cb2 get timers working, but seem a bit messy...

Daft-Freak:
Default screen mode is back, exit works and the examples seem to work.

Daft-Freak:
The Windows builds failed... because Travis didn't run them on Windows? :confused:

zenodante:
> Part two of the quest for PIC...
> 
> This probably needs a lot of testing. Known issues:
> 
> * ~If you don't call `set_screen_mode` in `init`, you don't get a screen.~
> * ~"Exit Game" is mostly broken (it does return to the file list, but with whatever screen mode the app set and the app's sound still going)~
> * The memory allocations might need adjusting (user code is similar to before, firmware heap/bss is in ITCMRAM)
> * ~Timers are broken~

Put heap/bss in ITCMRAM may not a right place. It's only for fast code in ram and ARM don't suggested you to put other stuff into it.

Daft-Freak:
Hmm, that makes sense. I think that part came from @Gadgetoid's original patch.

Daft-Freak:
Just fixed the "currently running from" and reset commands so that the tool doesn't attempt to write ext flahs while running from it... Interestingly, saving bins to SD worked in the background.

Daft-Freak:
(A few more rebases later...)

So, this is at the point where everything seems to work (I've been running it for a while now). Not sure if there's anything left that needs tidying up?

Gadgetoid:
We've got a tag and release for a previous stable build, so we should be able to merge this and go onwards and upwards!

Gadgetoid:
A momentus occasion!

Thanks for all your work on this monster of a change, it's a heck of an improvement in more ways than I could list.

PR #274, opened 2020-03-05
Minor Emscripten changes

Daft-Freak:
Shaves a bit off the .js and makes JPEG loading actually work... which adds ~150k to the .wasm.


Gadgetoid:
Ooh nice! Thank you.

PR #275, opened 2020-03-05
Giant pile of tidying

Daft-Freak:
Went to adjust a few things, got carried away... Some of this is more clang-tidy warnings. (Interestingly it can detect more double-promotions than GCC can.)

### DIFF
--- a/32blit-sdl/Input.cpp
+++ b/32blit-sdl/Input.cpp
@@ -133,7 +133,7 @@ void Input::_virtual_tilt(int x, int y) {
 	int z = 80;
 	x = x - (win_width / 2);
 	y = y - (win_height / 2);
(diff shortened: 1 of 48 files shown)


AndrewCapon:
Remove redundant voids!

Madness :)

Gadgetoid:
Is there any way we can run clang tidy with a `--check-only` flag in Travis to prevent any new issues from creeping in?

Gadgetoid:
Thank you- had a quick skim and it all looks completely rational. We could definitely use some gentle code style nudges, but don't want to put up too many barriers to entry.

AndrewCapon:
What we need is ```-outmoded-old-buggers``` for my code.

PR #278, opened 2020-03-06
Add -fno-exceptions to the toolchain

Daft-Freak:
Decreases code size, also libstdc++_nano is already compiled with this.

### DIFF
--- a/32blit.toolchain
+++ b/32blit.toolchain
@@ -36,7 +36,7 @@ set(CDC_FIFO_BUFFERS_EXT 4)
 set(COMMON_FLAGS "${MCU_FLAGS} -fsingle-precision-constant -Wall -fdata-sections -ffunction-sections -Wattributes -Wdouble-promotion -Werror=double-promotion -Wno-unused-variable -Wno-write-strings")
 
 set(CMAKE_C_FLAGS_INIT "${COMMON_FLAGS}")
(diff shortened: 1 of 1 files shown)


AndrewCapon:
Nice find, I can now build the firmware with debug on again.

AndrewCapon:
Interestingly a lot of the C functions have got much smaller, for example HAL_DMA_IRQHandler has gone from 3668 to 1848

Daft-Freak:
That's odd... those should be being compiled with the C compiler now, so this should have no effect there...

AndrewCapon:
Actually I think it was me deleting my build folder the grab the toolchain changes made the difference as doing the same to remove ```-fno-exceptions``` still builds fine and HAL_DMA_IRQHandler is exactly the same size.

I wonder why deleting the build folder is different to ```make clean; make```?

Daft-Freak:
Ah, was probably 3398068fe52e1badb6c4af166dd47664b248ae79 that fixed it. I think CMake only ever reads the toolchain file on the initial `cmake` run and never again, so the flags are baked into the makefiles... (Unlike the CMakeLists, which cause CMake to regenerate the makefiles when modified)

Gadgetoid:
Awesome! I might be able to play with #216 now without blowing up the build.

PR #281, opened 2020-03-06
Avoid initialising the API struct

Daft-Freak:
Possibly fixes the OpenOCD flashing problem.


Daft-Freak:
Includes bonus random LED colour feature...

Gadgetoid:
I've been wondering what the firmware could do with the LED to be useful- probably some secondary battery status.

Thanks!

Daft-Freak:
Probably need to make sure everything is getting initialised now...

PR #282, opened 2020-03-06
Surface  load from packed

ymauray:
Loading a packed image in a Surface resulted in said image to not be displayed properly. The exit condition for the loop was wrong, causing the last few bytes to not be loaded. On a big spritesheet that portion of the image would probably be transparent anyway, so it was not visible. On a small image, 10x10 pixels for example, it was very clear there was a problem !
I have tested with a number of small images, both on linux and on the device, and they display properly with this PR.


Daft-Freak:
The other way to fix this is to revert me breaking the tools (https://github.com/pimoroni/32blit-tools/commit/e3a9e8f89f7608e3b25514802610b1f8302d363b) (I have no preference). But there's also the issue that `b = *++bytes` reads 1 byte too far.

ymauray:
I’ve updated the PR with your fix for the buffer overflow

Gadgetoid:
Thank you! I think it's easier to forge ahead with slowly changing packed data to how we want it to work than moving back on ourselves. Onward and upward!

In theory if everyone is using the asset packer toolchain then they wont notice the underlying format change.

In theory...

PR #285, opened 2020-03-07
Paletted screen mode

Daft-Freak:
This was mostly a result of trying to get DOOM running again after the RGB888 framebuffer change. It seems to work but may need some adjusting. (The menu is a bit of a mess but there...)

Using a simple copy for flip as the DMA2D code had a few too many magic numbers for me to understand. Also, uses the last part of the LTDC section to give the heap some extra room. (Since the allocator would try to use it anyway.)

### DIFF
--- a/32blit-sdl/System.cpp
+++ b/32blit-sdl/System.cpp
@@ -17,6 +17,8 @@ blit::Surface __fb_hires((uint8_t *)framebuffer, blit::PixelFormat::RGB, blit::S
 blit::Surface __fb_hires_pal((uint8_t *)framebuffer, blit::PixelFormat::P, blit::Size(320, 240));
 blit::Surface __fb_lores((uint8_t *)framebuffer, blit::PixelFormat::RGB, blit::Size(160, 120));
 
+static blit::Pen palette[256];
+
 // blit debug callback
 void debug(std::string message) {
 	std::cout << message << std::endl;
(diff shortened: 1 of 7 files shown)


Gadgetoid:
Oh wow! Totally didn't expect you to tackle this. You've even used the LTDC colour LUTs!

I can imagine the menu is... uh... a little broken. I'm not quite sure how we'd resolve that since we presently have no blend modes that really understand paletted destinations and a `Pen` doesn't really have any knowledge of, nor is constricted by what's defined in the palette. Could be tricky to successfully draw an overlay menu in *true* paletted mode without, perhaps, having the user explicitly define what colours the menu can use. Food for thought!

Presumaby in DOOM you're just writing directly to the framebuffer and forgoing any of the engine's drawing functions?

Daft-Freak:
I'm not sure *I* expected me to tackle this. But after a quick hack it didn't seem so hard...

The only time I'm using any of the drawing functions is to display an error message, which is now doing `blit::screen.pen = blit::Pen(176);`. That happens to be `#FF0000` in DOOM's palette. (And I now know what the single-value Pen constructor is for...)

AndrewCapon:
But it but is there DMA2D stuff already done somewhere?

I started looking at it today and found nothing in master.

Daft-Freak:
There's this: https://github.com/pimoroni/32blit-beta/blob/master/32blit-stm32/Src/display.cpp#L64

AndrewCapon:
Thanks.

Daft-Freak:
I now have the menu looking okay by overriding the palette while it's open.

Gadgetoid:
I think we've resolved that *everything* will cause horrible conflicts with the menu stuff. Just a bridge we're going to have to cross at some point!

Overriding the palette seems like the sensible thing to do. Your handling of `Pen` is very graceful, I wonder if there's a way we can implement that into the engine. That said the number of overloads on Pen is sort of bonkers at this point since the first argument can be palette-index, mask alpha or red channel. Nice work!

PR #286, opened 2020-03-07
Fix opening files for writing through the constructor

Daft-Freak:
`blit::File f("blah", blit::OpenMode::write)` now doesn't ignore you and open the file for reading instead... (Oops, again.)


Gadgetoid:
Whoops! And thanks for the fix.

PR #288, opened 2020-03-09
More shared API data init

Daft-Freak:
Fixes the random LED colour after #281 and should also be useful for #269.


Gadgetoid:
Thank you. R.I.P. Happy Random LED Colour!

PR #289, opened 2020-03-09
Travis firmware size reporting

Daft-Freak:
Displays the change in the size of the firmware in PRs.

Example: https://github.com/Daft-Freak/32blit-beta/pull/1

(Discovered the script this is based on by accident while messing around with Travis/the GitHub API for something else...)

EDIT: Requires a `GITHUB_TOKEN` to be set with the `repo:status` scope.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -63,6 +63,9 @@ jobs:
           zip
       install:
         - python3 -m pip install bitstring construct
+      after_success:
+        - curl -O https://raw.githubusercontent.com/danvk/travis-weigh-in/adcc3a3e5bc16b360aa379d0fc23c3a6e1d9b21d/weigh_in.py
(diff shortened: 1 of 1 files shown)


Gadgetoid:
Thank you- I've granted the `repo:status` scope on our existing token, so it should hopefully work!

Daft-Freak:
Hmm, it seems to be working on master, but failing with `The GITHUB_TOKEN environment variable must be set.` on PRs...

Gadgetoid:
This sort of makes sense, but is somewhat frustrating. Something to do with the scope in which PR integrations are run?

Daft-Freak:
Ah, Travis doesn't send secure env vars to external PR builds since someone could get access to the token by submitting a PR that does something like `echo $GITHUB_TOKEN`...

PR #293, opened 2020-03-11
Asset tools in Travis/port platformer

Daft-Freak:
Gets the asset tools working on Emscripten(python too old)/STM32(wasn't installed) and uses them for the sprites in the platformer example.


Daft-Freak:
The asset packer commit is actually after the Travis updates, GitHub is just showing them in the wrong order (probably because I reordered them).

Gadgetoid:
Awesome! Thank you. This also serves as a defacto sort-of test for the asset packer functionality.

PR #294, opened 2020-03-11
Assets/memory as files

Daft-Freak:
Adds a way to use embedded assets (or any memory) as if they were files. Useful for porting things that expect data to be in files.

One minor issue is that:
```
File::add_buffer_file("some/dir/file.thing", ...);
directory_exists("some/dir");
```

will return false.


Gadgetoid:
This coupled with palette mode should make some ports a (comparitive) breeze- really nice work, thank you!

PR #295, opened 2020-03-11
JPEG example

Daft-Freak:
Simple JPEG example. Also some fixes. WIP as it depends on #293 to build properly on Travis. Also the drawing to the screen is currently a memcpy.


Gadgetoid:
Thank you!

PR #296, opened 2020-03-12
Use a newer toolchain for Travis STM32 builds

Daft-Freak:
Uses a PPA containing GCC/newlib backported from Ubuntu 19.10. I'm not planning on changing anything in there so it should be stable. Also adds the PPA to the WSL instructions.

The main reason for this is that Ubuntu 18.04 has GCC 6.3 and GCC 7 is required for most of C++17. 
https://gcc.gnu.org/projects/cxx-status.html#cxx17
https://gcc.gnu.org/onlinedocs/libstdc++/manual/status.html#status.iso.2017


Gadgetoid:
Thank you- I've got no qualms about using your PPA for the time being. This is a problem that will no doubt solve itself over time anyway.

PR #297, opened 2020-03-14
Cdc to lib pr

AndrewCapon:
This moves the CDC stuff into the 32blit library so it can be used by user apps as well as the firmware.


Gadgetoid:
We might have to do something with the firmware API to make CDC work across the boundary when running stock firmware and an app that isn't linked against the HAL. I think bad things will happen if user code and firmware try to access/control it at the same time.

Gadgetoid:
This has become hopelessly conflicted and probably incompatible with the way 32blit has moved since! Sorry.

Closing for now, but probably deserves some attention in future. Though I think Multiplayer probably covers *most* of the uses people might have for CDC Serial?

Daft-Freak:
The multiplayer stuff is basically built on top of a "USER" CDC handler, so I guess that gives games one handler to play with...

PR #298, opened 2020-03-15
contributing doc - a simple git guide

mgarcia-org:
I'm a git noob, maybe this will help others, at least get the ball rolling.
Mike.


Daft-Freak:
A few suggestions:
- I wouldn't recommend `git add .` or `git commit -a`, it's too easy to end up adding a bunch of files you don't want. (Also `commit -a` is redundant if you've already done `add .`)
- You should switch branch `git checkout -b branch-name` before committing anything (modifying your fork's master either makes updating master a pain or causes your PRs to be full of merge commits)
- Doing `git remote add upstream https://github.com/pimoroni/32blit-beta` is useful as it means you can do `git checkout master` `git pull upstream` `git checkout -b new-branch` to start a new PR branch (assuming master is clean).
- VS Code users will want to do `git config --global core.editor "code --wait"`
- Add some ``s around your commands for formatting.

Also, `sudo pip install` is a bad idea (it's basically "run random files off the internet as root").

Gadgetoid:
> Also, `sudo pip install` is a bad idea (it's basically "run random files off the internet as root").

Ooof yes. This is a bad habit that got trained into me by the Raspberry Pi- but, yeah, `setup.py` can import subprocess and do whatever the heck it wants to your system in theory.

mgarcia-org:
Thanks for the feedback, I'll processes it and try again..
I think cmake/make might also need an intro doc.

Gadgetoid:
You're right- I could have done with an intro to CMake when I got started with it for this project!

PR #299, opened 2020-03-15
Clearing probes fixes.

AndrewCapon:
Added clear_all_probes() to reset all probes to init state.

Some fixes to actually clear the probes correctly.

### DIFF
--- a/32blit/engine/profiler.cpp
+++ b/32blit/engine/profiler.cpp
@@ -90,20 +90,20 @@ void Profiler::remove_probe(ProfilerProbe *pProbe)
 
 void Profiler::start_all_probes()
 {
-	for(ProfilerProbesIter iP = m_probes.begin(); iP != m_probes.end(); iP++)
-		(*iP)->start();
+  for(ProfilerProbe *pProbe : m_probes)
+    pProbe->start();
 }
 
 void Profiler::clear_all_probes()
 {
-	for(ProfilerProbesIter iP = m_probes.begin(); iP != m_probes.end(); iP++)
-		(*iP)->clear();
+  for(ProfilerProbe *pProbe : m_probes)
+    pProbe->clear();
 }
+
 void Profiler::log_probes()
 {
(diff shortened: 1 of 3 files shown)


Daft-Freak:
Not entirely related to this PR, but clang-tidy wants the shiny new for loops:

`for(ProfilerProbesIter iP = m_probes.begin(); iP != m_probes.end(); iP++)` -> 
`for(auto probe : m_probes)`

(It's one of the warnings I haven't got around to fixing yet.)

AndrewCapon:
Does it always complain about iterator loops or only the case where you iterate through all?

Daft-Freak:
For a range, it usually suggests using `auto` instead of the iterator type. You can check the warnings in the output for the Travis Linux build: https://travis-ci.com/github/pimoroni/32blit-beta/jobs/298290759

AndrewCapon:
Sorry, I can't bring myself to use "auto"!

AndrewCapon:
How do you get the clang-tidy stuff when building locally?

Daft-Freak:
Assuming a `clang-tidy` in PATH somewhere
```
cmake "-DCMAKE_CXX_CLANG_TIDY=clang-tidy;-header-filter=32blit;-checks=performance-*,portability-*,modernize-*,-modernize-use-trailing-return-type,-modernize-avoid-c-arrays,-modernize-use-nodiscard" [other opts]
```
(What Travis is doing) or

```json
    "cmake.configureSettings": {
        "CMAKE_CXX_CLANG_TIDY": ["clang-tidy", "-header-filter=32blit", "-checks=performance-*,portability-*,modernize-*,-modernize-use-trailing-return-type,-modernize-avoid-c-arrays,-modernize-use-nodiscard"]
    },
```
With VS Code/CMake tools. There are a few defaults disabled due to being incredibly noisy. (And I don't think `auto f() -> int;` is an improvement on `int f();`...)

It will fail to build if you use it on an STM32 build though (it dies on the few `.c` files that actually contain C++)

AndrewCapon:
Thanks, I will start running everything through this to tidy up before committing.

PR #300, opened 2020-03-17
Vs sdl spaces in filename

illbewithee:
If the repo has a parent folder of one with a space, the command is incorrectly formed


Daft-Freak:
I was going to suggest fixing it to also copy SDL2_image, but you beat me to it. :smile:

Gadgetoid:
Oof. Good catch, thank you!

PR #302, opened 2020-03-18
More accutate audio timer

Daft-Freak:
Was ~21633Hz, should now be ~22048Hz (I think...). Seems to be an improvement in music-player and causes something else I'm doing to have less timing issues.


Gadgetoid:
This lines up with our findings via oscilloscope, but I don't think we properly double-checked our maths on the timer config. For some reason I was convinced the chip was running slower than it should have been. Ha!

Thanks for finding the magic numbers to get this right!

PR #303, opened 2020-03-19
Adjust readme requirements

Daft-Freak:
Python3 / the tools are now required and DFU stuff is only needed if you're working with the firmware.

(WSL/Ubuntu VM requirement probably needs adjusting, but there aren't instructions for setting up a native toolchain... yet)


Gadgetoid:
Good catch, thank you!

PR #304, opened 2020-03-23
Fix multi-block SD card writes

Daft-Freak:
it would fail after writing all the data as `resp` is uninitialised when sending a stop token.

(More bugs in the SD card code, yay)


Gadgetoid:
I think @lowfatcode might have run into this when playing with the firmware. Good catch! Thank you. And colour me surprised 🙄

Daft-Freak:
Fun one to track down, kept thinking "but saving works in DOOM...", guess it only does small writes...

PR #305, opened 2020-03-28
Toolchain tweaks

Daft-Freak:
Sets a couple of extra variables in the toolchain to help with configuring out-of-tree projects.

### DIFF
--- a/template/CMakeLists.txt
+++ b/template/CMakeLists.txt
@@ -1,5 +1,5 @@
 cmake_minimum_required(VERSION 3.8)
-set(32BLIT_PATH "../" CACHE PATH "Path to 32blit.cmake")
 project (game)
+set(32BLIT_PATH "../" CACHE PATH "Path to 32blit.cmake")
 include (${32BLIT_PATH}/32blit.cmake)
 blit_executable (game game.cpp game.hpp)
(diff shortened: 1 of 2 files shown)


template/CMakeLists.txt
```diff
@@ -1,5 +1,5 @@
 cmake_minimum_required(VERSION 3.8)
-set(32BLIT_PATH "../" CACHE PATH "Path to 32blit.cmake")
 project (game)
+set(32BLIT_PATH "../" CACHE PATH "Path to 32blit.cmake")
```
Daft-Freak, line 3: In case anyone is wondering, this change is because the toolchain isn't parsed until the `project()` call... Also, this works because `set(... CACHE ...)` doesn't override existing variables.

Daft-Freak:
Forgot to mention, the second patch is so that you can do:
```sh
cmake .. -DCMAKE_TOOLCHAIN_FILE=/path/to/32blit/repo/32blit.toolchain
```
instead of:
```sh
cmake .. -D32BLIT_PATH="/path/to/32blit/repo" -DCMAKE_TOOLCHAIN_FILE=/path/to/32blit/repo/32blit.toolchain
```
for out-of-tree device builds.

Gadgetoid:
Looks reasonable- and extremely useful! Thank you.

PR #306, opened 2020-03-31
Windows 99% upload bug

illbewithee:


Gadgetoid:
Thank you! I'm a little behind on this one- can anyone elaborate what the problem was and if this is confirmed to fix it?

illbewithee:
I did fix it for me - been working well since. The two issues I found were a) CloseCom() not resetting the handle so subsequent CloseComm() used an already-closed handle, and b) WriteFile() managing to be synchronous so no longer stalled waiting for a completion that isn't valid (see docs "WriteFile *may* return before the write operation is complete", my emphasis) ;)

Gadgetoid:
Brilliant! Thank you for taking the time to make these tweaks and clue my slow brain in on them too!

PR #307, opened 2020-04-01
Increase version requirements

Daft-Freak:
Effectively already the required versions, so this just gives more obvious errors.


Gadgetoid:
Thank you!

PR #308, opened 2020-04-22
File rename/remove

Daft-Freak:
File API again! Using these to help with save corruption in my emulator...


Gadgetoid:
Wow, it's getting pretty comprehensive now! Before long we'll have a port of Midnight Commander or something! Thank you.

PR #309, opened 2020-04-27
"Fix" ~1000 warnings

Daft-Freak:
`#define` out `register` when compiling `stm32h7xx_ll_adc.h` as C++. This file gets included indirectly by anything using the HAL, which causes a total of over 1000 warnings. This is a little annoying.

This should be entirely safe as GCC only obeys `register` when not optimising (or some other cases that don't apply here, see: https://gcc.gnu.org/onlinedocs/gcc/Hints-implementation.html)


Gadgetoid:
Thank you!

Issue #63, opened 2020-01-13
Improve button handling in the API


lenardg:
Currently the API provides us with button statuses only (is it pressed or is it released). The demo game scrolly-tile does some bitfield magic to keep track of what is actually done with the buttons - and not just if it is held or not. 

Relevant code:
``` C++
	uint16_t changed = blit::buttons ^ last_buttons;
	uint16_t pressed = changed & blit::buttons;
	uint16_t released = changed & ~blit::buttons;
```

It would be great if this same functionality would be part of the engine. 

Instead of just `blit::buttons` it could be a struct of some kind, where you would have three members:

* `status` - to check the status of a single button, like what blit::buttons is now.
* `pressed` - that would tell which button was pressed since last time
* `released` - that would tell which button was released since last time

ALthough it might be a little bit long to write, it would increase clarity.

For example: `if ( blit::buttons::pressed & blit::button::A )`

ali1234:
You can avoid the complement operation:

    uint16_t released = changed & last_buttons;

Daft-Freak:
I think I did (quite close to) this in #310?

PR #310, opened 2020-04-28
Button API improvements

Daft-Freak:
Roughly what was suggested in #63.


Gadgetoid:
This is pretty slick, thank you- really shows how much crusty boilerplate you can cut out. I'm a little apprehensive to merge this one until I'm fully up to speed, but I think it needs to be done sooner rather than later- suspect it might be a good idea to tag a release afterwards so there's somewhere to communicate that the button API has changed.

Daft-Freak:
It shouldn't break anything... 

Probably need a release anyway... [there's some big stuff already](https://github.com/pimoroni/32blit-beta/compare/v0.0.6...master)

PR #311, opened 2020-04-29
Cleanup declarations in 32blit.h

Daft-Freak:
These either don't exist, or are also declared where they're used.


Gadgetoid:
No surprised that this got into a bit of a mess- good catch and thanks for cleaning it up!

PR #312, opened 2020-05-28
Fix VS build

Daft-Freak:
Travis build broke on an unrelated change. Maybe an update?


Gadgetoid:
Thank you! I'm so woefully out of date with all of this- I'm worried whatever I touch will inevitably break something!

PR #314, opened 2020-06-03
Fixed: Game template did not compile with VS2019

lenardg:
The game template did not compile under VS2019. The other projects were using C++17 settings. I changed this to the game template as well.


lenardg:
Added another commit because the Voxel example also did not compile.

Daft-Freak:
Hah, I didn't find/replace hard enough.

Slightly confused as to how the C++ standard affects `std::string` being defined though... (Same Travis build error being fixed by #312).

Gadgetoid:
Thank you!

PR #317, opened 2020-06-12
A few Travis changes

Daft-Freak:
Updates the STM32 build to the new Ubuntu to get a better compiler without needing the PPA. Also splits out the SDL install scripts as they're getting a bit big. (I _may_ have something that requires adding more to them somewhere...)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -100,16 +100,7 @@ jobs:
           python3-setuptools
           zip
       before_script:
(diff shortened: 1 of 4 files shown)


Gadgetoid:
Oh nice! All those wrinkles melt away if you just wait long enough 😆 

Thanks for making a PR for this!

PR #318, opened 2020-06-12
WIP: MP3 playback

Daft-Freak:
I remember it being suggested that the performance of my MP3 demo was good enough that it could be a core feature. The decoder has got a little faster since then, and we now have embedded assets that remove most of the overhead... so here's the MP3 decoding feature!

The WIP is mostly the example, it currently lacks a file to play. (And is not the best looking example...)


Gadgetoid:
Any of my terrible soundcloud music could be volunteered to be a test file, albeit it might be a bit long-winded. Maybe the Bilge Tank theme? Haha. Otherwise I'm sure I can throw something together quickly that we can license permissively.

Gadgetoid:
We shall never speak of my github horror again.

🤣 

Thank you. Woohoo!

PR #319, opened 2020-06-19
Enable GitHub super-linter

Gadgetoid:
Attempt to enable the super-linter across the whole codebase. I don't actually know if the Action will run in this PR, or if I have to merge first. This could get messy.


Gadgetoid:
Okay, I can't seem to get this thing to fine any files to lint- guess it's not *that* super 😆

PR #320, opened 2020-07-09
WIP: Minimal "multiplayer" support

Daft-Freak:
Something I was experimenting with last month, still not quite finished.

Adds a basic `send_message` function and a `message_received` callback as a minimal way of sending user messages. Uses CDC on device and sockets (through SDL_net) for SDL.

TODO:
- ~Device host mode (WIP on another branch, currently unstable)~
- ~SDL host mode? (How to decide who is host, device is automatic)~
- ~Some way to specify address for SDL (currently hardcoded as localhost)~
- ~An "is connected" query?~
- ~Attempting to send a message on device with nothing connected locks up~ (mostly fixed I think)

Tested so far with this script as a server/bridge:
```python
import select
import serial
import serial.tools.list_ports
import socket
import struct

port = None
serial_port = None

for info in serial.tools.list_ports.comports():
  if info.vid == 0x0483 and info.pid == 0x5740:
    port = info.device

if port:
  serial_port = serial.Serial(port, 115200)
  serial_port.reset_output_buffer()

# net
server = socket.create_server(("localhost", 0x32B1))
inputs = [server]
clients = []

while True:
  readable, _, exceptional = select.select(inputs, [], inputs, 0)

  for s in readable:
      if s is server:
        # new connection
        client, addr = server.accept()
        inputs.append(client)
        clients.append(client)
      else:
        data = s.recv(1024)
        if data:
          if serial_port:
            serial_port.write(data)

          for other in clients:
            if other is s:
              continue

            other.sendall(data) 
        else:
          inputs.remove(s)
          clients.remove(s)
          s.close()

  if serial_port and serial_port.in_waiting:
    data = serial_port.read(serial_port.in_waiting)
    # echo
    #serial_port.write(data)
    #print(data)

    for s in clients:
      s.sendall(data)
```


Gadgetoid:
I wonder- electrically- what cable we'll need to make this happen. IIRC I ensured the correct sense pin was broken out to the USB port, but beyond that I don't know the specifics of making an out-of-spec microUSB to microUSB peer to peer cable. It may be as simple as pulling the ID pin at one end and detecting that in software. I have "USB" (not actually USB since the D+ and D- lines are not twisted) cables for Flotilla and Mote which would potentially work with manual server/client setting.

Daft-Freak:
Hmm, I expect some cable hacking... Need working host mode first though, not looked at that since I had no battery until recently. 

Though, not so easy to test with one device... Which is how I ended up with the bridge script.

Daft-Freak:
Many hacks later...
IMG_20200820_145410

Need to update this branch a bit.

Daft-Freak:
Hmm, so the ID pin detection (mostly) works, but I couldn't get the interrupt for it to work (enabling it just caused the OTG interrupt handler to get called constantly with no unmasked bits set). Also, host mode has some annoying bugs (NAK handling https://github.com/Daft-Freak/32blit-beta/commit/c046b1b44c0e2c7b185fc969ef0d3502a6bafe9f, not freeing pipes until it runs out and hard faults https://github.com/Daft-Freak/32blit-beta/commit/ac3aabb46415292c7e96d88ac39aa43900732c6f).

I think this would work with a standard OTG cable, but since the USB port doesn't supply power I had to do some cable splicing.

The host stuff is still sitting on my `uo-firmware-stuff` branch.

ali1234:
The spec says OTG cables should have micro A at the host end and micro B at the device end, with the A ID pin grounded and the B ID pin floating. These cables are available in a variety of lengths. The B plug fits into B and A/B sockets and the A plug will only fit A/B connectors (A sockets don't exist.) The plugs mirror their full size counterparts: A is rectangular while B is a rectangle with two corners cut off to make a "D" shape.

The spec also says devices that support both roles should have a micro A/B socket, but hardly anyone actually follows this (including 32blit). Everyone just uses micro B sockets, So B to B OTG cables are also widely available. Sometimes these cables have both ID pins grounded, on the assumption that only one of the devices will support host mode and the other will ignore the ID pin. For example, when connecting a hard drive to an Android phone, or an Arduino to a Raspberry Pi Zero. Adafruit sells a cable like this. This type of cable is usually too short to be used as a multiplayer link cable.

So for multiplayer you would need a B to B cable (out-of-spec) with one ID pin grounded and the other floating (in-spec), and a decent length. This might be tricky to obtain as most sellers simply aren't familiar with the technicalities. Alternatively change the board to have a micro A/B socket, and then you can use a cable with the right plugs, which will certainly have correctly wired ID pins. It should be possible to get an A/B socket with the same footprint as the existing B socket, so no board changes required.

For reference, the Nucleo STM32H743ZI development board has a micro B socket on the programmer (because it is device only) and a micro A/B socket on main board, and is the only product I have seen that gets this right.

Daft-Freak:
Ah yes, the common "OTG cables" are a bit non-standard... guess "standard" was the wrong word there.

Instead of the possibly difficult task of finding an appropriate cable, there's always the "lazy" method of just chaining one of those micro B to A OTG cables with a regular A to micro B cable... Which is what I was doing, except I had to splice some extra power wires in due to lack of batteries.

... and I entirely forgot about VBUS sensing being a thing because of that...

Gadgetoid:
Cables are a problem we can solve by throwing money at a factory in China. They'll make whatever the heck we ask for- including the unholy monster that is Flotilla cables- so if we can make a case for multiplayer being actually usable then the hardware side should be easy to fix.

Epic work here, @Daft-Freak!

Daft-Freak:
There, everything after "Immediately check for packet data " is the current, slightly messy host mode code.

Daft-Freak:
With a second battery and VBUS sensing disabled...
```diff
diff --git a/32blit-stm32/Src/usbd_conf.c b/32blit-stm32/Src/usbd_conf.c
index 5b468914..34fee79f 100644
--- a/32blit-stm32/Src/usbd_conf.c
+++ b/32blit-stm32/Src/usbd_conf.c
@@ -320,7 +320,7 @@ USBD_StatusTypeDef USBD_LL_Init(USBD_HandleTypeDef *pdev)
   hpcd_USB_OTG_HS.Init.low_power_enable = DISABLE;
   hpcd_USB_OTG_HS.Init.lpm_enable = DISABLE;
   hpcd_USB_OTG_HS.Init.battery_charging_enable = ENABLE;
-  hpcd_USB_OTG_HS.Init.vbus_sensing_enable = ENABLE;
+  hpcd_USB_OTG_HS.Init.vbus_sensing_enable = DISABLE;
   hpcd_USB_OTG_HS.Init.use_dedicated_ep1 = DISABLE;
   hpcd_USB_OTG_HS.Init.use_external_vbus = DISABLE;
   if (HAL_PCD_Init(&hpcd_USB_OTG_HS) != HAL_OK)

```

... I got it to work with a "standard" OTG adapter... but that would cause hotplug detection to not work.

Gadgetoid:
Can confirm, with this "fix" it works using my Amazon-acquired "link" cable (a direct microUSB to microUSB abomination with a red end to denote "host").

I can plug the cable in while the app is booted and it's still fine, still pings on my PC.

I wonder if we can change this mode on the fly and just add it to the firmware menu? "USB mode: Hotplug Detect / Multiplayer"

Daft-Freak:
Now with metadata, SDL host support and `is_multiplayer_connected`. Might need something like `multiplayer_enable()` to adjust the USB config and also to not have SDL games always listening on a socket.

Gadgetoid:
I'm going to have to get myself a second blit that isn't a broken mess!

I wonder if one of those tunnelling apps could make this work over WAN without any fuss 😆

Daft-Freak:
Before I forget where this is...

Host mode seems to only work right the first time it's switched to, possibly some cleanup code missing. Device mode doesn't detect removal after the first connect (expected as no VBUS/seising disabled). Occasionally, plugging in the cable causes a reset.

If you have the cable connected before launching the game it usually works, also the SDL side works great. (Might be useful if it handled connecting to devices over serial...)

Gadgetoid:
Happy days it's finally merged! :laughing: 

Thank you, right now I'm somewhat lacking another functional blit but I'm keen to see if someone throws together an online multiplayer lobby. Could be fun!

PR #322, opened 2020-07-29
Fix using the second half of the flash

Daft-Freak:
Switches to 32bit addresses as 24 is one too short. Tested with a modified bin flashed to the very end of the flash. (And a WIP "installed" list).


Gadgetoid:
Whoops! I think we'd always skirted under the size required to run into this. Good catch! Thank you.

PR #323, opened 2020-08-14
Make SD cards fast

### LINKED ISSUE #1: 32blit-stm32 - Investigate SD-card IO speed
SD card throughput has been less than stellar, and while we're using a 1-bit bus I don't believe we're quite as close to the theoretical maximum speed as it's possible to get. Looking over the SD card code reveals codepaths and potential inefficiencies that could cause slowdown, I've fixed some of these with significant resulting gains in performance. I think it's worth trying to find more opportunity to eke throughput out of the SD card.

Daft-Freak:
Replaces the SPI functions for accessing the SD card with simpler more direct versions. Improves read speed by up to 7x on my new card and ~2x on my old one.

Possibly needs some testing on more cards, but doesn't cause the cards I tested with to stop working. I've also tested that mass storage still works, since it was previously broken by changes here. (It isn't any faster though...)

[test bins](https://github.com/pimoroni/32blit-beta/files/5074139/sd-test.zip)

Before/After:
comparison

Fixes #1?

### DIFF
--- a/32blit-stm32/Src/fatfs_sd.c
+++ b/32blit-stm32/Src/fatfs_sd.c
@@ -191,9 +191,28 @@ static bool SD_RxDataBlock(BYTE *buff, UINT len)
 	/* receive data */
 	if(len > 16)
 	{
-		// SD card expects 0xff dummy bytes for reads
-		memset(buff, 0xFF, len);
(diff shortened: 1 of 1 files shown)


Gadgetoid:
Tested against my SD card- it works! I'm happy to merge this with crossed fingers and see how it goes.

PR #326, opened 2020-08-18
Screenshots

Daft-Freak:
Got tired of taking photos of the device, so added a screenshot button to the menu. Also restores rendering the firmware behind the menu since I moved the button handling recently.

Screenshots are saved with a new method added to `Surface` that saves the contents to an uncompressed .bmp file. I've tested it with RGB/paletted images. ~230k for a hires screenshot.


Gadgetoid:
I'm going to have to try this out! Thank you.

PR #328, opened 2020-08-20
Enable LTO for 32blit-stm32

Daft-Freak:
Makes the firmware build ~9.6k smaller. I was getting quite close to the limit when adding USB host code...

(The startup file is moved to the top of the source list as otherwise GCC optimises out all of the interrupt handlers...)


Gadgetoid:
I'm going to place a bet that this will lead to some weird and interesting errors about 6 months from now... let's see 😆 

And, thank you!

PR #330, opened 2020-08-21
Another tiny documentation tweak

ahnlak:
Added note to include 32BLIT_PATH in cmake, in the Emscripten build instructions so that template-based projects build.


Gadgetoid:
Ooh, good catch!

PR #331, opened 2020-08-25
Faster blends/fills (not mine)

Daft-Freak:
This is just the `blend-fast-alpha` branch with two minor fixes. Tested a few examples and nothing looked broken.

Reduces the render time in my emulator by ~1.7ms. Before this, clearing the screen took nearly 4x as long as actually drawing anything...


Gadgetoid:
I can't remember why these never made it into mainline, but I suppose we can try 'em on for size! Thank you.

PR #336, opened 2020-08-28
flash-loader -> firmware

Daft-Freak:
No new stuff in here, just moving code around.


Gadgetoid:
This is probably long, long overdue 😆 

Thank you!

PR #342, opened 2020-09-10
Sort firmware file list on init

Daft-Freak:
Just a minor annoyance... Also fixes the opt level to what I thought it was, which avoids this change increasing size by a few kB...


Gadgetoid:
Haha! Well worth it. Thank you.

PR #343, opened 2020-09-11
Remove clip param to Surface::text

Daft-Freak:
Use the surface clip rect instead. Removes the default clipping from the `Rect` version of `text` but makes it more consistent with the rest of the API. (Somehow missed the surface clip when adding this...)


Gadgetoid:
This makes a lot of sense. I've no doubt the API will be rife with things like this. It needs a solid sprint to hammer out the final shape. Thank you!

PR #346, opened 2020-09-14
Tweak docs for installing tools

Daft-Freak:
Adds installing the tools to the OS-specific pages (except Windows/VS, which doesn't currently cover installing Python). Also adds some basic instructions on how to use the packer when ~duplicating build confi~ using a VS project.


Gadgetoid:
Thank you!

PR #347, opened 2020-09-15
Define a symbol for the end of (used) flash

Daft-Freak:
Useful for appending data post-build. (Like I did with my emulator)


Gadgetoid:
Ah! Handy! Thank you.

PR #348, opened 2020-09-15
Asset packer for flight example

Daft-Freak:
I was trying to fix the 2-3 minute build time I have for the flight example (only happens for debug builds with the device toolchain). This didn't fix it, but it's still a cleanup... Also gets the example drawing something again.

I think the huge compile times I'm getting are due to the giant `vector`s. Also, there seems to be two classes for tilemaps? (`Map`/`TileMap`) The other one uses pointers for its data... (But they have different features)


Gadgetoid:
Not sure the what's/why's of the Map/TileMap, it's been a while since we discussed them. /me pokes @lowfatcode

Gadgetoid:
In the interim there's no reason why this should not be merged. Thank you.

PR #349, opened 2020-09-17
Expanded Visual Studio asset stream doc

shane-powell:


Daft-Freak:
Whoops. Yeah, actually compiling the file does help.

Gadgetoid:
Thanks for bouncing off this and updating the docs accordingly! It's much appreciated.

Issue #97, opened 2020-01-31
Mode7 exception in Emscripten builds


Daft-Freak:
In `screen_to_world` : 
`float distance = ((far - near) / float(s.y - viewport.y)) + near;` is inf for the first line. Which results in nan getting returned. That eventually gets converted to a uint16_t in `mipmap_texture_span`, resulting in `RuntimeError: invalid conversion to integer`.

Example: https://cupboard.daftgames.net/32blit/shell.html?app=flight

Gadgetoid:
Mode7 is currently horribly broken, so I've assigned this to Jon for consideration when he gets it back up and running.

Daft-Freak:
Doesn't throw an exception any more, but the first line is still broken.

PR #352, opened 2020-09-18
Switch to official Emscripten Docker image

Daft-Freak:
... and update to the latest release. Includes switching to the "upstream" backend, which seems to build a bit faster (~2 mins on travis). Also should include the workaround for audio.

Bonus: flight and the `perspective` effect in tilemap-test now work (#97/#340). (The nans/infs are still there, they just don't cause exceptions any more...)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,7 +14,7 @@ cache:
 jobs:
   include:
     - name: "Emscripten"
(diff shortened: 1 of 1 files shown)


Gadgetoid:
Awesome, thanks. And good news regarding flight and tilemap!

PR #353, opened 2020-09-18
Deploy Emscripten examples

Daft-Freak:
Pushes the Emscripten compiled examples to pages, also includes the API docs as they're currently the content of https://pimoroni.github.io/32blit-beta/ (also it was easy). May want to add `on: tags` at some point.

(The accelerated renderer commit is here mostly because WebGL is currently refusing to work in my Firefox...)


Gadgetoid:
Thank you- here goes nothing!

PR #355, opened 2020-09-21
Support SDL2_image in WSL/MinGw + docs

Gadgetoid:
Instructions for installing SDL2_image.dll and other dev files for MinGW WSL build. Plus support for copying SDL2_image.dll added to the cmake toolchain file.

Perhaps the SDL2 compile instructions should be substituted for a precompiled binary package and a bootstrap script to do all of this automagically.

### DIFF
--- a/docs/Windows-WSL.md
+++ b/docs/Windows-WSL.md
@@ -52,14 +52,14 @@ You will need to cross-compile SDL2 for MinGW and install both it, and SDL2-imag
 
 ### Installing SDL2 & SDL2_image
 
-This will install the SDL2 development headers and libraries into `/usr/local/cross-tools/x86_64-w64-mingw32/`.
+This will install the SDL2 64bit mingw development headers and libraries into `/opt/local/x86_64-w64-mingw32/`.
 
-If you use a different directory then you will have to supply the SDL2 dir to the `cmake` command below using `-DSDL2_DIR`. For example: `cmake .. -DCMAKE_TOOLCHAIN_FILE=../mingw.toolchain -DSDL2_DIR=/usr/local/cross-tools/x86_64-w64-mingw32/lib/cmake/SDL2`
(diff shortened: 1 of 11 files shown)


docs/Windows-WSL.md
```diff
@@ -57,6 +71,20 @@ sudo make install
...
 This will install the SDL2 development headers and libraries into `/usr/local/cross-tools/x86_64-w64-mingw32/` if you use a different directory then you will have to supply the SDL2 dir to the `cmake` command below using `-DSDL2_DIR=/usr/local/cross-tools/x86_64-w64-mingw32/lib/cmake/SDL2`
 
+### Installing SLD2-Image
+
+SDL2 Image can be a little trickier to build from source, so we'll grab the pre-compiled mingw packages and install them alongside SDL2 in `/usr/local/cross-tools/x86_64-w64-mingw32/`:
+
+```shell
+wget https://www.libsdl.org/projects/SDL_image/release/SDL2_image-devel-2.0.5-mingw.tar.gz
+tar xzf SDL2_image-devel-2.0.5-mingw.tar.gz
+sudo cp -r bin /usr/local/cross-tools/x86_64-w64-mingw32/
```
Daft-Freak, line 81: `sudo cp -r x86_64-w64-mingw32 /usr/local/cross-tools/` ?

Gadgetoid: Ah you're right- I saw the additional files in the root directory and switched- unecessarily- to copying only the bits I wanted. But they're in a subdirectory anyway.

Daft-Freak:
Is it worth mentioning that MinGW builds are also significantly larger than VS builds?

Gadgetoid:
Probably. MinGW builds- on balance- are pretty horrible and probably only necessary if you're intending to ship from Windows -> Linux (though we can provide a CI file that will make your GitHub repository CI actually build the release binaries).

This is one of those "on the off chance you're mad enough to want to do this" cases.

PR #356, opened 2020-09-21
Attempt to document some of my API additions

Daft-Freak:
I think I wrote these while waiting for Travis builds... Adds at least SOME documentation to the file/text/font/jpeg APIs.


Gadgetoid:
SOME documentation is better than none! That's for sure. Thank you. Maybe API doc auto-gen wont be a folly after all.

Issue #350, opened 2020-09-17
Add common easing functions and cool demo to browse them


Gadgetoid:
I think we should implement most, if not all, of the easing functions demonstrated here - https://easings.net/

Additionally a demo along the same lines to show easing functions in action would be cool.

PR #358, opened 2020-09-23
Fix tween functions

Daft-Freak:
Was looking at #350 and found that most of the existing functions didn't handle start not being 0, `ease_in_out_quad` seemed especially broken.

broken

fixed


Gadgetoid:
Yikes. Yes, that would appear to be hilariously broken. Thank you!

PR #361, opened 2020-09-24
Clamp joystick range

Gadgetoid:
Joystick was hitting cardinal directions fine, but diagonals overflowed the circular range.

This change normalises the joystick if its vector length exceeds 1. Clamping it to a circle of r=1.

### DIFF
--- a/32blit-stm32/Src/32blit.cpp
+++ b/32blit-stm32/Src/32blit.cpp
@@ -804,6 +804,10 @@ void blit_process_input() {
   }
   blit::joystick.y = -joystick_y / 7168.0f;
 
+  if(blit::joystick.length() > 1.0){
+    blit::joystick.normalize();
+  }
+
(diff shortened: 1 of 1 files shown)


Daft-Freak:
SDL might need clamping as well. The gamepad I have is definitely (1, 1) at the corners.

Gadgetoid:
That's weird- I wonder if being (1,1) at the corners is normal.

It makes virtually no sense in some cases, but an awful lot in others.

Consider a case where you're accelerating *and* steering with a joystick- if you hold it forward and then "steer" then if it can't go to (1, 1) it will decrease the Y axis value and decelerate. This would apply to 3D games, vehicles, etc.

But if you're controlling something in a top down view and you want the joystick input mapped 1:1 to motion then the ability to hit (1, 1) or (-1, -1) would cause your input to be "faster" at the diagonals than it is directly.

Perhaps it is our responsibility to supply the full "square" range of motion, and the developer to then clamp that to a radial input if they desire?

PR #363, opened 2020-09-24
Increase width of circles by one

### LINKED ISSUE #31: blit::circle is 1 pixel too narrow
This one is definitely wrong. Code:

```
    blit::fb.pen(blit::rgba(0, 0, 0));
    blit::fb.circle(blit::point(4, 4), 3);
```

Result:

circle

Daft-Freak:
This makes the width match the height (and also matches what the original unfilled algorithm would do). It does mean that the circle size is actually r * 2 + 1 though...

Result:
circles

Should fix #31

### DIFF
--- a/32blit/graphics/primitive.cpp
+++ b/32blit/graphics/primitive.cpp
@@ -115,16 +115,16 @@ namespace blit {
 
       err += y; y++; err += y;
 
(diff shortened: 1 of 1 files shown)


Gadgetoid:
Oooh excellent stuff, thank you!

PR #367, opened 2020-09-26
Replace double-precision logs in packed sprite load

Daft-Freak:
Went a step further than the `log2f` patch, this one is two instructions (on ARM). Saves nearly 3k of flash. Also clamps the minimum bit size to 1, which fixes images with one colour.

(Wasn't sure where to put the helper function, so just threw it at the top of the file)


Gadgetoid:
Dang, that's a heck of a difference. Thank you!

PR #368, opened 2020-09-26
Fix "No Files Found" message

Daft-Freak:
Back to being readable. Also move it in a bit.


Gadgetoid:
Haha! Whoops. And thanks.

PR #370, opened 2020-09-27
Flash game listing and metadata handling

Daft-Freak:
Changes for listing games in flash and using the game metadata. Also some cleanup stuff. Has a few unfinished bits:

- ~Doesn't unload metadata when a game has none~
- A block is a 64k flash sector, could be anywhere down to 4k with subsector erases
- The CMake function for metadata only detects changes on the yaml, not the images
- ~There's a TODO in there for skipping to the end of a game while scanning flash~
- ~After you flash a game from the SD, it has two identical entries in the list~
- No RGB splash handling (tool doesn't generate them (yet?))

### DIFF
--- a/32blit-sdl/CMakeLists.txt
+++ b/32blit-sdl/CMakeLists.txt
@@ -142,3 +142,7 @@ endfunction()
 function(blit_executable_int_flash NAME SOURCES)
 	blit_executable(${NAME} ${SOURCES} ${ARGN})
 endfunction()
+
+function(blit_metadata TARGET FILE)
+	# do nothing
+endfunction()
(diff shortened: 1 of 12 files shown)


Daft-Freak:
Now that the metadata tool is released, what's left to do here? (I'm assuming at least some UI stuff.)

Gadgetoid:
Haven't built and run this for a while, will try. Broadly I think we need:

* Some UI - tabs along the top with the current directory centered (this will be fun if they're not fixed width) or left aligned (easier but you wont see what the prev directory is)
* Author displayed under description, maybe version too?
* ... I'm not sure!

image

Gadgetoid:
'tis a glorious day that this most immense of PRs gets merged and makes our menu

S U P E R  A W E S O M E

PR #371, opened 2020-09-28
Fix sprite unpacking to RGBA

Daft-Freak:
The else branch in `load_from_packed` had the same issues that were fixed in #282.


Gadgetoid:
For some reason this is stuck in CI purgatory. 😬

Daft-Freak:
Somehow managed to push an empty branch, but hopefully this one actually completes.

PR #374, opened 2020-09-29
Drop the second HAL build

Daft-Freak:
And avoid building it entirely in external projects. Unneeded since we're not linking it into games any more.


Gadgetoid:
The almighty CI has spoken and says this hasn't burst into unholy fire. Good catch- thank you!

PR #365, opened 2020-09-24
Assets pipeline for palette-cycle/swap and shmup

Daft-Freak:
The tilemap-test PR reminded me, opening this so I don't forget again. VS fails to build as it can't find the palette. (https://github.com/pimoroni/32blit-tools/issues/15)

Fixes #88?

PR #375, opened 2020-09-29
Asset pipleline for tunnel

Daft-Freak:
With #365, this should be all the examples converted. (Was going to include shmup here, but it does palette stuff).


Gadgetoid:
Whew! Thank you.

PR #379, opened 2020-09-30
WIP: Add a version field to the API section

Daft-Freak:
Break everything. Needs some checking somewhere.

### DIFF
--- a/32blit-stm32/Inc/executable.hpp
+++ b/32blit-stm32/Inc/executable.hpp
@@ -5,7 +5,7 @@ constexpr uint32_t blit_game_magic = 0x54494C42; // "BLIT"
 
 using BlitRenderFunction = void(*)(uint32_t);
 using BlitTickFunction = bool(*)(uint32_t);
-using BlitInitFunction = void(*)();
+using BlitInitFunction = bool(*)();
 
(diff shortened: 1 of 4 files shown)


Gadgetoid:
Yeah I can't quite envision *how* this should be checked.

I guess it would have to be load-time- since it would have to trap:
1. Games being loaded from ext flash that are incompatible with a newly updated firmware
2. Games being loaded from SD that are incompatible

I think if we append-only the API we can gently skirt around issues and avoid needing to to this.

Daft-Freak:
I was thinking maybe an init check somewhere for the version being too low. Still only handles append-only, but at least a game can abort early instead of randomly crashing later. Could even be a major/minor thing and abort on major mismatch...

(Sorry, intended to expand the description.)

Daft-Freak:
So, that approach was okay then? (Forgot again to comment after updating...)

Gadgetoid:
It seems sensible enough- we can expand it by barfing up a human-readable error but that's not super important.

I did wonder about something like `errno` so launching a game can fail in one of many ways which can then be communicated...  but then I couldn't think of any other error cases that shouldn't just be handled by the game itself (missing assets for example)

PR #380, opened 2020-10-01
Reset to firmware on abort/exit

Daft-Freak:
Found this while rebasing all my branches yesterday, wasn't included with the initial error handling PR due to depending on user-only.

Intended to catch abort and do something more useful than hang in an infinite loop, `exit` working is an err... undocumented bonus feature.


Gadgetoid:
😆 

Thank you!

PR #381, opened 2020-10-01
Tween demo

Daft-Freak:
An attempt at the "cool demo" part of #350, left/right to cycle through functions. There's a screenshot of this in #358.


Gadgetoid:
Love it 😆 thank you!

Might have to play with some tween functions now.

PR #382, opened 2020-10-01
Reduce calls to init_pin

Daft-Freak:
Take advantage of the fact that `HAL_GPIO_Init` can setup multiple pins at once to save a few hundred bytes (`gpio::init` ends up as one of the largest non-HAL functions). Had this one lying around for a while as I can't decide if the LTDC bit is too messy.


Gadgetoid:
Good grief, this saves a byte or two 😆 

Nice work- the LTDC stuff is an abomination no matter what you do with it- and the .ioc documents the pin config anyway. Would rather have the saved bytes.

PR #384, opened 2020-10-02
Check packed_image header

Daft-Freak:
Validate the header before attempting to load. Only checks for `SPRITEPK` as `SPRITERW` isn't handled anywhere.


Gadgetoid:
Thank you- will have to look into SPRITERW. We might have decided at some point that it was only worth having packed sprites... but raw must have its uses.

Daft-Freak:
The one thing I can think of for raw sprites is that you could use them to save RAM by using the data directly from flash.

PR #385, opened 2020-10-02
Rename storage product from "Product"

Daft-Freak:
No more slightly confusing "STM Product" device in my disk list. Other fields may also want updating.


Gadgetoid:
Haha! Handy. Thank you.

PR #387, opened 2020-10-02
Trigger re-link on icon/splash updates

Daft-Freak:
Small issue with the tool: if you have neither defined, it won't generate the .cmake file and the build fails due to a missing include.


Gadgetoid:
Is a blank cmake file acceptable?

Daft-Freak:
Looks like it is. `set(METADATA_DEPENDS)` also would work.

PR #388, opened 2020-10-03
WIP: PIC for games

Daft-Freak:
The last part is finally here...

1. Custom PIC/non-PIC newlib/libstdc++ builds (currently based on the versions from the 7.3 toolchain, build scripts in `stdlib/`)
2. Toolchain automatically downloads and extracts these
3. Games are built with PIC flags linked against the PIC libraries, firmware uses the non-PIC libs.
4. Games have a prepended "RELO" header, containing offsets to patch (GOT, init/fini arrays, data relocations) (currently `tools/gather-relocs`)
5. At flash time (from SD or over CDC) the relocation header is parsed and skipped, the flash offset is applied as we go
6. Many games at once!

A few relatively minor issues remain:

- CDC flashing doesn't check if the game is already there. (SD flashing uses the checksum + name)
- ~`gather-relocs` and the metadata tool don't work together (`[target].bin` contains metadata, `[target].reloc.bin` has reloc data, metadata tool runs after `gather-relocs` and doesn't handle the extra header)~
- ~The stdlibs are currently being downloaded from my "test server".~ Also, since CMake's zip extration command is only supported in the latest version the extraction is done by running `zip`.~
- The calculation for where a game should be flashed is "just throw it after the previous one" with no handling for running out of space or deleting things.
- Probably more, it's been a while...

### DIFF
--- a/firmware/firmware.cpp
+++ b/firmware/firmware.cpp
@@ -592,21 +592,6 @@ uint32_t flash_from_sd_to_qspi_flash(const char *filename)
     return false;
   }
 
-  // check header
-  BlitGameHeader header;
-  if(f_read(&file, (void *)&header, sizeof(header), &bytes_read) != FR_OK) {
-    f_close(&file);
-    return false;
-  }
-  f_lseek(&file, 0);
-
-  uint32_t flash_offset = get_flash_offset_for_file(header);
-
-  // erase the sectors needed to write the image
-  erase_qspi_flash(flash_offset / qspi_flash_sector_size, bytes_total);
-
-  progress.show("Copying from SD card to flash...", bytes_total);
-
   // check for prepended relocation info
   char buf[4];
   f_read(&file, buf, 4, &bytes_read);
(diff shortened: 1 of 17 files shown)


Gadgetoid:
TODO for me:
- [ ] Test all the things!
- [ ] Fix metadata tool to handle reloc header
- [ ] Grab the stdlibs and host them somewhere like get.pimoroni.com or just shove 'em onto GitHub for distribution
- [ ] ???

Handling out-of-space errors and fragmentation is going to be tricky. I guess we could:
1. Scan for empty blocks big enough for the game and fit it in where it fits
2. If no suitable empty block region is found, throw a "you need to free up N blocks" error and..
3. maybe drop people to a tool that visualises the Flash layout (those 8x8 icons could come in useful here) and gives them a rectangle corresponding to the game they were trying to load
4. they have to fit this rectangle somewhere, it will delete any games it overlaps and replace them
5. AAAAh this is going to get complicated fast!

Remember the Dreamcast savegame manager?

image


If blocks are 64k... this might be a dicey screen 😆

Gadgetoid:
Okay actually working out a block manager concept shows me how much I'm overthinking this. Could look neat... 

image


but it's massively overkill when the menu just needs a button bind for deleting games (from SD or FLASH):

image

With maybe some confirmation:

image

And then we can worry about fancy block view external Flash managers later 😆

firmware/firmware.cpp
```diff
@@ -558,13 +577,12 @@ void update(uint32_t time) {
 // returns address to flash file to
 uint32_t get_flash_offset_for_file(BlitGameHeader &bin_header) {
 
-  // temporary load address for working on multiple app support without PIC being ready
-  // in future this will probably be more of a "find free space" function
   if(bin_header.magic == blit_game_magic) {
-    auto expected_addr = bin_header.start;
+    // TODO: handle full
```
Gadgetoid, line 581: Handle full what? 😆

Daft-Freak: I think that was about the flash being full. Currently it just fails if you manage to do that...

Gadgetoid:
YOLO

PR #389, opened 2020-10-04
Draw sprites a line at a time in sprite_blit

Daft-Freak:
Should be faster due to less offset calcs and calls to `bbf`. Probably more optimisations to do around here... I just kept noticing this one.


Gadgetoid:
Gets shmup (a.k.a. ridonkulous mess of alpha blended sprites) from 15-16ms to 12-13ms per frame. Good catch, thanks!

PR #391, opened 2020-10-05
Smaller sort function

Daft-Freak:
Switch out the sort function for the file list and save ~800 bytes.


Gadgetoid:
There wont be any bytes left at this rate 😆 

Thanks!

PR #392, opened 2020-10-05
Menu cleanup

Daft-Freak:
A bunch of changes to reduce the size of `blit_menu_render`, which was larger than the firmware's `render`... (-~460 bytes)

And a little tidying.


Gadgetoid:
Thank you!

PR #393, opened 2020-10-05
Use python tool for .flash targets

Daft-Freak:
Removes the need to do a local build before a device build... Attempted to update the docs, `docs/32Blit-Loader.md` will need updated/replaced...


Gadgetoid:
Finally, no more: 
```
CMake Warning at 32blit.cmake:48 (message):
  32Blit tool not found.  Looking for 32Blit.exe instead
```

Cheers!

PR #394, opened 2020-10-06
Specify Python version for the VS sln build

Daft-Freak:
The CMake one already did this, should fix the build.


Daft-Freak:
Oops, small mistake. Will probably want to squash that...

Gadgetoid:
Haha, thank you! I didn't even notice the build had blown up.

PR #397, opened 2020-10-08
Menu refactor

Daft-Freak:
Reworks the firmware menu into a basic reusable menu class and adds scrolling and customisation options. Also changes the menu appearance a bit:

screenshot13

Customised menu from my emulator:
emumenu

And even the file browser:
emubrowse

Usage example:
```c++
class MyMenu final : public Menu {
public:
  MyMenu() : Menu("Title", item_data, 2) {
    // optionally replace some pens/layout
  }

  // optionally render_item/update_item for more customisation

  void item_activated(const Item &item) override {
   // do things
  }

  Item item_data[2]{
    {0, "Item 1"},
    {1, "Item 2"}
  };
};

MyMenu menu;

...
menu.update(time);

...
menu.render();

```

All the base class does is display a scrollable list of items that you can sellect with A.

Has (possibly a lot of) overlap with #216. I think that one does a more in the common menu class. (Sub menus/navigation, lists, sliders). Hmm... probably need a compromise somewhere between features and blowing up the build size...

### DIFF
--- a/32blit-stm32/Src/32blit.cpp
+++ b/32blit-stm32/Src/32blit.cpp
@@ -460,6 +460,16 @@ public:
     selected_item_background = get_menu_colour(4);
   }
 
+  void draw_slider(Point pos, int width, float value, Pen colour) const {
+    const int bar_margin = 2;
+    const int bar_height = item_h - bar_margin * 2;
+
+    screen.pen = bar_background_color;
+    screen.rectangle(Rect(pos, Size(width, bar_height)));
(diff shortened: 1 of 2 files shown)


Gadgetoid:
I think #216 should ideally be an optional library for users to build enhanced menus with if they chose- I'd hate to just throw away the effort put into it. That said we're definitely getting cozy in the firmware and don't want to blow our budget, so this is a better choice of mainline enhanced menu. Thanks!

PR #396, opened 2020-10-06
Firmware list/erase commands

Daft-Freak:
Small cleanup/fixes and adds two new commands to the firmware:

- `__LS` - lists games in flash (returns offset and size (uint32), then metadata for each game)
- `ERSE[32bit offset]` - erases game at an offset

(`__LS` already existed, but all it did was reload the file list)

WIP for the tools:
```diff
diff --git a/src/ttblit/tool/flasher.py b/src/ttblit/tool/flasher.py
index 053106c..530ac28 100644
--- a/src/ttblit/tool/flasher.py
+++ b/src/ttblit/tool/flasher.py
@@ -1,5 +1,6 @@
 import logging
 import pathlib
+import struct
 import time
 
 import serial.tools.list_ports
@@ -28,6 +29,8 @@ class Flasher(Tool):
         self.op_flash.add_argument('--file', type=pathlib.Path, required=True, help='File to flash')
 
         self.op_delete = operations.add_parser('delete', help='Delete a game/file from your 32Blit')
+        self.op_delete.add_argument('--offset', type=int, help='Flash offset of game to delete')
+
         self.op_list = operations.add_parser('list', help='List games/files on your 32Blit')
         self.op_debug = operations.add_parser('debug', help='Enter serial debug mode')
         self.op_reset = operations.add_parser('reset', help='Reset your 32Blit')
@@ -138,11 +141,28 @@ class Flasher(Tool):
 
     @serial_command
     def run_delete(self, serial, args):
-        pass
+        self._reset_to_firmware(serial)
+        serial.write(b'32BLERSE\x00')
+        serial.write(struct.pack("<I", args.get('offset')))
 
     @serial_command
     def run_list(self, serial, args):
-        pass
+        self._reset_to_firmware(serial)
+
+        serial.write(b'32BL__LS\x00')
+        offset_str = serial.read(4)
+
+        while offset_str != '' and offset_str != b'\xff\xff\xff\xff':
+            offset, = struct.unpack('<I', offset_str)
+            size, meta_size = struct.unpack('<IH', serial.read(6))
+
+            metadata = None
+            if meta_size:
+                metadata = serial.read(meta_size)
+
+            print(offset, size, metadata[20:45].rstrip(b'\0') if metadata else None)
+
+            offset_str = serial.read(4)
 
     @serial_command
     def run_debug(self, serial, args):

```
Needs more metadata parsing...

PR #399, opened 2020-10-13
Fix offset calc in scan_flash

Daft-Freak:
Included in #388 and #396, ignore if either of those get merged first.


Gadgetoid:
Yikes! Thank you. I'll merge this one first and hope it doesn't make a conflictey headache.

PR #400, opened 2020-10-14
Remove darwin toolchain

Daft-Freak:
Pretty sure this is unneeded as building for macOS on a mac isn't cross-compiling and it works fine on mine (CMake defaults to clang anyway)...


Gadgetoid:
We'll find out, I guess! Thanks!

PR #401, opened 2020-10-14
Make audio sample rate constant

Daft-Freak:
Assuming this wasn't intentional since the actual output rate is hardcoded elsewhere. Significantly reduces time spent in `get_audio_frame` (quickly tested in an SDL build).


Gadgetoid:
Definitely not intentional, probably some legacy nonsense we never cleaned up.

PR #402, opened 2020-10-14
Avoid copying entire screen surface in set_screen_mode

Daft-Freak:
This avoids the blend functions from the firmware being used in user code, which are slightly slower due to optimisation level differences. The fact that `set_screen_mode` resets most of `screen` should probably be documented somewhere... (it already did that)

I only noticed this due to drawing not being any faster in a release build...


Gadgetoid:
Ooof! Whoops. And thanks. Actually amazing that we can *accidentally* call the firmware blend functions from user code. Darn blend pointers!

PR #404, opened 2020-10-17
Expand uploading docs

Daft-Freak:
...and remove a few references to the old tool. Might need a big "DO NOT UPLOAD GAMES USING DFU" somewhere.


Gadgetoid:
Thank you!

PR #405, opened 2020-10-21
Add VS Code CMake tools config

Daft-Freak:
This makes the toolchain automatically appear as a kit in VS Code (still needs the manual config in the docs for out of tree projects)


Gadgetoid:
Aha! Awesome, thank you!

PR #406, opened 2020-10-22
Improve slash handling for buffer files in list_files

Daft-Freak:
Makes sure that the leading slash is stripped after doing something like `add_buffer_file("/file", ...)`. Also fixes `list_files("dir/")` not working for buffer files.


Gadgetoid:
Thank you!

PR #407, opened 2020-11-02
SD card insert/remove detection

Daft-Freak:
Handles inserting/removing the SD card without needing to reset. Also displays a different message in the menu for no card detected.


Gadgetoid:
I need this in my life! Thank you.

PR #409, opened 2020-11-06
Check if the tools are installed

Daft-Freak:
Fail CMake configure if the tools are not installed instead of blowing up later.


Gadgetoid:
Thank you!

Daft-Freak:
That was fast! :smile:

PR #410, opened 2020-11-08
Packed image/surface improvements

Daft-Freak:
Reverts `byte_count` back to how it was originally (before I broke it), handles a 0-size palette as a full 256 colours and changes the packed image format a bit to allow larger images.

Also, handles loading unpacked images, adjusts `Surface::load` to be more consistent with `SpriteSheet::load` and adds a helper to use an unpacked image without loading it into ram...

Requires https://github.com/pimoroni/32blit-tools/pull/32 for breaking changes to `packed_image` format.

### DIFF
--- a/32blit/graphics/surface.cpp
+++ b/32blit/graphics/surface.cpp
@@ -52,6 +52,20 @@ namespace blit {
     return load((const packed_image *)data);
   }
 
+  Surface *Surface::load_read_only(const packed_image *image) {
+    if(memcmp(image->type, "SPRITERW", 8) != 0)
+      return nullptr;
+
+    if(image->format > (uint8_t)PixelFormat::M)
+      return nullptr;
+
+    return new Surface(nullptr, (PixelFormat)image->format, image);
+  }
+
+  Surface *Surface::load_read_only(const uint8_t *data) {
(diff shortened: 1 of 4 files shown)


Gadgetoid:
I've merged pimoroni/32blit-tools#32 plus my additional changes to migrate metadata and other byte-style fiddling to Construct. This is now released as v0.1.0 of the toolset - https://github.com/pimoroni/32blit-tools/releases/tag/v0.1.0

I tested a complete build + firmware with these changes and against v0.1.0, and they seem fine.

32blit/graphics/surface.cpp
```diff
@@ -485,45 +537,65 @@ namespace blit {
    */
   void Surface::load_from_packed(const packed_image *image) {        
     uint8_t *palette_entries = ((uint8_t *)image) + sizeof(packed_image);
-    uint8_t *bytes = ((uint8_t *)image) + sizeof(packed_image) + (image->palette_entry_count * 4);
+
+    int palette_entry_count = image->palette_entry_count;
+    if(palette_entry_count == 0)
```
Daft-Freak, line 542: Hmm, this should probably be `&& format == P` for non-paletted images to work... (Not that the tool can generate those)

PR #411, opened 2020-11-09
Spritesheet/Surface load from file

Gadgetoid:
A crude attempt at loading Spritesheet or Surface data from a file.

There seems to be a fair amount of duplication here, and if we switch to bytewise loading of images then `load_from_packed` could end up looking something like this:

```c++
  void Surface::load_from_packed(const packed_image *image, File *file) {        
    uint8_t *palette_entries = ((uint8_t *)image) + sizeof(packed_image);

    int palette_entry_count = image->palette_entry_count;
    if(palette_entry_count == 0)
      palette_entry_count = 256;

    bool is_raw = image->type[6] == 'R' && image->type[7] == 'W'; // "SPRITERW"

    uint8_t *bytes = ((uint8_t *)image) + sizeof(packed_image) + (palette_entry_count * 4);

    bounds = Size(image->width, image->height);

    uint8_t bit_depth = log2i(std::max(1, palette_entry_count - 1)) + 1;

    uint8_t col = 0;
    uint8_t bit = 0;

    if (format == PixelFormat::P) {
      if(file){
        // load palette
        palette = new Pen[palette_entry_count];
        file->read(sizeof(packed_image), (palette_entry_count * 4), (char *)&palette);
      } else {
        // load palette
        palette = new Pen[palette_entry_count];
        for (int pidx = 0; pidx < palette_entry_count; pidx++) {
          palette[pidx] = Pen(
            palette_entries[pidx * 4 + 0],
            palette_entries[pidx * 4 + 1],
            palette_entries[pidx * 4 + 2],
            palette_entries[pidx * 4 + 3]);
        }
      }
    }

    if (is_raw) {
      if(data) { // just copy the data
        if(file){
          file->read(sizeof(packed_image) + (palette_entry_count * 4), image->width * image->height * pixel_format_stride[image->format], (char *)data);
        }
        else
        {
          memcpy(data, bytes, image->width * image->height * pixel_format_stride[image->format]);
        }
      } else { // no data pointer, assume load_read_only is being used
        data = bytes;
      }
      return;
    }
```

### DIFF
--- a/32blit/engine/file.cpp
+++ b/32blit/engine/file.cpp
@@ -102,6 +102,14 @@ namespace blit {
     return api.remove_file(path);
   }
 
+  bool File::open(const uint8_t *buf, uint32_t buf_len) {
+    close();
+
+    this->buf = buf;
+    this->buf_len = buf_len;
+    return true;
+  }
+
   /**
    * Open a file. If a file is already open it will be automatically closed.
    *
--- a/32blit/engine/file.hpp
+++ b/32blit/engine/file.hpp
@@ -66,6 +66,7 @@ namespace blit {
     }
 
     bool open(const std::string &file, int mode = OpenMode::read);
+    bool File::open(const uint8_t *buf, uint32_t buf_len);
     int32_t read(uint32_t offset, uint32_t length, char *buffer);
     int32_t write(uint32_t offset, uint32_t length, const char *buffer);
     void close();
--- a/32blit/engine/file.hpp
+++ b/32blit/engine/file.hpp
@@ -66,7 +66,7 @@ namespace blit {
     }
 
     bool open(const std::string &file, int mode = OpenMode::read);
-    bool File::open(const uint8_t *buf, uint32_t buf_len);
+    bool open(const uint8_t *buf, uint32_t buf_len);
     int32_t read(uint32_t offset, uint32_t length, char *buffer);
     int32_t write(uint32_t offset, uint32_t length, const char *buffer);
     void close();
--- a/32blit/graphics/sprite.cpp
+++ b/32blit/graphics/sprite.cpp
@@ -6,6 +6,8 @@
 #include "surface.hpp"
 #include "sprite.hpp"
 
+#include "../engine/file.hpp"
+
 using namespace blit;
 
 namespace blit {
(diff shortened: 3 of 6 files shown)


Daft-Freak:
Hmm, if we add this to file...
```c++
  bool File::open(const uint8_t *buf, uint32_t buf_len) {
    close();

    this->buf = buf;
    this->buf_len = buf_len;
    return true;
  }
```

Then it could be something like:
```c++
 void Surface::load_from_packed(File &file) {
    packed_image image;
    file.read(0, sizeof(packed_image), (char *)&header);

    int palette_entry_count = image.palette_entry_count;
    if(palette_entry_count == 0)
      palette_entry_count = 256;

    bool is_raw = image.type[6] == 'R' && image.type[7] == 'W'; // "SPRITERW"

    uint8_t *bytes = ((uint8_t *)image) + sizeof(packed_image) + (palette_entry_count * 4);

    bounds = Size(image.width, image.height);

    uint8_t bit_depth = log2i(std::max(1, palette_entry_count - 1)) + 1;

    uint8_t col = 0;
    uint8_t bit = 0;

    uint32_t offset = sizeof(packed_image);

    if (format == PixelFormat::P || !is_raw) {
      // load palette
      palette = new Pen[256];
      file.read(offset, palette_entry_count * 4, (char *)&palette);

      offset += palette_entry_count * 4;
    }

    if (is_raw) {
      if(data) // just read/copy the data
        file.read(offset, image.width * image.height * pixel_format_stride[image.format], data);
      else // no data pointer, assume load_read_only is being used
        data = (uint8_t *)file.get_ptr() + offset;

      return;
    }

    // avoid allocating if in flash
    const uint8_t *bytes, *end;

    if(file.get_ptr())
      bytes = file.get_ptr() + offset;
    else {
      bytes = new uint8_t[image.width * image.height * pixel_format_stride[image.format]];
      file.read(offset, image.width * image.height * pixel_format_stride[image.format], bytes);
    }

    end = bytes + image.byte_count - offset;

    if (format == PixelFormat::P) {
      // load paletted
      uint8_t *pdest = (uint8_t *)data;

      for (; bytes < end; ++bytes) {
        uint8_t b = *bytes;
        for (auto j = 0; j < 8; j++) {
          col <<= 1;
          col |= ((0b10000000 >> j) & b) ? 1 : 0;

          bit++;
          if (bit == bit_depth) {
            *pdest++ = col;
            bit = 0; col = 0;
          }
        }
      }
    } else {
      // packed RGBA
      Pen *pdest = (Pen *)data;

      for (; bytes < end; ++bytes) {
        uint8_t b = *bytes;
        for (auto j = 0; j < 8; j++) {
          col <<= 1;
          col |= ((0b10000000 >> j) & b) ? 1 : 0;

          bit++;
          if (bit == bit_depth) {
            *pdest++ = palette[col];

            bit = 0; col = 0;
          }
        }
      }

      // unpacked, no longer needed
      delete[] palette;
      palette = nullptr;
    }

    if(!file.get_ptr())
      delete[] bytes;
  }
```

With the load helpers using the appropriate `File::open`

(Not tested, roughly what I was thinking when Discord died earlier)

Gadgetoid:
Aaaah! I see now. I was trying to wrap my head around how a common stream-like object could work, nice!

Daft-Freak:
Yeah, fortunately the "fake file" stuff can be reused for that. (I seem to be using that in almost everything...)

Hmm, it would be pretty easy to handle loading JPEGs into a surface as well
```c++
JPEGImage jpeg = decode_jpeg_file(filename);
return new Surface(jpeg.data, PixelFormat::RGB, jpeg.size);
```

I think I left that out because blitting from an RGB surface doesn't (currently) work...

Daft-Freak:
Maybe rename `bytes` to something like `image_data` and change the loops to `for (auto bytes = image_data; bytes < end; ++bytes)`? Then you can just `delete[] image_data`.

Gadgetoid:
Thanks for all your pointers in the right direction here, @Daft-Freak!

PR #412, opened 2020-11-09
Fix repeated typo in file docs

Daft-Freak:
"the the" -> "to the"

Looks like I copy/pasted that a few times... Oops.


Gadgetoid:
We've all been there, thank you!

PR #414, opened 2020-11-11
Move CI to GitHub Actions

Daft-Freak:
Mostly because the Travis "free" tier seems to be a trial now... and also it seems to be getting slower

Currently missing:
 - caching (though still seems faster than Travis)
 - MinGW
 - VS (sln)
 - Travis is still there


Gadgetoid:
Oh wow, we've totally blown the credit limit on Travis - 
image

I'm not totally surprised, but that's a total pain in my ass because CI across the whole org has gone kaput.

Honestly I'm not sure GitHub actions will be much better but at least - as far as I can tell - we'll just get a shock bill at the end of the month instead of CI just totally breaking. And- glancing above- the GitHub actions reporting is *much* nicer than Travis.

Interestingly both GitHub and Travis charge on the order of about 10x for MacOS builds. I wonder if those are worth dropping once we're relatively stable.

Thanks for taking the initiative on this- I had thought something was up with Travis earlier and assumed it was just broken, but... nope 😬

Gadgetoid:
This note might be relevant to us, too - https://github.com/dask/community/issues/107#issuecomment-720555156

> A note of light caution -- currently if you have a matrix of say Python versions and OS in the same workflow file, if one of those matrix entries failed you can't restart it individually, you have to trigger a full re-run of everything defined in that file. On xonsh we've subsetted them into one os per workflow file so we can at least individually restart Windows runs or MacOS runs

Daft-Freak:
Had a bit of this done already to experiment with the warnings stuff. (Need to do something about VS, currently spamming conversion warnings)

> Honestly I'm not sure GitHub actions will be much better but at least - as far as I can tell - we'll just get a shock bill at the end of the month instead of CI just totally breaking.

At least GitHub credits are per-month instead of a one-time trial allocation.

> Interestingly both GitHub and Travis charge on the order of about 10x for MacOS builds. I wonder if those are worth dropping once we're relatively stable.

The Emscripten build actually does a good job of detecting issues that would hit macOS, as it also uses clang/libc++.

I had everything in one job to avoid duplicating steps (as there doesn't seem to be a nice way to reuse common steps), can probably split some stuff out... macOS would be easy as it currently doesn't do releases and the .sln build has very little in common with the other builds.

Gadgetoid:
I think the only time we might want to re-run a *single* step is when something in that step has a temporary (network, probably) failure. That hasn't been super common, to be fair!

Daft-Freak:
Nearly there... Emscripten might need some caching (currently building libraries every time), might split that out as it has a few extra steps.

VS builds are significantly faster now that we don't have to install Python (or update VS).

Daft-Freak:
I think this should do everything the Travis build did and I've tested that releases still work (It does generate a few extra files now as there are always both .zip and .tar.gz).

Three are a few other changes like both VS builds now use VS 2019 and all the Ubuntu-based builds are on 20.04.

Hmm, ended up with quite a few commits, maybe squash this?

Gadgetoid:
A few more commits wont hurt us, I guess! Don't worry about it.

GitHub is being weird and the names of status checks in the branch protection settings don't match the checks here- I guess that'll fix itself!

PR #415, opened 2020-11-13
VS + CMake out-of-tree build docs

Daft-Freak:
Since I keep saying this is easy, I thought I should check that it actually works... which it nearly did. Just a little tweak to SDL detection needed... I think this is easier than having to edit the current .sln to point to the right place...

Definitely no VS files here:
Screenshot
With that many warnings, it has to be a VS build :smile:

Bonus: Also works in VS 2017.

### DIFF
--- a/32blit-sdl/CMakeLists.txt
+++ b/32blit-sdl/CMakeLists.txt
@@ -25,6 +25,11 @@ else()
 			PATHS ~/Library/Frameworks /Library/Frameworks
 		)
 	endif()
+	
+	# fallback guess for SDL location on Windows
+	if(WIN32 AND NOT SDL2_DIR AND 32BLIT_PATH)
+		set(SDL2_DIR "${32BLIT_PATH}/vs/sdl")
+	endif()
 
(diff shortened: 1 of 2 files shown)


ahnlak:
Works brilliantly; presumably also makes the notes [here](https://github.com/pimoroni/32blit-beta/blob/master/docs/Tools.md#visual-studio) about running a separate pre-build step redundant too?

Daft-Freak:
Yeah, nothing extra needed since it takes everything from CMake. (Also, there's nowhere to put that anyway)

Gadgetoid:
R.I.P. sln

Wonder if we should just... quietly sweep it under the rug and hope nobody notices 😆

Daft-Freak:
...and then there's [this](https://github.com/Daft-Freak/32blit-beta/commit/a6aa6d270f5344d659d99c04a77cfff819119aa2). Device builds in VS with no WSL :smile:

Slightly annoying to setup though as you have to edit the JSON a bit (copy/paste, delete `inheritEnvironments`, add `cmakeToolchain`)

PR #416, opened 2020-11-14
VS warning fixes

Daft-Freak:
This fixes most of the non-"converting X to Y" warnings in VS. (With [this slightly nasty patch](https://github.com/Daft-Freak/32blit-beta/commit/98b07a02e62509190030230f3f10f65a3e358b28), there are two in the tool).


Gadgetoid:
Thank you!

I still see a lot of C4244 in the Github Actions build, should that flag be squashing them in CI or just a local VS build?

Daft-Freak:
Didn't include that patch here. Thinking about it again. it's not really that hacky... Would need this one as well to not get warnings when building the library as part of an out-of-tree project: https://github.com/pimoroni/32blit-beta/commit/15a3a4ad41c4330517575e4761004158fbd6fff9

PR #417, opened 2020-11-15
Make SDL window title settable

ahnlak:
Added 32BLIT_SDL_TITLE value, to allow projects to override the default "CMake!" window title.

### DIFF
--- a/32blit-sdl/CMakeLists.txt
+++ b/32blit-sdl/CMakeLists.txt
@@ -85,9 +85,13 @@ if(SDL2_IMAGE_DLL)
 	install(FILES ${SDL2_IMAGE_DLL} DESTINATION bin)
 endif()
 
+if(NOT 32BLIT_SDL_TITLE)
+	set(32BLIT_SDL_TITLE CMake!)
+endif()
+
 target_compile_definitions(BlitHalSDL
 	PRIVATE
-		-DWINDOW_TITLE=\"CMake!\"
+		-DWINDOW_TITLE=\"${32BLIT_SDL_TITLE}\"
 )
 
 if(DEFINED VIDEO_CAPTURE AND VIDEO_CAPTURE)


Daft-Freak:
There's a fallback in the code (https://github.com/pimoroni/32blit-beta/blob/master/32blit-sdl/Main.cpp#L22), so you can probably simplify this.

ahnlak:
There is, but the current CMakelist overrides it with `-DWINDOW_TITLE=\"CMake!\"` - I suppose you could make the case for removing that entirely from CMake, and just using the `#define WINDOW_TITLE` directly?

I partially assumed that the long term intent for introducing the override in CMake was so that one day it could do something clever with the metadata, but figured it would always be nice to have the final say...

Daft-Freak:
Hmm, yeah it has always set "CMake!" as the title... Guess the (more reasonable looking) fallback only gets used by VS right now. I'd say that removing the hardcoded "CMake!" title was an improvement anyway... 

Getting the title from the metadata would be an... interesting problem. Metadata gets parsed as part of the executable build, which is too late to affect the library... and then there's the fact that you can build multiple games with the same library (like the examples).

(I was thinking about this recently, would need something like `extern const char *title;` in the library, then generating and linking in a file with the definitions later...)

Anyway, not really relevant to this significantly more simple change.

ahnlak:
I did briefly wonder if I should be making it settable through the API, but then I started worrying that I was overthinking it because the primary target (?) is the Blit rather than SDL builds, and obviously window titles have no meaning on the Blit.

Plus, it would involve touching quite a few files and I didn't want my first code PR to be an epic Daft-Freaked sized one ;-)

Gadgetoid:
I'm happy to look into parsing metadata into a C-include/object file within the 32blit tools- if it means the same title can be used in 32blit vs desktop games.

It might also be possible to have the blit tools dump these values into the ENV early in the build process, or potentially directly into CMake via just invoking a special command? I mean... we already call the tool to generate CMake files for the asset pipeline? 🤔 

This is a good start, though!

Daft-Freak:
> It might also be possible to have the blit tools dump these values into the ENV early in the build process, or potentially directly into CMake via just invoking a special command? I mean... we already call the tool to generate CMake files for the asset pipeline? thinking

The problem with that is that the metadata processing is done as part of the executable build, and the title would have to be specified as part of the library build... You could probably append the compile definitions later but it's a little nasty to go messing with another target and would sort-of explode if you were building multiple executables.

ahnlak:
Well... you could have the library expose functions to override the window title (and other metadata stuff), and have the build tool generate code from the metadata file to fill it in (or is that just getting horribly messy?)

PR #418, opened 2020-11-16
Improve macOS output

Daft-Freak:
Generates a `.app` bundle by default, copies the SDL frameworks into it and used the metadata to partially setup the bundle info.

(Requires https://github.com/pimoroni/32blit-tools/pull/36)

This is what the examples look like after this:
Screenshot


Gadgetoid:
Looking into the `icns` format required for macOS bundles suggests we could convert our 128x96 thumbnail images into icons and they might even (given how high density displays are these days) look passable. By way of example, here's the Another World icon scaled to 128x128 using the extra pixels at the bottom to squeeze in a 32blit logo:

another-world-icns

And if we really wanted to, while we're faffing about processing images we could bake text in there too:

image

Food for thought, anyway! The icns file format is dead simple.


(And thank you! I know shipping onto other platforms is a secondary concern, but it's really cool that people can share their games even outside of the 32blit ecosystem)

Daft-Freak:
Seeing as my mac is currently behaving, I might check what that looks like later...

PR #419, opened 2020-11-16
Defaults fix

ahnlak:
Some functions in 32blit have defaults declared in their `.cpp` instead of `.hpp`, which pretty much renders them invisible.

First noticed it playing with Tweens (which are, by the way, awesome), then found other instances, in Timers and Maps.


Gadgetoid:
Good catch! I was grappling with something like this today on a different project and had to look up where to put the blasted things!

Thank you!

PR #421, opened 2020-11-17
Font fix

ahnlak:
Centres narrow letters when rendering as a fixed font; otherwise these are left-aligned and the kerning is horrible.

### DIFF
--- a/32blit/graphics/text.cpp
+++ b/32blit/graphics/text.cpp
@@ -98,9 +98,7 @@ namespace blit {
       }
 
       if (!variable)
-      {
         char_width = font.char_w - fixwidth;
-      }
       else
         char_width = font.char_w_variable[chr_idx];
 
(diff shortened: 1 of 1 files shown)


Daft-Freak:
I'd move the `- fixwidth` down to the `c.x += char_width;`, then you could move setting `char_width` above the loops and fix the x loop issue you also pointed out.


Oh, and your `{` should be on the line above...

ahnlak:
Hmm, good point on moving the `char_width` setting; I did fiddle with the x loop issue but wasn't entirely convinced the savings were measurable (you have to knock one off the variable width, so that's an extra op per letter to save time on *some* of them...)

Now, if I update this branch do I have to resend the PR or does it all just work itself out?

Daft-Freak:
Ah, right `char_w_variable` can be greater than `char_w` because padding... Maybe just leave that alone then...

Updating the branch should just work.

Gadgetoid:
Thank you!

Definitely an improvement, although it's still a little janky. Occupational hazard if you need fixed-width, I guess!

Before:

image

After:

image

PR #422, opened 2020-11-17
Unbreak non-framework macOS builds

Daft-Freak:
Should build, won't package the libraries though. (That would need something like https://github.com/Daft-Freak/32blit-beta/commit/647d0fe1cd4eac77bfe3709305d186ae089b5fb8, which ended up [failing with permission errors when I tried it](https://github.com/Daft-Freak/32blit-beta/runs/1412024010?check_suite_focus=true) )

### DIFF
--- a/32blit-sdl/CMakeLists.txt
+++ b/32blit-sdl/CMakeLists.txt
@@ -148,7 +148,7 @@ function(blit_executable NAME SOURCES)
 		)
 	endif()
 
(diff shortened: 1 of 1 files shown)


ahnlak:
Can confirm it fixes my builds - yay!

Gadgetoid:
haha Daft to the rescue! Thank you!

PR #423, opened 2020-11-17
Disable VS "conversion from X to Y, possible loss of data" warning

Daft-Freak:
So much noise, just give up and disable that warning. You can decide if this is a good idea :smile:


Gadgetoid:
I think we'll live 😆 

It can always be re-enabled in future if it turns out to catch people out.

Thank you!

PR #424, opened 2020-11-17
Documentation for asset options

Daft-Freak:
I went here to add something for fonts as I don't think that's documented anywhere. May have got a bit carried away...

Almost guaranteed to have some typos somewhere.


ahnlak:
Typos you say? Well, perhaps a little.... (pallete != palette)

Daft-Freak:
... and I only got the important one wrong :facepalm:

Gadgetoid:
This is much needed, thank you!

PR #425, opened 2020-11-18
Delay parts of screen mode switching

Daft-Freak:
This should avoid displaying distorted frames with the wrong mode.

This demo switches between modes constantly, and can be used to see the improvement: [screen-mode.zip](https://github.com/pimoroni/32blit-beta/files/5560113/screen-mode.zip)

(Also, it looks like the rightmost column of the screen is missing...)

### DIFF
--- a/32blit-stm32/Src/display.cpp
+++ b/32blit-stm32/Src/display.cpp
@@ -37,9 +37,13 @@ namespace display {
   Pen palette[256];
 
   ScreenMode mode = ScreenMode::lores;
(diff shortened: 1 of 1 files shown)


Gadgetoid:
Thank you!

Rightmost column of the screen, you say? 🤔

Daft-Freak:
You can see it in that test bin, in lores the right edge is a single pixel, in hires it's missing.


...or just `screen.line(Point(319,   0), Point(319, 239));`

PR #426, opened 2020-11-18
Remove some old and broken build stuff

Daft-Freak:
Just deleting some files that must be unused, because they're broken.


Gadgetoid:
Hulk. Smash!

Thank you!

PR #427, opened 2020-11-20
Hide folders beginning with '.' in the firmware menu

Daft-Freak:
No more `.Trash-` folders.


Gadgetoid:
oooh! Thank you. I meant to do something along these lines so I had somewhere to put old versions of games that I somehow compulsively cannot bring myself to delete.

PR #431, opened 2020-12-02
Make SD reads even faster

Daft-Freak:
Managed to get even more speed out of this...

Tested on _three_ cards this time (all MB/s):
- 2.32 -> 2.42
- 5.68 -> 6.47
- 5.84 -> 6.67

Fastest is ~53Mbits, which is getting pretty close to the 60MHz clock.


Gadgetoid:
Works for me, and is good times for Lua! Thank you. Can't believe you're still squeezing perf out of this, haha!

PR #432, opened 2020-12-02
Fix HOME+reset and crash reset

Daft-Freak:
Broken in 71195ae as that check happens after the firmware has done the switch.

(Though somehow still seemed to work sometimes...)


Gadgetoid:
Was juuuuust building a test of this after trying out SD card speedup 👍 

/me waiiiits patiently for CI

Thank you!

PR #433, opened 2020-12-02
Make sdl2-config.cmake auto-fetch libraries

Gadgetoid:
Rather than assuming the user has read the docs, this update to `sdl2-config.cmake` will auto-fetch the required SDL2 libraries for a Visual Studio CMake build when first configured.

If the user already has an `include` directory, it will do nothing.

May not be totally robust and catch all edge cases, and also does things outside of the build dir (which is potentially really bad) so:

TODO: suggestions for making this a better CMake citizen.

### DIFF
--- a/.github/workflows/cmake.yml
+++ b/.github/workflows/cmake.yml
@@ -82,7 +82,6 @@ jobs:
       if: runner.os == 'Windows'
       shell: bash
       run: |
-        ci/install_sdl_vs.sh
         python -m pip install 32blit
 
     - name: MinGW deps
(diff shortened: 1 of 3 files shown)


Daft-Freak:
Hmm, since fetching the libs would be the same for any build and you really only want it to happen once, installing to the "source" dir is maybe not that bad here...

Installing SDL2_image` in the config file for SDL2 is a little weird though. Then again, installing _anything_ in a config file is weird since it's supposed to be part of the install... but the windows build of SDL doesn't have one.

So uuh... :shrug: Probably want to remove [this](https://github.com/pimoroni/32blit-beta/blob/master/.github/workflows/cmake.yml#L85) so that CI tests it though.

Gadgetoid:
Whoop, this passes CI now I've fixed the file hash- I think my weird borrowed WiFi setup was causing corrupt downlaods, or I was confusing the expected/actual file hashes.

I'm not particularly happy with how fault tolerant this... isn't. It's more or less a case of "if anything goes wrong, delete these 2 files and 4 directories" and that's a bit jank. I expect it to go right in 99% of cases though so maybe the odd support dance is an acceptable compromise.

I think this loosely falls within the remit of `sdl2-config.cmake` 😆

PR #439, opened 2020-12-08
Install the right bins

Daft-Freak:
Noticed the release contained the non-reloc'd bins.

(Another reminder to change this to a nicer ext...)


Gadgetoid:
D'oh! Yeah we do need to go for `blitgame` or `32blit` or `blit` or `32b` or something. Thank you.

PR #441, opened 2020-12-08
Fix matrix-test crash

Daft-Freak:
By "fixing" the sign conversion warning here, I turned it into an infinite loop. (Which would read out of bounds by a few billion).

Replace the reverse loop with a reverse iterator.


Gadgetoid:
Serves me right for not manually running all of the examples one-by-one before merging 😆 

Thank you!

PR #444, opened 2020-12-08
Fix the compare to find .blit

Gadgetoid:
This could not have had a better PR number.


Daft-Freak:
Note to self: learn to count. I guess one of us should've tested that...

Gadgetoid:
I did test it *after* I merged 🤣

PR #445, opened 2020-12-08
Finish the reloc-tool move

Daft-Freak:
Switch over to the one in 32blit-tools and clean up.


Daft-Freak:
Next to be cleaned up is probably the stdlib build scripts, since the ones [here](https://github.com/Daft-Freak/stdlibs) actually work. (That repo should be moved to the org?)

Gadgetoid:
You should have authority to transfer the stlib repo into the org, if you wish!

Gadgetoid:
Waiting on #446 and then this should be Good To Go! Thank you.

PR #447, opened 2020-12-09
one liner build script.

shane-powell:
Updates 32Blit.
Removes and recreates build folder and builds everything.
Moves all .blits into single blits folder within build directory.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -1,3 +1,4 @@
+pip3 install --upgrade pyelftools
 pip3 install --upgrade 32blit
 rm -r build.stm32/
 mkdir build.stm32
--- a/build.sh
+++ b/build.sh
@@ -1,4 +1,3 @@
-pip3 install --upgrade pyelftools
 pip3 install --upgrade 32blit
 rm -r build.stm32/
 mkdir build.stm32
--- a/build.sh
+++ b/build.sh
@@ -1,0 +1,8 @@
+pip3 install --upgrade 32blit
+rm -r build.stm32/
+mkdir build.stm32
+cd build.stm32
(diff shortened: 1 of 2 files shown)


shane-powell:
Not sure if you already have something like this. Its not much but it saves me some time.

Gadgetoid:
Oh should probably move this to `tools/`

I could see myself using this, though I have an awful habit of putting things in my build folder that shouldn't be there. 😆 

Jon had something like this called go.sh IIRC. Waaay back when!

I'm surprised nobody has tried to dockerise this yet. 😬

shane-powell:
Moved to tools and tested. :)

Daft-Freak:
Maybe do what the CI does and `-DCMAKE_INSTALL_PREFIX=`$PWD`, then `make install` which would get everything in `build/bin` (including the firmware).

Gadgetoid:
I'm... genuinely shocked I never tried using an install prefix 😆 but yeah that's a really sensible idea.

I guess:

```
cmake .. -DCMAKE_TOOLCHAIN_FILE=../32blit.toolchain -DCMAKE_INSTALL_PREFIX=$PWD
make
make install
```

And then everything is in `bin` and no need to mkdir/copy.

Gadgetoid:
I think it's safe to say this is a little redundant now? We probably support too many host operating systems to make preferring build scripts for one make sense. Could, instead, document the above revelation?

Daft-Freak:
Hmm, I guess the easy thing to do now... is just download the examples from the release files. Only really need to build the SDK itself if you're working on the firmware.

Gadgetoid:
Closed as stale. Thank you all!

PR #448, opened 2020-12-09
Fix time going negative in the MP3 example

Daft-Freak:
Switches the sample position to unsigned to avoid `(stream.get_current_sample() * 1000) / 22050` wrapping to negative at the `* 1000`.


Gadgetoid:
Ooh! Thank you. I actually excluded the mp3 example from the shipped units because it took too long to flash. Along with voxel 😢 
Will have to release a top-up-pack for those users to get fixed versions of the extra stuff!

PR #449, opened 2020-12-09
Add joystick support and version info in loader

mikerr:
Adds in joystick control for scrolling lists of games in the loader.

Also displays a game's version in the metadata display section.
(replaces the display of game size / blocks)


Gadgetoid:
Awesome, it's like you read Jon's mind since it was only mentioned today that it's weird the joystick does nothing in the menu 😆 

I think we've got room to keep the block size if it's shuftied down a bit. It can be useful for setting expectations when launching/flashing games. Though better handling of flash contents is on the TODO list for... sometime in the future!

Gadgetoid:
P.S. welcome back! 😁 and thank you!

PR #450, opened 2020-12-09
WIP: Launcher/firmware split

Daft-Freak:
Slightly WIP due to having no user. Just enough to allow a launcher to be built.


Gadgetoid:
Awesome!

What's your perspective on assigning files to be launched- ie: saving a path to a file on SD card somewhere for the firmware to pick up, or just having the launcher invoke the right application in the first place?

I guess we'll have to do a breaking change to metadata to include "application type" and "supported file formats".

Daft-Freak:
The launch API just takes a filename so the firmware could handle it transparently, though I suppose the launcher needs to know what it _can_ launch... 

Hmm... as the firmware would have to scan flash to find the launcher, it could build a list of "launchable" file types and expose a `can_launch` API that checks that. Or maybe we want the launcher to show some of the metadata from the "handler app".

Daft-Freak:
Oh, and can probably just reserve some room for a string in `API` or `Persist`. Probably the latter to err... persist.

Daft-Freak:
Super rough launcher split out from the firmware.

Missing/TODO:
- SD detection (not exposed through the API)
- Launching directly from flash/without an SD card
- Entering mass storage (Move to the menu?)
- Most of the `flash` subcommands from the tool don't work (can't reset to firmware and stay there, most of them could be made to work _without_ resetting)
- A bunch of duplicated code, mostly metadata parsing

Gadgetoid:
Moving Mass Storage to the menu is an all-out good idea- I'm constantly hitting it accidentally 😆 

SD detection through the API would be useful for games that require resources- helping them fail a little more gracefully.

Struck me today that the firmware needs to have enough smarts to get you started with a loader. I think:

1. If there's no loader in flash and
2. There's a loader.bin on SD card
3. Flash and launch it

If there's no loader in flash and no SD card I guess it would just have to dump the user to an info page: "flash a loader or stick loader.bin on your SD card".

Sorry bit of a brain dump!

Daft-Freak:
Have more updates here, will push/comment when less tired... *yawn*

Daft-Freak:
- Now flashes `launcher.blit` automatically if there isn't one in flash, complains if there isn't one on the SD either.
- SD detection is exposed as `blit::is_storage_available`, though it turns out the the timeout handling in the SD code is entirely broken.
- Mass storage toggle added to the menu
- Flasher commands working (with `_reset_to_firmware` removed from the tool)

The handling of the flasher commands and mass storage is not great, it switches back to the firmware and doesn't clear in `render` to work as a sort-of overlay.

Still duplicated between the firmware/launcher:
 - metadata parsing
 - dialog/progress UI
 - `executable.hpp` - contains structs for the game header/ "raw" metadata

Also, starting to collect a few random bugfixes here...

Gadgetoid:
Whoop, some of the fundamentals are in place. I'm afraid I massively underestimated just how much plumbing would be required to get this off the ground- you're a regular Atlas juggling celestial spheres right now.

Daft-Freak:
Mostly functional, lots of small issues:

- menu displays "exit game" in the launcher, pressing it just restarts the launcher (to the firmware there isn't a difference between the launcher/ a game)
- no current item persistence, but that hasn't really worked right since adding folders
- if the launcher crashes, selecting either "yes" or "no" from the oops dialog relaunches the launcher
- displaying progress bars from the firmware is done by not rendering the "game", which kind-of works as long as we're in hires mode... and you don't open the menu (only possible during CDC commands, everything else is blocking)
- crash dialog just renders on top of whatever was on the screen before the crash with... interesting effects
- need to make sure audio callbacks don't happen if we've unmapped the flash

And then there's the duplicated code:
- common metadata parsing/executable format structs (probably not useful to expose for anything that isn't a "launcher")
- dialog helper (maybe this is useful for games?)
- general launcher/firmware code cleanup, avoiding changing too much here since it already has to be rebased whenever the firmware is touched...


Oh, and CI failed due to a random APT failure, yay.

Gadgetoid:
Probably a good idea to get this merged so we have some good exposure to the issues and don't keep conflicting your work with random firmware tweaks, then :laughing: 

Agreed that game don't need metadata parsing. If we need to include extra headers/source files to build a "launcher"- ie: it's a more complicated process than building a game- then that's kinda par for the course.

Dialog box stuff could probably be pretty useful for games, though that might be best left to a user lib since the potential for feature creep is huge and the firmware dialogs are clean and direct and should probably stay that way. Again probably not a bad thing if this is exposed for launchers but not generally.

Escaping a crashed launcher could be fun. Maybe in future we need a little "Launcher has crashed, do you want to reinstall from SD or pick a different launcher?" dialog in future.

The firmware could, in theory, know a launcher is a launcher by looking at the metadata of the current running offset and act accordingly. Category/file assoc metadata gives us some potentially useful bailout info anyway.

I guess there's no harm- wrt to progress bars and firmware just going for a full screen "I'm doing a thing now, get over it." But slightly less elegant when the running application is a launcher and you want these things to feel like they're part of the launcher.

Daft-Freak:
I could just throw the "shared launcher" stuff into a tiny library (it's mostly headers + one source file).

The launcher detection is currently "is the title \"Launcher\"", guess that could be changed to a category or something (fun if someone calls their game that). There's already some code for reading the metadata of the current game (from the save handling), so should be possible for the firmware to adjust behaviour based on if a launcher is running/crashed.

I think the biggest problem with the "overlays" is the potential screen mode mismatch (if someone's running in lores/palette to save RAM, suddenly using 3 or 4 times more isn't going to end well), but we're still resetting to the launcher to do any commands, so shouldn't be a massive problem unless that gets removed from the tool.

Gadgetoid:
Ooof yes the overlay would have to know what the current screen mode is and - in the case of paletted - understand how to draw into it. I think that's doable up to a point but it's a lot simpler to require a reset to the launcher before doing any system commands. Flashing a game while playing another game is a weird edge case :laughing:

Daft-Freak:
Yeah, and with the current tools it doesn't happen at all... (unless you're a launcher and call `launch`)

Gadgetoid:
Be afraid.  Be very afraid.

Thanks for pulling off another EPIC PR. Looking at all the PRs flying in I figured it would be better to merge this now than consign you to yet more rebase hell. It seems pretty stable and getting people using it should help iron out any outstanding kinks.

Also keen to see some alternate launchers :laughing:

Daft-Freak:
One thing I've just realised is that deleting from flash does nothing. Which is not unexpected as I deleted that bit...

Gadgetoid:
Might need that back :laughing: god knows I seem to get the flash into a terrible state sometimes. I'm still entertaining the thought of having a "Flash Erase" item in the system menu.

PR #451, opened 2020-12-10
Update dma2d_hires_pal_flip() function, fixed bug in dma2d_hires_flip()

zenodante:
fixed bug in dma2d_hires_flip(). It is 0x0001 instead of 0x0000 for RGB888 type. Privous one will copy extra 1/4 memroy out of range.
DMA2D function for hires_pal mode was added. I think it would be even better if we use dma2d for color mode transform in lut mode. It will avoid the ltdc mode changes and lut copy operation.

### DIFF
--- a/32blit-stm32/Src/display.cpp
+++ b/32blit-stm32/Src/display.cpp
@@ -96,15 +96,15 @@ namespace display {
     SCB_CleanInvalidateDCache_by_Addr((uint32_t *)(source.data), 320 * 240 * 3); 
 
     // set the transform type (clear bits 17..16 of control register)
-    DMA2D->CR &= 0xfcff;
+    DMA2D->CR &= 0xfffcffff;
     // set target pixel format (clear bits 3..0 of foreground format register)
-    DMA2D->FGPFCCR &= 0xfff0; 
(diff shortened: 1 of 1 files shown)


Daft-Freak:
Haha, looks like we had roughly the same ideas :smile:

https://github.com/pimoroni/32blit-beta/compare/master...Daft-Freak:dma2d

Gadgetoid:
You snooze you lose 😆 

Thank you! Need to get around to testing and merging this. Bear with me- everything is absolute chaos and I'm surrounded by half packed boxes.

Daft-Freak:
Yeah, I think the only difference in mine is that I de-magic'd the numbers a bit and fixed a comment. (Took me a while to work out what the code was doing...)

Gadgetoid:
A de-magic PR following up this one would be super welcome. Otherwise I can try to beg borrow and steal your commits.

The more understandable the code is, the more chance we have of drive by magic making it better 😆

Gadgetoid:
Pretty sure this nets me 1-1.5ms quicker render times in Ray Caster. Either way, thank you!

PR #453, opened 2020-12-11
Improve stdlibs download/unzip

Daft-Freak:
Check that the download/unzip succeeded, always do the download (it exits early if the file exists with the expected hash) and move the EXISTS check to the extracted files. Still not the most helpful message if `unzip` isn't installed. "No such file or directory"


Gadgetoid:
I think I can patch up the unzip message using FIND_TOOL or something. Lemme see! Thank you!

PR #454, opened 2020-12-11
Use CMake to generate the release zip/tar

Daft-Freak:
This should make it easier to include things that aren't binaries... May also be useful for the boilerplate.


Gadgetoid:
Aha! This should allow me to check off some things from the list. Thank you!

PR #455, opened 2020-12-11
Patch cmake to make moar noise

Gadgetoid:
Adds a status output to the SDL and STM32 executables, since I just notice CMake had a long, quiet period of contemplating during which it wasn't obvious what on earth it was doing. This is just a gentle affirmation to users that Something Is Happening.

I have also tweaked the `stdlib.zip` download message, frustratingly recreating some of the logic that `file` uses to avoid spamming the message repeatedly during the build process.


Gadgetoid:
My code makes me weep, but it's for the good of User Befuddlement Reduction.

Issue #376, opened 2020-09-30
Release Readyness Checklist


Gadgetoid:
With release hardware on the horizon we're ramping up toward a launchable codebase with wrinkles ironed out and a small wishlist of day-1 features:

1. [ ] The blit API should be reasonably stable but as long as  no. 2 is enforced this shouldn't be a huge deal, + known bugs fixed (#354 #344 #221)
2. [ ] The firmware->user code API should be FINAL (or relatively so). Changing this potentially makes binaries incompatible with the firmware.
3. [ ] We should be building/releasing firmware for Beta and Final units (merge #360 and do a test release)
4. [ ] Examples should be tidied up and converted to asset pipeline where appropriate (#365)
5. [ ] Examples should all have metadata config, icon and splash
6. [ ] PIC should be finalised and merged though I'm *keenly* aware how heavily we're leaning on @Daft-Freak for this!
7. [ ] Documentation should be given a shakeup
8. [ ] Releases should include license and documentation (#145)

Nice to have's:

1. [ ] Some method of updating firmware without DFU (#192)

Daft-Freak:
For the firmware API, I think we need a version field there and somewhere in game headers so that it can possibly be extended.

I started on an easing demo somewhere...

For PIC there are a few things left:
 - Need to fix CDC flashing (doesn't handle relocations)
 - Sort out the stdlib build, currently requires that you either build most of GCC, or needs 100MB of libs to be downloadable from somewhere. (~20MB zipped)
 - Fix the travis build
 - There's a script to prepend a header with relocation info, that probably needs something doing to it (https://github.com/Daft-Freak/32blit-beta/blob/pic-returns/tools/gather-relocs)
 - Will need a little adjusting after the metadata stuff

Should replacing the flash tool with the new one be on the list somewhere? I think it's close.

Gadgetoid:
Agreed regarding versioning the firmware API, at least that would give us some way to feed back the incompatibility rather than just crashing outright. Games would need to define a "target fw version" or something, I guess.

* We could earmark CDC flashing as a "dev only" thing and not worry too much about it for now, if it helps.
* stdlib is 😬  I guess it's easy enough just to host the libs somewhere- maybe tagged onto github releases? Guess they'd ideally need to be fetched automagically for building.
* Should `gather-relocs` be part of the 32blit Python tooling?

Added flash tool to the list! Good point.

Daft-Freak:
[...]

PR #456, opened 2020-12-11
Adds LICENSE and README to dist for #376

Gadgetoid:
Could probably stand to have some basic docs here, but some stuff might necessarily be platform dependent, ie:

1. instructions for installing games/flashing firmward in blit release
2. instructions for joining/hosting multiplayer (2 player) in SDL releases
3. ???

### DIFF
--- a/32blit-sdl/CMakeLists.txt
+++ b/32blit-sdl/CMakeLists.txt
@@ -116,6 +116,7 @@ if(DEFINED VIDEO_CAPTURE AND VIDEO_CAPTURE)
 endif()
 
 function(blit_executable NAME SOURCES)
+	message(STATUS "Processing ${NAME}")
 	add_executable(${NAME} MACOSX_BUNDLE ${SOURCES} ${ARGN})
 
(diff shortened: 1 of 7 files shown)


Daft-Freak:
Won't that include those files in every user's game as well? (Since 32blit.cmake is included there too). Also `DESTINATION .` is probably more appropriate (or just not `bin`...)

Gadgetoid:
Weirdly pushing into `bin` seemed to work, all files in the zip are in the base directory which is handy. I guess that's because CPack is eating the subdir? I'm not sure!

Good point re this ending up in ALL THE THINGS. Drat. It's a GOOD THING:tm: that the license does that, but the README really isn't generic enough.

Gadgetoid:
Oh actually it extracts to a `bin` folder, I was just being dumb.

Gadgetoid:
night time mergin' oooh yeah
night time mergin' give ya a scare
night time mergin' don't you dare
night time mergin' don't even care

PR #459, opened 2020-12-13
Avoid passing basic types by reference

Daft-Freak:
Unless I'm missing some case this is _less_ efficient than passing by value... Probably only by a few instructions. Shaves about 300 bytes off Babble. (Lots of calls to `sprite`)

This charge was a find/replace of `const (u?int[12368]+_t|int|float) &` -> `$1 `

### DIFF
--- a/32blit/graphics/blend.cpp
+++ b/32blit/graphics/blend.cpp
@@ -14,19 +14,19 @@
 
 namespace blit {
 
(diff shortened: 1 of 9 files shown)


Gadgetoid:
In lieu of being able to explain why an earth we were doing this, I'm inclined to agree. Someone will tug-of-war it the other way if there are significant performance considerations. Thank you!

PR #462, opened 2020-12-15
Fix some nasty leaks from the API + other API changes

Daft-Freak:
This started as me discovering that calling `list_files` repeatedly would kill the firmware. This was because the data for the `vector` containing the file list was allocated from one heap, and deallocated from the other (well, it _attempted_ to do that, nothing actually gets deallocated...). `get_save_path` had a similar issue.

Since this is already breaking the API, this also adds `blit::is_storage_available`, cleans up `decode_jpeg_file` a bit and merges the two "debug" functions (`debugf` also gets float support by doing the formatting outside of the firmware).

### DIFF
--- a/32blit-sdl/File.cpp
+++ b/32blit-sdl/File.cpp
@@ -86,16 +86,14 @@ uint32_t get_file_length(void *fh)
   return (uint32_t)SDL_RWtell(file);
 }
 
-std::vector<blit::FileInfo> list_files(const std::string &path) {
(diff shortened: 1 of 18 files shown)


Gadgetoid:
Tested and a few examples still work since they don't even touch the API yet. Probably rain heck on some user games, but it's just a binary incompatibility for the most part. These changes are too nice! Onwards and upwards. Thank you.

PR #463, opened 2020-12-15
Mass storage menu rework

Daft-Freak:
Moves mass storage mode enabling to the menu... which was easier said than done. Split from #450.

Adds "Storage Mode" to the menu, with the current status on the right. Since this means that it con now be enabled _anywhere_, also add some checks to prevent enabling storage while files are open... and the other way around.

I haven't tested the auto-disable on eject yet (still broken on my Ubuntu, Windows laptop a bit dead...)

### DIFF
--- a/32blit-stm32/Inc/USBManager.h
+++ b/32blit-stm32/Inc/USBManager.h
@@ -74,6 +74,34 @@ public:
 
 	State GetState(void)
 	{
+		return m_state;
+	}
+
+	const char *GetStateName(void)
+	{
+		return m_stateNames[GetState()];
+	}
+
+	bool IsMSCReady(void)
+	{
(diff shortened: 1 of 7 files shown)


Daft-Freak:
Seems to still disable on eject on Windows...

Gadgetoid:
Yep. I concur. I thought the really s l o w attach was due to my weird PWMing for the battery LED but it doesn't seem to help if I turn it off. This stuff is just complicated!

Gadgetoid:
Works nicely, anyway! Thank you. That cursed button can finally stop accidentally bumping me into mass storage mode 😆

Though now it's free for fun and games and it's labelled "Home" so... I couldn't possibly think of a use for that 😆

PR #464, opened 2020-12-15
Feeble try to bringup USART for RGB LED PWM

Gadgetoid:
I was taking another look over peripherals to drive the system RGB LED, since the current technique of using TIM16 to increment a 9bit counter (with an 8bit duty cycle) is not especially conservative of system resources and is very prone to interruption by other things happening in system.

The intent is to use USART5 to drive the Red channel, and USART8 to drive the Green and Blue channels by flip-flopping TX/RX on the fifo interrupt.

Since the LEDs have a very low maximum duty cycle to keep brightness below annoying levels, it should be possible to round-robin them on the TX/RX pins and use the time one is being output as the off time for the other.

The lucky Red LED gets its own UART TX so it should be comparatively easy to drive. Though I haven't managed to get any interrupt-driven UART output working- yet!

I have proven the LEDs will do what I expect them to do by pushing out non-interrupt driven serial data in the main loop and using TX/RX swap initialised both swapped and unswapped.

TODO (help appreciated, my brain is melting here):

1. [ ] Get a basic interrupt driven UART output stream happening on UART8 and UART5
2. [x] Figure out how to enable/disable SWAP on the fly - current crude tests don't yield much success: `MODIFY_REG((&huart8)->Instance->CR2, USART_CR2_SWAP, UART_ADVFEATURE_SWAP_DISABLE);`
3. [ ] Combine the two to clock out the LED brightness data alternately on UART8 TX/RX
4. [ ] Push out LED brightness and 0x00 on UART5 TX


Gadgetoid:
Disabled the interrupts for now, as I can't seem to get them working. But I've upgraded my basic proof of concept to handle point 2 which may be more complicated with interrupts.

Either way this code now produces a white LED, demonstrating that UART data is being output on three pins "simultaneously". This is a very strong indication that my idea is possible.

Gadgetoid:
I've set up some *very* crude Binary Code Modulation- okay it's not even close to crude, it simply uses the LED value to vary the duty cycle of what ends up a fairly low(ish) frequency pulse. On the upside it sort of works.

Key issues are UART simply not being designed for this, and - I think - the stop bits or start bits are impossible to turn off so even though it's possible to invert the TX/RX logic levels *one* of those bits will always be high, meaning the LED can never be fully off.

Perhaps I chalk this one up as a failed experiment!

zenodante:
I'm wondering why not use the auto reload function for the PWM and just update the auto reload value for RGB value?

Gadgetoid:
No PWM on these pins, afaik! These are the system and not the user LED.

Not sure if PWM can be routed to alternate pins by some mechanism? But right now I just roll Timer 16 with a high-frequency interrupt and basically do soft PWM with a count & compare to set the output pins.

UART was about the only hardware feature on these pins I could envision doing any kind of mostly-hands-off BCM dimming of the LEDs.

zenodante:
OMG, that would be really expansive RGB. As you are doing software PWM, any chance to combine it into systick handler? 1000 time per second could at least create 3bit depth for each color.

zenodante:
Just put following code into SysTick_Handler   

     currentCycle = uwTick &0xf;

    HAL_GPIO_WritePin(LED_CHG_RED_Port, LED_CHG_RED_Pin, charge_led_r <= currentCycle ? GPIO_PIN_RESET : GPIO_PIN_SET);
    HAL_GPIO_WritePin(LED_CHG_GREEN_Port, LED_CHG_GREEN_Pin, charge_led_g <=currentCycle ? GPIO_PIN_RESET : GPIO_PIN_SET);
    HAL_GPIO_WritePin(LED_CHG_BLUE_Port, LED_CHG_BLUE_Pin, charge_led_b <= currentCycle ? GPIO_PIN_RESET : GPIO_PIN_SET);


And the max range for charge_led_r/g/b would be 15. It would be a almost CPU time free RGB solution then.

Gadgetoid:
Might try these modifications to the Tim16 handler, since systick would be about 2000 times faster.

Need some blanking time to keep the LED brightness down- though to be fair this is the system LED and it really ought never to do anything particularly complicated so maybe just using *just* level 1 out of the 15 brightness steps would be fine. If it gets any fancier it'll be too confusing to puzzle out what the LED means anyway.

I should have asked for a bigger inline resistor :laughing:

zenodante:
systick time counting is faster. However it only break 1000 times per second, and it already fix cpu budgets. If doing above code, the PWM is about 64hz, (could be as small as 32hz maybe). If we still use TIM16, it would be another cpu cost for push and pop. M7 pipeline is very long, so this operation is relative more expansive than m4.

zenodante:
As I haven't received my 32blit hardware, I tested on the stm32h750DK board today, the 64Hz PWM by above code works well for 1/16 PWM level. But the 32Hz PWM would show some blink effect. So if the 1/16 strongth is good enough on real hardware, it would be a nice solution for the power led RGB solution.

void SysTick_Handler(void){
  HAL_IncTick();
  currentCycle = uwTick &0xf;

  HAL_GPIO_WritePin(LED_CHG_RED_Port, LED_CHG_RED_Pin, charge_led_r <= currentCycle ? GPIO_PIN_RESET : GPIO_PIN_SET);
  HAL_GPIO_WritePin(LED_CHG_GREEN_Port, LED_CHG_GREEN_Pin, charge_led_g <=currentCycle ? GPIO_PIN_RESET : GPIO_PIN_SET);
  HAL_GPIO_WritePin(LED_CHG_BLUE_Port, LED_CHG_BLUE_Pin, charge_led_b <= currentCycle ? GPIO_PIN_RESET : GPIO_PIN_SET);
}

charge_led_r  would be 1, > or < depends on the hardware design.

Gadgetoid:
You're right. It works really well. Thank you!

For some reason I'd got it in my head (I've got too much going on at the moment) that SysTick was the CPU frequency, rather than the millisecond counter.

`SysTick_Handler` now looks like so, with the `charge_led_r` etc extern'd in `main.h`:

```c++
void SysTick_Handler(void)
{
  /* USER CODE BEGIN SysTick_IRQn 0 */

  /* USER CODE END SysTick_IRQn 0 */
  HAL_IncTick();
  /* USER CODE BEGIN SysTick_IRQn 1 */
  uint32_t currentCycle = HAL_GetTick() & 0x0000000f;

  HAL_GPIO_WritePin(LED_CHG_RED_Port, LED_CHG_RED_Pin, charge_led_r > currentCycle ? GPIO_PIN_RESET : GPIO_PIN_SET);
  HAL_GPIO_WritePin(LED_CHG_GREEN_Port, LED_CHG_GREEN_Pin, charge_led_g > currentCycle ? GPIO_PIN_RESET : GPIO_PIN_SET);
  HAL_GPIO_WritePin(LED_CHG_BLUE_Port, LED_CHG_BLUE_Pin, charge_led_b > currentCycle ? GPIO_PIN_RESET : GPIO_PIN_SET);
  /* USER CODE END SysTick_IRQn 1 */
}
```

It looks good enough to serve as PWM dimming and - though I'm not especially sensitive to these things- I don't notice any flicker.

Gadgetoid:
Closing in favour of #467

PR #465, opened 2020-12-16
Prevent duplicate games

Daft-Freak:
When flashing from the SD card:
 - check for flashed games with the same author/title
 - overwrite if the same or greater size (in blocks)
 - delete it otherwise and flash as normal

When flashing through CDC
 - flash as before
 - clean up old copies afterwards

Also, track free space between games and fix a few bugs. Some of this is a little inefficient, will be easier to fix when the launcher is a separate thing.


Gadgetoid:
Sorely needed. After my attempt at debugging the 32blit tool I found myself with about 20 copies of Doom Fire on my Flash :laughing:

Gadgetoid:
Works a treat! Thank you.

PR #466, opened 2020-12-17
Make Vec2(float, float) explicit

Daft-Freak:
Currently if you do `Point(1)` what that actually does is `Point(Vec2(float(1), 0.0f))` which is... unexpected. Make the `Vec2` constructor that takes one/two floats `explicit` to avoid that.

There are probably more constructors that could be marked explicit, but I've hit this one a few times.


Gadgetoid:
That's bizarre! Thank you.

PR #468, opened 2020-12-17
"Magic" firmware updater

Daft-Freak:
(Thought I opened this earlier, but apparently not)

Tool to update firmware from flash, may need more testing but seems to work. In releases, generates a `magic-[version].blit` containing the firmware from that build. Metadata is also generated to refer to the correct version.

Author is set to "pimoroni" since you probably don't want it to look like firmware updates are coming from some random person :smile:. Could maybe do with a unique splash...

(Hmm, there's a bit of name inconsistency here "fw-flash"/"firmware updater"/"magic")


Gadgetoid:
This going out in releases is just... :cake: ... beautiful! Thank you.

I guess we could solve the name inconsistency by standardizing on "firmware-update" across the board.

The "magic" origins could be alluded to with some clever cover art. :laughing:

Daft-Freak:
It's all `firmware-update`/`firmware-update-vBLAH` now. I'll leave clever cover art to someone with more art skills than me...

Just realised you can do `make firmware-update.flash` to flash the firmware now...

Gadgetoid:
Thank you! I'm giving cover art some thought. Will see what I can whip up tomorrow eve.

PR #473, opened 2020-12-18
Make Platformer actually.... platform!

Gadgetoid:
I stole most of the collision detection logic from Rainbow Ascent for this, and fixes some magic number stuff that made collision not really very easy to reason about.

Right now it's still pretty heinous and we probably need a better example, but it *feels* okay. And maybe that's all that matters :laughing:

### DIFF
--- a/examples/platformer/platformer.cpp
+++ b/examples/platformer/platformer.cpp
@@ -50,13 +50,19 @@ float clamp(float v, float min, float max) {
 struct Player {
   Vec2 vel;
   Vec2 pos;
+  Size size;
   bool flip = false;
 
-  enum {
+  enum State {
     STILL,
     WALKING,
-    JUMPING
+    JUMPING,
+    SWIMMING,
+    CLIMBING
   };
+
+  State state = STILL;
+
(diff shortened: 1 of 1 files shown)


examples/platformer/platformer.cpp
```diff
@@ -125,40 +144,121 @@ struct Player {
...
     static float duration = 0.01f;
-    static Vec2 gravity(0, 9.8f * 5.0f);   // normal gravity is boring!
+    static float jump_velocity = 4.0f;
+    static Vec2 gravity(0, 0.98f / 10.0f);
 
-    if (pressed(Button::DPAD_LEFT)) {
-      vel.x = vel.x - (on_ground() ? ground_acceleration_x : air_acceleration_x);
-    }
+    bool is_on_ground = on_ground();
+    state == STILL;
```
Daft-Freak, line 156: Bit late, but I think _every_ build threw a warning on that. Also a couple of double -> float conversions below.

Gadgetoid: D'oh! I'm pretty sure I saw this warning and forgot to fix it, too. It's... been hectic!

Issue #470, opened 2020-12-18
Graphical corruption when installing a game with no metadata


ali1234:
![screenshot0](https://user-images.githubusercontent.com/651970/102626465-bb3c4880-413e-11eb-90ab-3729be4dd1b0.jpg)

There should be some kind of fallback to prevent this from happening. Either in the launcher, or by providing default metadata during build if metadata.yaml is not present.

ahnlak:
I would say in the launcher (at least); feels risky to rely solely on the build

ali1234:
Agree, but it may not be possible depending on how the metadata is formatted inside the binary.

Daft-Freak:
There's a sort-of fallback now in that the metadata is blank and the title is "game@[block]". Could certainly be improved a bit though.

ali1234:
I just realised that my screenshot doesn't show corruption. That is just what "@" looks like in the tiny font.

PR #475, opened 2020-12-19
Fix free space check

Daft-Freak:
I must have tested this with a file that fit exactly. Another case of a single character breaking everything...

### DIFF
--- a/firmware/firmware.cpp
+++ b/firmware/firmware.cpp
@@ -235,7 +235,7 @@ void scan_flash() {
     } else {
       // fallback "title"
       game.title.resize(20);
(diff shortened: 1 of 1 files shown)


Gadgetoid:
Haha you've been code golfing your game sizes? :laughing: Thank you! I'll tag another release.

Daft-Freak:
Yeah, after shrinking stuff my flash was full of single-block holes.

ahnlak:
So, next up is flash-defrag? :)

Daft-Freak:
Haha, think that's quite far down the list! :smile:

ali1234:
Is that also a fix for #470?

PR #476, opened 2020-12-19
Misc fixes from the "launch" branch

Daft-Freak:
- Avoid a conflict if a game contains a `metadata.cpp` (launcher hits this)
- Delete unused variables and a function from `32blit.cpp`.
- Fix the progress bar for CDC flashing. (started at the flash offset instead of 0)
- Include the size when using `list_files` on "buffer" files


Gadgetoid:
Thank you!

PR #478, opened 2020-12-22
Fix building with older CMake

Daft-Freak:
`TARGET_FILE_BASE_NAME` requires CMake 3.15, we only require 3.8 and debian buster only has 3.13. It's only used for building the updater and the output filename only matters in releases so fallback to the old way if it's not available.


Gadgetoid:
As much as I love CMake the disparity of features available in different versions is *painful*. Doesn't seem to be any way to polyfill them, either. Thank you.

Daft-Freak:
Tempted to add an action that does a configure on the min CMake version we claim to support to prevent this kind of thing happening...

PR #479, opened 2020-12-22
Added information if DFU mode device does not show up in Windows

lenardg:
I have changed my PC since I last developed for 32blit (different laptop, newer Windows version), and to my surprise the device was not recognized in DFU mode. I don't recall if I needed to do this driver update last time, but since now I had to, decided to add info about it to the documentation.


Gadgetoid:
I don't recall needing to faff with drivers, but it's been sooo long that I probably forgot Thanks for getting the docs sorted, much appreciated!

PR #480, opened 2020-12-22
Fix platformer warnings

Daft-Freak:
Unsure if the removed comparison should be an assignment, it seemed to work...


Gadgetoid:
I might need to add it back in as an assignment when I figure out walk cycle animation, right now it's trying to keep a state machine of character state in sync with input actions. Seemed to work well for Rainbow Ascent so I applied it here. We probably need better, teachable, canonical ways of doing stuff like this.

Thanks!

PR #481, opened 2020-12-23
Fixed: holding the menu button will not queue exit event if user code was not running

lenardg:
If I hold down the MENU button for a long time in the main menu, it will queue an exit event, even if the user code is not running. Upon launching the next game, it will auto-quit right away, because the flag was set to terminate. 

Fixed it by adding a check to see if user code is running before flipping the flag.


lenardg:
I presume you (@Gadgetoid) take care of gathering fixes together and releasing them with a new version number for the firmware alongside a changelog? :D

Gadgetoid:
Haha this is very much one of those "how the heck did you find this?" bugs. Nice catch, though.

Yes I am responsible for vetting patches and batching them up into releases- I'm hoping this will go out soon because there are some interesting things in the pipeline that would be better released sooner rather than later.

Gadgetoid:
Totally expected launcher/firmware split to conflict this. Happy days.

Thank you!

PR #482, opened 2020-12-23
Suggestion: use the A B X Y buttons to change volume and backlight keys

lenardg:
I had the displeasure of my 32blit emitting loud sounds while testing the retail unit. Using the sliders controls to mute it seemed to take ages. So, here is my suggestion:

When in the system menu, while backlight or volume is selected, one can use the A B X Y keys to adjust the volume/backlight more quickly. 

* A: increase by 1/4 of the value
* B: decrease by 1/4 of the value
* X: Full value
* Y: Zero value (mute audio / turn off backlight)

I have no suggestions on how to indicate this in the UI, especially when using the big resolution - there is not much space there.


Daft-Freak:
Definitely been in need of a mute button :smile:. 

Hmm, maybe could just make the default step larger and not have two steps. I mean, do we really need 256 volume levels? (That's just me thinking out loud...)

(I would've used a reference for `value`, but that's _extremely_ minor)

lenardg:
Was debating between reference and pointer actually. But was too lazy to ask which people preferred. 🤓

Gadgetoid:
A slightly darkside approach might be to make the volume the angle of the analog sticks direction vector / 360.0f, turning the stick into a little volume dial. How you'd actually facilitate that I don't know :laughing: 

I think a combo of your proposal and Daft's would work quite well. While we *really* could use a mute, and full-volume isn't a bad idea either it's probably sensible just to drop the number of possible volume steps (and brightness steps?) since communicating the purpose of all the buttons might get tricky.

I'll give this a try and see how it feels!

lenardg:
10 steps ought to be enough? Plus the mute. Mute could be a global button in the menu maybe, like X == mute?

Gadgetoid:
Yeah mute/unmute is probably something that deserves its own button- though you may have to add some extra logic to mute without changing the volume level, since that's what I (and most people I guess?) would expect.

The UI in general needs some gentle prodding to remove the slightly wordy and over-informative battery information (maybe elsewhere in a battery diagnostic menu item) and give some of that sweet sweet bottom bar real-estate to a little mute/unmute icon.

lenardg:
I have some ideas how to move this further. I will make some changes this weekend and push it here for you to check out. Stand by :)

Gadgetoid:
Thank you! It's great to have more eyes on the core stuff- it's really going to shape how the 32blit feels.

The firmware menu really needs some rolling credits, too :laughing:

lenardg:
Rolling credits? Uhm, I was thinking about adding something like that, was unsure. But now I consider this as an official request 😃

Gadgetoid:
These shortcuts actually work pretty well, and aren't too tricky to puzzle out even without any UI hints.

I did a quick investigation into the possibility of adding icons, but the "firmware" and gooey system stuff have a weird split at the moment that didn't make it very easy. Need to make some baby steps toward cleaning that all up.

Daft-Freak:
Having a split between "firmware" and the HAL probably doesn't make sense any more seeing as the HAL library is only used by the firmware now... It is a giant mess in there that 200% needs a clean up though.

(I guess the original split was that "firmware" contained stuff that no games would use... but now there's the launcher that does)

Also I just got the firmware to not be loading sprites... :cry:

zenodante:
> Having a split between "firmware" and the HAL probably doesn't make sense any more seeing as the HAL library is only used by the firmware now... It is a giant mess in there that 200% needs a clean up though.
> 
> 
> 
> (I guess the original split was that "firmware" contained stuff that no games would use... but now there's the launcher that does)
> 
> 
> 
> Also I just got the firmware to not be loading sprites... :cry:


Removing stm32 HAL lib by reg operation could save >10KB flash and a lot of cpu time....

lenardg:
I added the following changes now:

* Removed the misc button changes I added previously
* Changed the brightness and volume scale to be 1/12 steps (12 steps total)
* Added X button to toggle MUTE. Works anywhere in the main firmware menu. 
* Muting will remember the previous volume setting. It will show MUTED as the volume
* Removed battery status from main menu
* Added Battery Info menu, that shows the battery info
* Made some groundwork to support many menus. Should be easier to add now. 

Please try it out and tell me what you think

Daft-Freak:
I know the code in 32blit-stm32 is inconsistent but I _think_ it should be 2-space indentation and snake_case filenames... (Just trying to avoid huge "FIX INDENT" changes later...)

Also, this helper may be of interest: https://github.com/pimoroni/32blit-beta/blob/master/launcher-shared/dialog.hpp#L32...

32blit-stm32/Src/SystemMenu/SystemMenuController.hpp
```diff
@@ -0,0 +1,34 @@
+#pragma once
+
+#include "persistence.h"
+
+Pen get_menu_colour(int index);
```
Daft-Freak, line 5: A bit odd the declaration being here when the implementation is in 32blit.cpp.

32blit-stm32/Src/SystemMenu/BatteryMenu.hpp
```diff
@@ -0,0 +1,33 @@
...
+class BatteryMenu final : public blit::Menu {
+public:
+    using blit::Menu::Menu;
+
+    //
+    // Prepare to show the battery menu
+    //
+    void prepare();
+
+    void render_footer();
```
Daft-Freak, line 20: Hmm, Might be generally useful to have a virtual `render_footer` in the Menu class... Could call it from here-ish https://github.com/pimoroni/32blit-beta/blob/master/32blit/engine/menu.hpp#L48

lenardg: Great idea!

32blit-stm32/Src/SystemMenu/BatteryMenu.cpp
```diff
@@ -0,0 +1,121 @@
...
+        system_menu.set_menu(SystemMenus::Firmware);
+    }
+}
+
+//
+// The 'A' button was clicked on a menu item
+//
+void BatteryMenu::item_activated(const Item &item) {
+    // Nothing to do in this menu
+}
```
Daft-Freak, line 107: Not much point in overriding this if we're not doing anything...

lenardg: Will fix

32blit-stm32/Src/SystemMenu/SystemMenuController.hpp
lenardg: I will try to clean this up in a later update / separate PR. (was discussed on Discord that some cleanup might be good)

32blit-stm32/Src/SystemMenu/BatteryMenu.hpp
lenardg: Implemented this.

32blit-stm32/Src/SystemMenu/BatteryMenu.cpp
lenardg: Fixed.

lenardg:
Renamed the files, did some cleanup, improved the menu. 
While I can imagine a lot more cleanup to do, they do not belong here - this is just to implement MUTE and the battery menu. If it's ok with you guys, I do not want to implement anything else in this PR beyond what is here, except of course any fixes that might be needed :)

lenardg:
I added DPAD_RIGHT to open the menu. It felt natural while navigating it :)

32blit-stm32/Src/SystemMenu/firmware_menu.cpp
```diff
@@ -0,0 +1,218 @@
...
+
+    background_colour = get_menu_colour(1);
+    foreground_colour = get_menu_colour(2);
+    bar_background_color = get_menu_colour(3);
+    selected_item_background = get_menu_colour(4);
+    header_background = get_menu_colour(9);
+    header_foreground = get_menu_colour(10);
+
+    display_rect.w = screen.bounds.w;
+    display_rect.h = screen.bounds.h;
```
Daft-Freak, line 64: Indentation is still a bit off in this file

32blit-stm32/Src/SystemMenu/firmware_menu.cpp
```diff
@@ -0,0 +1,218 @@
...
+    switch(item.id) {
+      case BACKLIGHT:
+        draw_slider(Point(bar_x, y + bar_margin), bar_width, persist.backlight, foreground_colour);
+        break;
+      case VOLUME:
+        if ( persist.is_muted ) {
+          auto prev_pen = screen.pen;
+          screen.pen = get_menu_colour(11);
+          screen.text("Muted", minimal_font, Point(screen_width - item_padding_x, y + 1), true, TextAlign::right);
+          screen.pen = prev_pen;
```
Daft-Freak, line 96: Don't think this needs to save/restore the pen, nothing else here does.

32blit-stm32/Src/SystemMenu/system_menu_controller.cpp
```diff
@@ -0,0 +1,136 @@
...
+void SystemMenuController::render_header_battery_status(uint32_t time) {
+  const int screen_width = blit::screen.bounds.w;
+  const int screen_height = blit::screen.bounds.h;
+
+  //   const Pen foreground_colour = get_menu_colour(10);
+  const Pen bar_background_color = get_menu_colour(3);
+
+  BatteryInformation bat = blit_get_battery_info();
+
+  screen.text("bat", minimal_font, Point(screen_width - 80, 4));
```
Daft-Freak, line 80: This text isn't visible any more. Wrong pen?

32blit/engine/menu.hpp
```diff
@@ -127,20 +131,56 @@ namespace blit {
...
       item_rect.x += item_padding_x;
       item_rect.y += item_adjust_y;
       item_rect.h += font.spacing_y; // adjust for alignment
       screen.text(item.label, font, item_rect, true, TextAlign::center_left);
     }
 
+    //
+    // Can be used in a derived class for global updates to the entire menu
+    //
+    virtual void update_menu(uint32_t time) {}
```
Daft-Freak, line 145: Hmm, `Menu::update_menu` is a bit weird, but I don't have a better idea...

32blit/engine/menu.hpp
```diff
@@ -127,20 +131,56 @@ namespace blit {
...
+    }
+
+    //
+    // Called when the footer should be rendered
+    // Derived classes should call base, as this clears the rectangle as well
+    //
+    virtual void render_footer(int x, int y, int w) {
+        screen.pen = header_background;
+        Rect footer_rect(x, y, w, footer_h);
+        screen.rectangle(footer_rect);
```
Daft-Freak, line 181: Indentation

Daft-Freak (commented):
Thanks for putting up with my nitpicks, here's some more :smile: (Okay one of them is a _tiny_ bug)

Firmware size is only up by ~1k... which isn't that bad.

32blit/engine/menu.hpp
lenardg: I wanted it separate from update, but did not want to override update itself. This was the term I came up with. 🤷🏻‍♂️

32blit-stm32/Src/SystemMenu/system_menu_controller.cpp
lenardg: Yes, most likely. When I moved code around, the pens would get set to the background and not reset to foreground. I fixed multiple similar  pen errors, but did not catch this one.

lenardg:
Ok @Daft-Freak everything should now be fixed that you mentioned :)

Gadgetoid:
Right I'm going to M E R G E this because I'm loving that clean footer and the potential of submenus :laughing: 

Hopefully this is the first of many steps toward the firmware/STM32-HAL codebase not making us all want to scream.

Thanks all!

PR #483, opened 2020-12-23
Updated the SDL documentation

lenardg:
Updated the SDL version number in the docs for Visual Studio. 
Also, added a bulleted list - on a quick read it was not immediately obvious there are two libraries to get.


Daft-Freak:
I think we should be consistent about the SDL versions we're recommending/automagically installing... WSL/MinGW docs also specify 2.0.10 and the various CI scripts install that.

Auto-install in `vs/sdl/sdl2-config.cmake` is doing 2.0.12 though... err...

lenardg:
I can update the other references as well and add them to this PR.

lenardg:
Changed every place to be the same version 2.0.14, docs, scripts, everything :)

Gadgetoid:
Thanks for giving the docs a shakeup and trying to get the SDL setup consistent. It's appreciated!

Judging by the test failures on MinGW I think you also need to update -  https://github.com/pimoroni/32blit-beta/blob/master/ci/install_sdl_mingw.sh

lenardg:
I actually did update that file, and now I am trying to figure out the error logs, why it failed here on Github? Strange.

Daft-Freak:
There is something nasty going on with SDL's CMake config in the MinGW build... The library path is very wrong...

Partly caused by this: https://github.com/pimoroni/32blit-beta/blob/master/mingw.toolchain#L4-L5 (library name ends up `SDL2`). But still wrong without that (`libSDL2.dll`, should be `libSDL2.dll.a`)

Daft-Freak:
Okay at least some of it is bugs in SDL's CMake config... (not setting `IMPORTED_IMPLIB` for Windows, I should probably go open a bug about that...)

Even fixing that and some [stuff on the blit side](https://github.com/pimoroni/32blit-beta/compare/master...Daft-Freak:mingw-sdl-fix), it still fails because `-lmingw32` is getting linked in the wrong place...

Gadgetoid:
I think this needs rebased to see if it builds with the latest fixes?

Daft-Freak:
Looks like SDL has fixed the MinGW problem: https://github.com/libsdl-org/SDL/commit/501fe465977a13e3cb5239e965e949b7e5405582...

Daft-Freak:
Yep, 2.0.16 works: https://github.com/32blit/32blit-sdk/compare/master...Daft-Freak:lenardg/docs/sdl

Gadgetoid:
Replaced by #783 - thank you all!

PR #485, opened 2020-12-24
Black magic for lores mode DMA2d flip

zenodante:
This is the balck magic for lores mode DMA2D flip. Now as all mode has used DMA2d for flip, I changed the ltdc buffer to rgb565 type. It will save 75kB ram. 
So the ltdc address shift of the pal mode also changed from 2*240*320 to 1*240*320. The file for memory map should also changed.


zenodante:
It has been tested on stm32H750DK board. I havn't got my 32blit hardware yet....

zenodante:
LTDC (rw)           : ORIGIN = 0x24047C00, LENGTH = 225K
it may change to :
LTDC (rw)           : ORIGIN = 0x2405A800, LENGTH = 150K

Daft-Freak:
Ohh, I like any change that gives me more RAM :smile: After a quick check it seems to be working. The only downside here losing one bit from R/B right?

(Though, could you delete the old code instead of commenting it out? That file's getting a little messy...)

Gadgetoid:
That's an unholy chunk of RAM saved, amazing! Your DMA2d wrangling skills are the stuff of legend.

Tested on my 32blit and it seems to run fine. Had to double-check I'd built the right source since was expecting to see.... something :laughing: 

Thank you!

And - yes - don't  be afraid to delete dead code. It's all in the git history if we need to recover it. Deleting lines of defunct code in favour of your own dark arts is a well-earned privilege.

zenodante:
> Ooh, I like any change that gives me more RAM :smile: After a quick check it seems to be working. The only downside here losing one bit from R/B right?
> 
> 
> 
> (Though, could you delete the old code instead of commenting it out? That file's getting a little messy...)
Sure! Current one is still wasting some cpu time on waiting dma2d to finish the job. I would like to have a command fifo with interrupt refill to parallel the dma2d work with cpu. It may need a pointer in api for fifo.

zenodante:
now it should be fine.

Gadgetoid:
Seems to work well for me! Thank you!

Getting this to work in tandem with some of the firmware tasks would be awesome,  but despite the DMA transfers being effectively blocking I haven't noticed any impact on framerates. The overall cycle count is probably less than the loop, since DMA2d is 1-pixel-per-cycle IIRC?

PR #486, opened 2020-12-24
Really REALLY fix Windows SDL dll copy

Daft-Freak:
Tested with a clean configure and no vcpkg.


Gadgetoid:
Haha cursed vcpkg! Thank you.

PR #487, opened 2020-12-24
Faster Mass Storage mode

Daft-Freak:
(Almost forgot this one...) Increases the packet size from 512 bytes to 4k. On my "fast" card this gets me ~900k both directions.


Daft-Freak:
Looks like `get.pimoroni.com` is having a bad time...

Gadgetoid:
Guess I should update the link to https://github.com/32blit/stdlibs/releases/download/newlib-3.3.0-gcc-9.2.0/stdlibs.zip now the GitHub repo is doing its thang!

Gadgetoid:
Kicked the tests and all is well. Thank you. Will be interesting to see how others fare with this change. Results could be mixed!

PR #488, opened 2020-12-24
Fixes/cleanups for the firmware 

Daft-Freak:
- Removes some now unneeded flash scanning and metadata parsing from the firmware (saving near 3k, mostly from dropping image unpacking)
 - Fixes a few minor bugs (including A/B being backwards on the dialogs)
 - Brings back deleting games from flash (`api.erase_game`)
 - Makes part of the save handling reusable for getting the metadata of whatever is running


Gadgetoid:
All works for me- when I actually build the right source, ha- thank you!

PR #490, opened 2020-12-24
Update stdlib.zip URL to GitHub repo build

Gadgetoid:


Daft-Freak:
It's also an update... that I keep forgetting to test. It seems to work, at least with the tools I'm using... Should test the toolchain you get on debian/raspbian though (think that's the next most common one after the one in ubuntu)

(Other build there matches the current version)

Gadgetoid:
Might have to dig a Pi out of a box somewhere for the latter.

Works for me on Mint, with presumably the same tools as Ubuntu:

```
arm-none-eabi-gcc --version
arm-none-eabi-gcc (15:9-2019-q4-0ubuntu1) 9.2.1 20191025 (release) [ARM/arm-9-branch revision 277599]
```

Daft-Freak:
Yeah, that's the same one. Just did a build on my Pi 400 and it seems fine. (version is `7.3.1 20180622`)

Did find a CMake issue though...

PR #491, opened 2020-12-24
More CMake version fixes

Daft-Freak:
Fix another use of `TARGET_FILE_BASE_NAME` that I missed and bump the required version to 3.9 everywhere. Building with 3.8 would fail with an error about an unrecognised policy immediately. Since nobody seems to have noticed this, it's probably save to increase.

(Tested with an action that configures with a specific version: https://github.com/Daft-Freak/32blit-beta/commit/b65c18cfe559072fda26852ab4923a7db1c27c9e)


Gadgetoid:
Yeesh CMake's minimum required version command finds its way everywhere. Heck of a changeset!

Looks good though, thank you!

PR #493, opened 2020-12-27
Attempt to get a version string for non-release builds

Daft-Freak:
Fixes one of my old TODOs. Results in the tag name if you're on a tag, or something like `v0.1.4-56-gbf167fe7` otherwise.


Gadgetoid:
Seems effective enough- far more traceable if we're jugging dev builds. Thank you!

PR #492, opened 2020-12-26
Swap HOME and MENU functionality for #471

Gadgetoid:
This swaps the functionality of HOME and MENU, bringing them in-line with the shipped button labels.

HOME now does:
1. Long press = Exit game
2. Short press = System menu
3. Press + Reset = Boot to launcher (instead of game)

MENU is now intended for developers to implement pause/game menu features, or whatever they fancy.

PR #494, opened 2020-12-27
Don't toggle system menu in interrupt

Gadgetoid:
Defers the switch in/out of the system menu into the main loop, rather than the interrupt where - potentially - strange things could happen.


Gadgetoid:
This is the patch from #492

PR #497, opened 2020-12-27
Make sure git is executed in the right dir

Daft-Freak:
Just saw this fail when I attempted to build a project...


Gadgetoid:
Doh. Thanks!

PR #498, opened 2020-12-27
Added About menu (was: Scrolling credits)

lenardg:
Added:
* A new about menu with scrolling credits
* Menu now has support for separator items, so they can be spaced better (whoever wrote the scrolling code for the menu, thank you, otherwise this would not have been so great with the smaller resolution)

The scrolling credits can be modified. Just remember about the strings:
* If it starts with a star, it will be green
* If it starts with a dot (.) it will end the credits

The credits will scroll endlessly.

Needed to add a new virtual method to the menu in order to support rendering the credits, but I guess making it more modular / extendable is a good thing?


Gadgetoid:
Can anyone else hear that guttural screaming? Something about firmware size?

Jokes aside - in retrospect I think we could probably transplant this to the launcher in future, *just* because firmware space is super precious.

Since the launcher is required to do *anything* in the 32blit it should be easy enough to come across.

Thank you - though - I think this is super important to celebrate just how much of a community effort this software has been.

Daft-Freak:
Random thought: instead of having a magic string that means "separator", just use a null label?

Gadgetoid:
I'd like to see separators on the main system menu, too, somehow. Things are a bit chaotic at the moment, and could use some subtle grouping. But maybe that's what sub menus are for anyway? :laughing:

Gadgetoid:
Ha this literally adds spacing between some menu items :laughing: 

Thoughts on making "Exit Game" and "Power Off" the bottom two options respectively? Just so a non longpress exit is "Home -> Up -> A"

Guess full menu would (eventually) be something like::

```
Backlight
Volume
Screenshot

Connectivity???
  - DFU Mode
  - Mass Storage
  - Multiplayer Mode???

Battery Info
  - Charge Status
  - Vbus
  - Voltage

Power Off
Exit Game
```

lenardg:
Yeah, null is a better idea. I dunno what I was thinking 🤦🏻‍♂️

lenardg:
Changed the separator in the menu to work on `nullptr` instead of my magic string and string comparison. 

@Gadgetoid would you want me to add that menu structure into this PR as well? Shouldn't take too long. Also, you envision the credits into Launcher -- is that something also for this PR?

I would still add an "about" section to the system menu that could show the _build date_ for example, or other technical things. Like is this a beta device or not, and whatever technical details we want to expose?

32blit-stm32/Src/SystemMenu/firmware_menu.cpp
```diff
@@ -39,13 +39,30 @@ enum MenuItem {
   VOLUME,
   SCREENSHOT,
   DFU,
-  BATTERY_INFO,
   SHIPPING,
   SWITCH_EXE,
   STORAGE,
+  SEPARATOR1,
+  BATTERY_INFO,
+  SEPARATOR2,
```
Daft-Freak, line 47: The menu class doesn't care about the IDs, so the two separators don't need unique ones.

32blit-stm32/Src/SystemMenu/firmware_menu.cpp
```diff
@@ -161,6 +178,12 @@ void FirmwareMenu::update_item(const Item &item) {
       system_menu.set_menu(SystemMenus::Battery);
     }
   }
+  else if (item.id == ABOUT) {
+    if (blit::buttons.released & blit::Button::DPAD_RIGHT)
+    {
+      system_menu.set_menu(SystemMenus::About);
+    }
+  }
```
Daft-Freak, line 186: Few extra newlines here

32blit-stm32/Src/SystemMenu/about_menu.cpp
```diff
@@ -0,0 +1,156 @@
...
+ * source file for About menu
+ *
+ * "The" Credits
+ */
+
+#include "32blit.h"
+#include "32blit.hpp"
+#include "file.hpp"
+#include "USBManager.h"
+#include "i2c-bq24295.h"
```
Daft-Freak, line 11: Most of these aren't needed

32blit-stm32/Src/SystemMenu/about_menu.cpp
```diff
@@ -0,0 +1,156 @@
...
+
+using namespace blit;
+
+#include "about_menu.hpp"
+#include "system_menu_controller.hpp"
+
+#define SPEED 30
+
+// To use the Item class without specifier
+using blit::Menu;
```
Daft-Freak, line 21: Looks like you're still using `Menu::Item` below

32blit-stm32/Src/SystemMenu/about_menu.cpp
```diff
@@ -0,0 +1,156 @@
...
+  "ymauray",
+  "zenodante",
+  "",
+  "*Kickstarted in 2019",
+  "",
+  "No parrots were harmed in",
+  "the making of this product.",
+  "",
+  "*(c) MMXX",
+  "." // credits ends with a string that starts with a dot!
```
Daft-Freak, line 72: Hmm, this could also be a null

Daft-Freak (commented):
(As I seem to be avoiding writing code by reviewing PRs now...)

32blit-stm32/Src/SystemMenu/firmware_menu.cpp
lenardg: I know. But the enum values are/were used to index the items, so there's that. I am gonna fix that too, as part of the menu overhault @Gadgetoid envisioned in one of these comments. Until then, it is like that.  :)

32blit-stm32/Src/SystemMenu/about_menu.cpp
lenardg: 🤦‍♂️

32blit-stm32/Src/SystemMenu/firmware_menu.cpp
Daft-Freak: Ah, right `LAST_COUNT` is still used as the size... (And the thing with `SWITCH_EXE`... which doesn't even do anything useful any more)

lenardg: We could maybe `sizeof(items) / sizeof(Item)` to figure out the size, and use the enum values as IDs only. And no indexing, lookup by ID if needed. Then we would not even need separator IDs or we could just `#define SEPARATOR -1` or something.

32blit-stm32/Src/SystemMenu/about_menu.cpp
Gadgetoid: Story of my life, this!

Gadgetoid:
@lenardg actually totally agree with you about an "About" section in the firmware, we have a pin pulled to determine if the device is beta and we might be able to pull other - possibly useful? - information such as the STM32 UUID (pretty sure there is one) and who knows what else (SD card inserted & capacity?). Probably a basic project title and GitHub URL or web URL wouldn't go amiss either- stuff people might need to know if they got a 32blit from someone sight unseen and wanted to know what the heck it was. :laughing: 

Thanks for all the effort so far! If you're up for porting this to the launcher, I'd welcome it. But I'm not against smashing it into the firmware for the time being with the caveat that things will be shuffled around in future when it's all tidied up.

lenardg:
Would it be ok if I change this PR to have the About menu updates we talk above? Including the support for separator in the menu. 
I will submit a separate PR with the credits added to the Launcher.
@Gadgetoid please confirm :-)

Gadgetoid:
Yep, that makes perfect sense!

lenardg:
Ok @Gadgetoid, have a look at the brand new system menu with 3 submenus and a nifty about screen with device info :-)

Gadgetoid:
This is beautiful, thank you! Just in time for v0.1.5!

Now... waiting.... for...... tests.....

PR #499, opened 2020-12-27
Prevent interaction with invalid games

a2:
In this PR, I've stopped showing hidden files (beginning with `.`) and prevented interaction (play with A, delete with Y) on the "no games found" screen.


Gadgetoid:
Thank you!

PR #502, opened 2020-12-28
Fixes/split 32blit code in firmware

lenardg:
Here is a first stab at separating some code out of the 32blit.cpp file in the firmware. This PR goes for I2C and battery management, which are now located in their own files. Battery state is also wrapped into a class, that serves enum values to the other parts of the firmware, instead of mystical bitfields.


32blit-stm32/Inc/32blit_battery.h
```diff
@@ -7,21 +7,28 @@
...
+    BatteryChargeStatus charge_status;
+    const char * charge_text;
+
+    BatteryVbusStatus vbus_status;
+    const char * vbus_text;
+
+    float voltage;
 };
 
 class BatteryState final {
```
Daft-Freak, line 34: `namespace battery` instead of a class + a single instance? (Also matches display/sound/gpio.hpp, which I think is the direction the code was heading in)

32blit-stm32/Src/32blit/32blit_battery.cpp
```diff
@@ -0,0 +1,129 @@
...
+  auto status = (battery_status >> 4) & 0b11;
+  switch(status){
+    case 0b00: // Not charging
+      return BatteryChargeStatus::NotCharging;
+    case 0b01: // Pre-charge
+      return BatteryChargeStatus::PreCharging;
+    case 0b10: // Fast Charging
+      return BatteryChargeStatus::FastCharging;
+    case 0b11: // Charge Done
+      return BatteryChargeStatus::ChargingComplete;
```
Daft-Freak, line 62: Hmm, I'd define the enum values at these to avoid translating. Then the string getters could also use the enums for even less magic numbers.

32blit-stm32/Src/SystemMenu/battery_menu.cpp
```diff
@@ -5,6 +5,7 @@
  */
 
 #include "32blit.h"
+#include "32blit_battery.h"
 #include "32blit.hpp"
 #include "file.hpp"
 #include "USBManager.h"
```
Daft-Freak, line 11: Little off-topic, but I think some of these includes are unused.

32blit-stm32/Inc/32blit_battery.h
```diff
@@ -0,0 +1,66 @@
...
+
+    // Return the current battery vbus status
+    BatteryVbusStatus get_battery_vbus_status();
+
+  public:
+    // Update the battery status (called only from i2c processing)
+    void update_battery_status ( uint8_t status, uint8_t fault );
+
+    // Update the battery charge value
+    void update_battery_charge ( float charge_value );
```
Daft-Freak, line 53: Maybe remove battery from all the function names...

32blit-stm32/Inc/32blit_battery.h
```diff
@@ -0,0 +1,66 @@
...
+
+  private:
+    const char * get_vbus_status_string();
+    const char * get_charge_status_string();
+
+  private:
+    blit::RunningAverage<float> battery_average;
+    float battery;
+    uint8_t battery_status;
+    uint8_t battery_fault;
```
Daft-Freak, line 63: Namespace would also mean these could be `static` in the cpp

Daft-Freak (commented):
... and I was just about to rebase multiplayer. Anyway, cleaning this up is good!

32blit-stm32/Inc/32blit_battery.h
lenardg: There is currently no namespace used in the firmware part. The engine is namespaced into `blit`. Should we somehow try to make hierarchies here, or just a top level `battery` namespace? 

Or perhaps `blit::firmware::battery` ?

Daft-Freak: There is in the three files I mentioned :smile:

https://github.com/pimoroni/32blit-beta/blob/master/32blit-stm32/Inc/display.hpp#L15
https://github.com/pimoroni/32blit-beta/blob/master/32blit-stm32/Inc/sound.hpp#L16
https://github.com/pimoroni/32blit-beta/blob/master/32blit-stm32/Inc/gpio.hpp

and "soon" https://github.com/Daft-Freak/32blit-beta/blob/minimulti/32blit-stm32/Inc/multiplayer.hpp

Gadgetoid: Yeah I think a `battery` namespace is a nice way to clean things up and it seems - as you suggest - to have emerged as a defacto standard. I think "go with the flow" applies here, even though we've not really sat around and thought about it.

32blit-stm32/Src/32blit/32blit_battery.cpp
Gadgetoid: +1 to reducing magic number count.

Gadgetoid:
@Daft-Freak haha, cleaning it up is good... someone else cleaning it up is better, right? :laughing: (god knows I despair every time I look at the STM32 guts of this thing, and I still somehow understand most of it)

Doing the lord's good work here @lenardg, thank you!

lenardg:
I changed the battery class to be a namespace. 

I did not remove the battery name from the functions, because when it is not a class, I fear it would be all too easy to use the namespace and then have cryptic function calls.

Daft-Freak:
That's true of most namespaces though... and you're less likely to want to use `using` if the names are shorter :smile:

Gadgetoid:
This is true- don't be afraid to drop "Battery" and "battery_" from the front of function names or classes, since right now you can forego "using" and just use `Battery::battery_charge_status` and that's weird in its own way.

lenardg:
There you go :-) Removed the battery word from the function names.

Daft-Freak:
_Tiny_ nitpick: `.hpp` for the new headers. (Okay, most of the files in 32blit-stl32 are `.h`, but most of those are used for pure C HAL stuff...)

lenardg:
Fixed the `.hpp` file thing now. 
Ok, Mr Nitpicker (@Daft-Freak), any more comments? :-P

Gadgetoid:
Boof! Merged. Thank you all!

PR #503, opened 2020-12-28
Fix some indentation in my examples

Daft-Freak:
Can't really go around telling people they're indenting wrong when I've left a giant mess in one of the examples...


Gadgetoid:
It's painfully easy to do. Thank you!

PR #505, opened 2020-12-28
Added Visual Studio projects for the launcher

lenardg:
So Windows freaks like myself can also work on the launcher project ;-)

### DIFF
--- a/32blit-sdl/Main.cpp
+++ b/32blit-sdl/Main.cpp
@@ -138,7 +138,8 @@ void em_loop() {
 
 int main(int argc, char *argv[]) {
 
-	std::cout << "Hello World" << std::endl;
+  std::cout << "32Blit SDL2 runtime" << std::endl;
+  std::cout << "(c) Pimoroni et.al. 2019-2020" << std::endl;
 
 	if (SDL_Init(SDL_INIT_VIDEO|SDL_INIT_GAMECONTROLLER|SDL_INIT_AUDIO) < 0) {
 		fprintf(stderr, "could not initialize SDL2: %s\n", SDL_GetError());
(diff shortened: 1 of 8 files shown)


32blit-sdl/Main.cpp
```diff
@@ -138,7 +138,8 @@ void em_loop() {
 
 int main(int argc, char *argv[]) {
 
-	std::cout << "Hello World" << std::endl;
+  std::cout << "32Blit SDL2 runtime" << std::endl;
+  std::cout << "(c) Pimoroni et.al. 2019-2020" << std::endl;
```
Daft-Freak, line 142: Hah, I was wondering when someone would change that!

Daft-Freak:
+437... Ouch. Something something open folder something.

... anyway...

lenardg:
> +437... Ouch. Something something open folder something.
> 
> ... anyway...

Visual Studio project files are too complex and long 🤷‍♂️

Gadgetoid:
Thank you! I guess someone has to fight for poor ol' Visual Studio's native projects :laughing:

And they are at least better than opening the CMake in some ways.

Daft-Freak:
So, my major problem with the VS sln, is that except in the (very) short term it saves you from _nothing_. You still need to setup CMake to do a device build, which is kind of the whole point... and then you have the same problem we have, you need to maintain duplicate build configs.

So, while you can say it's easier to setup for VS users, unless you for some reason don't want to build for the device it's more work in the end. You might as well setup something that works for both local and device (which could still be VS, cross-builds are possible there). And then there's the fact that CMake is also required to do anything with metadata, and releases...

Maybe I just don't like duplicated work and only being the occasional VS user am probably not the best person to comment on VS .sln vs VS integrated CMake. (Not going to suggest a CMake generated VS .sln, those are... messy)

Anyway, I'll stop complaining now... (probably until it gets in my way _again_)

Issue #506, opened 2020-12-29
Tween behaviour with non zero starting floats.


Loxrie:
Is it tweens intention to always start from 0.0f and progress towards 1.0f?

If I initialise tween like so, for example:

```
sprite_tween.init(tween_linear, 152.0f, 180.0f, 1000, 1);
sprite_tween.start();
```

Then the first "tic" of my tween in an update or render can produce a value of 0 as "from" is not copied to "value" in init() or start().

I was hoping to use it for controlling speed of some sprite animations fixed by 8 pixel widths, so i could use it for simulating trudging or sliding through terrain but it doesn't seem suitable as it's not guaranteed to start at my from value, and it's not guaranteed to reach my destination before finishing.

ahnlak:
I definitely works across different ranges - I use them a fair bit. You're right, though, that `start()` should probably set a default value.

Loxrie:
Excellent good to know I wasn't off the mark.  I did try forcible setting value myself but didn't have much luck.

PR #507, opened 2020-12-29
Set value to <from> when a tween is started

ahnlak:
As mentioned in [this issue](https://github.com/pimoroni/32blit-beta/issues/506), tweens don't set a default value when you `start()` them (or rather, it defaults to 0.0f regardless of the tweening range).

This will force the value to start at `from` when starting the tween.


Gadgetoid:
Awesome, thank you!

PR #508, opened 2020-12-29
Added two more text examples

ahnlak:
Added two more example into `text`:

* drawing to a Point instead of a Rect (and showing the resulting differences in alignment)
* rendering a custom font from a TTF file.

The TTF file is an Open Font License one (that I like for it's 8bit-ness), so I've included the OFL licence file too.


ahnlak:
(I'd just like to apologise for the ridiculous number of commits just to get the VS solution building)

examples/text/text.cpp
```diff
@@ -75,6 +75,33 @@ void render(uint32_t time) {
...
+  text_rect.y += 120;
+
+  screen.pen = Pen(64, 64, 64);
+  screen.rectangle(text_rect);
+
+  text = "This text uses\nan imported\nTrueType font.";
+  text = screen.wrap_text(text, text_rect.w, minimal_font, variable_width);
+
+  screen.pen = Pen(0xFF, 0xFF, 0xFF);
+  screen.text(text, press_start_font, text_rect, variable_width, alignment);
```
Daft-Freak, line 91: Slightly surprised this works, but it's a single argument constructor so I guess it does... Probably not the most efficient to be constructing a `Font` every call though.

(I usually have the asset be `thing_font_data` and have a `const Font thing_font(thing_font_data)` somewhere.)

ahnlak: Huh, you're right. I think it's a bit of a hangover from the early days when fonts were "just" int arrays. Will fix before @Gadgetoid wakes up, because examples should probably Do Things Properly...

Daft-Freak: Might be an idea to mark that constructor `explicit` to avoid that...

Gadgetoid:
Flippin' beautiful! Thank you.

PR #509, opened 2020-12-29
Misc tilemap-related changes

Daft-Freak:
Was attempting to use TileMap for something and forgot about the restriction on the bounds. `constexpr` patch allows doing stuff like
```c++
constexpr blit::Size bg_size(64, 32);
uint8_t bg_tiles[bgSize.w * bgSize.h];
```

Also noticed that the transform flags for TileMap don't match the ones for sprites (horizontal flip and x/y swap are swapped)... (didn't change that here)


Gadgetoid:
Nothing seems to have caught fire! Thank you.

PR #513, opened 2020-12-30
Feature/fix 512 - Moved location of TimerCallback execution.

### LINKED ISSUE #512: Should it be possible to start a Timer again in the callback?
In my TimerCallback I've tried using the Timer param, and an external pointer reference, to reinitialise the timer and start it but it never fires, for example:

```c++

std::vector<ghostState> cycleStates = {
  ghostState::CHASE,
  ghostState::SCATTER,
  ghostState::CHASE,
  ghostState::SCATTER,
  ghostState::CHASE,
  ghostState::SCATTER,
  ghostState::CHASE
};

std::vector<int8_t> cycleTimes = {
  20,
  7,
  20,
  5,
  20,
  5,
  -1
};

int8_t blinky_cycle_index = 0;
int8_t pinky_cycle_index = 0;
int8_t inky_cycle_index = 0;
int8_t clyde_cycle_index = 0;

void manage_ghost_move_state(Timer &t, Ghost *g, int8_t *cycle_index) {
  ghostState nextState = cycleStates[*cycle_index];
  printf("manage_ghost_move_state setting state %d for ghost %s\n", nextState, g->name.c_str());
  g->set_move_state(nextState);
  if (*cycle_index < cycleTimes.size()) {
    int8_t next_cycle_time = cycleTimes[*cycle_index];
    if (next_cycle_time != -1) {
      // Setup next ghost cycle.
      printf("manage_ghost_move_state %s timer pre  dur - %d started - %d loops - %d state - %d\n", g->name.c_str(), g->move_cycle_timer.duration, g->move_cycle_timer.started, g->move_cycle_timer.loops, g->move_cycle_timer.state);
      g->move_cycle_timer.duration = cycleTimes[*cycle_index] * 1000;
      g->move_cycle_timer.loops = 1;
      g->move_cycle_timer.start();
      printf("manage_ghost_move_state %s timer post dur - %d started - %d loops - %d state - %d\n", g->name.c_str(), g->move_cycle_timer.duration, g->move_cycle_timer.started, g->move_cycle_timer.loops, g->move_cycle_timer.state);
      cycle_index++;
    }
  }
}

void blinky_cycle_timer_callback(Timer &t) {
  printf("blinky::move_cycle_timer_callback fired %d.\n", blinky_cycle_index);
  manage_ghost_move_state(t, blinky, &blinky_cycle_index);
}
```

I think this is because the TimerCallback is executed prior to the transition to FINISHED, so my code resets duration and loops then calls start only for the Timer code itself to immediately mark my timer as finished.

I don't know if this is intentional or if there are any knock on effects from modifying this behaviour?  My hope was a simple way of increasingly changing object behaviour over time.

I could of course move this logic out to the Ghost update() cycle if that is more appropriate?

Loxrie:
Proposed Fix #512 

TimerCallback is now executed after t->started set for infinite timers, or after timer has transitioned to FINISHED state for non infinite timers.

This allows the timer callback function to change timer params and restart the timer without it immediately transitioning to FINISHED.

This could be a breaking change for other developers.

### DIFF
--- a/32blit/engine/timer.cpp
+++ b/32blit/engine/timer.cpp
@@ -47,7 +47,6 @@ namespace blit {
       if (t->state == Timer::RUNNING){
         if (time > (t->started + t->duration)) { // timer triggered
           if(t->loops == -1){
-            t->callback(*t);
             t->started = time; // reset the start time correcting for any error
           }
           else
@@ -57,8 +56,8 @@ namespace blit {
             if (t->loops == 0){
               t->state = Timer::FINISHED;
             }
-            t->callback(*t);
           }
+          t->callback(*t);
         }
       }
     }
--- a/32blit/engine/timer.cpp
+++ b/32blit/engine/timer.cpp
@@ -46,9 +46,8 @@ namespace blit {
     for (auto t: timers) {
       if (t->state == Timer::RUNNING){
         if (time > (t->started + t->duration)) { // timer triggered
-          t->callback(*t);
-
           if(t->loops == -1){
+            t->callback(*t);
             t->started = time; // reset the start time correcting for any error
           }
           else
@@ -58,6 +57,7 @@ namespace blit {
             if (t->loops == 0){
               t->state = Timer::FINISHED;
             }
+            t->callback(*t);
           }
         }
       }


Gadgetoid:
Thank you. Breaking or not, I think this is the *right* place to trigger the callback.

PR #514, opened 2020-12-30
Update WSL docs a bit

Daft-Freak:
Adjust the WSL docs a bit based on some recent questions/comments on Discord.

- Move the MinGW warnings down to when we start talking about MinGW
- Split the package lists for MinGW/ARM
- Update toolchain install to be clearer that you don't need any 3rd party packages on Ubuntu 20.04


Gadgetoid:
Ooh thanks for being on the ball with this!

PR #515, opened 2020-12-31
Remove the dma2d busy check in do_render

zenodante:
Even in the lowest render work load, we still don't need to check the dma2d statement. Because the dma2d will always finished much earlier than next vblank interrupt. So we could trigger next vblank directly. When next v blank happen, the dma2d is already ready to use.

image

### DIFF
--- a/32blit-stm32/Src/32blit.cpp
+++ b/32blit-stm32/Src/32blit.cpp
@@ -124,9 +124,10 @@ static void do_render() {
       //if framebuff still occupied by last dma2d flip
     }
     blit::render(blit::now());
-    while (display::is_dma2d_occupied()){
-      //if framebuff still occupied by last dma2d flip
-    }
+    //while (display::is_dma2d_occupied()){
+      //Even in the lowest render work load, we still don't need to check the dma2d statement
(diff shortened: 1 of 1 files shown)


zenodante:
image

zenodante:
I drawed some diagrams to help understanding the whole process
image

zenodante:
Ooops, the first diagram has Chinese language input note。 Here is correct one.
image

zenodante:
So under current setup, for HiRes and HiRes pal mode, the worest situation is the cpu waste the same cycle like the waiting dma2d to finish method(old one). But it is rare. Only when the need render flag set by vblank interrupt and return, the return point is just before the do_render() check happen, then the cpu would need to wait for the framebuff release.  In most cases, the cpu time would be saved for other task.
In LoRes mode, the worest situtation is the cpu waste 1/4 cycles like the waiting dma2d to finish method. (also very rare)

zenodante:
So from above diagram, we may combine the need render signal and framebuff lock signal into one. Sent out the framebuff release signal after the dma2d done. And check it before do_render

PR #518, opened 2020-12-31
Rearrange the signals in render process

zenodante:
Now we don't need to use need_render / framebuff busy / dma2d busy three signal in controlling. Only one need_render signal! And current setup could avoid any waste in cpu time in flip operation. 
Sorry I need to modified the file one by one in web page. The git cli to github in China is really slow :(

### DIFF
--- a/32blit-stm32/Inc/display.hpp
+++ b/32blit-stm32/Inc/display.hpp
@@ -1,4 +1,3 @@
-  
 #ifndef DISPLAY_H
 #define DISPLAY_H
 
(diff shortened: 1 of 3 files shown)


zenodante:
image
now in any situation, cpu won't wait for flip framebuff copy operation done by dma2d. The need_render signal controlled by dma2d statement instead of vblank interrupt.

Gadgetoid:
Thank you! And no worries regarding the terrible GitHub UI. Really happy to have your expertise here! I'd accept your pull requests if you had to mail them to me by post.

PR #520, opened 2020-12-31
Display SD card status and free space in the about menu

Daft-Freak:
For @lenardg :smile: (Okay, I had a rough idea how to do this and had to do it before I forgot...)

Displays if there is an SD card inserted and if it is, the free/total space (in megabytes). Will also display "not mounted" if it was detected but failed to mount.


Gadgetoid:
Bootiful, thank you!

lenardg:
Cool! Have to install this in the afternoon!

PR #522, opened 2021-01-02
Move cpp tool to https://github.com/32blit/tools-cpp

Gadgetoid:
Doesn't actually move the tool, just removes it from here.

Tool files have been filtered into https://github.com/32blit/tools-cpp complete with history.

Readme/license required.


Gadgetoid:
Going to rock 'n' roll with this, hopefully I haven't blown anything subtle up.

I've taken pains to preserve the history of the moved files.

PR #525, opened 2021-01-02
SDL_net related things

Daft-Freak:
Update some docs and fixup the windows DLL copy. (I should really test that...)

Though macOS docs install SDL+friends through `brew`, while we use the frameworks for CI... May want to cover that in the docs too (since it's currently the easiest way to create a redistributable binary)


Gadgetoid:
Thanks for giving the docs some TLC, I know they're not fun to work on. But dang are they useful, specially when it covers stuff like this that I have no visibility on.

PR #527, opened 2021-01-03
Firmware metadata cleanup

Daft-Freak:
Shuffles around the firmware's metadata handling to avoid duplicated work and std::string copying. This shaves ~1.5k off the firmware (and fixes the bug that blew up my launcher...)

Then there's the last commit, which moves the CDC writing into `blit_debug` instead of going through `printf`... saving ~4k. (Yes, this is moving more code into 32blit.cpp)


Gadgetoid:
Nice. Perchance you will leave *me* with some rebasing to do for a change. Well played :laughing: Thank you!

PR #528, opened 2021-01-03
Screenshot Browser (Actually works edition)

Gadgetoid:
I figured since we had information on the running game, it was probably high time to pick a better name than "screenshotx.bmp".

Then- ho ho- we've more or less laid the groundwork for a screenshot browser in the launcher and have plenty of room.

Except there's no `surface.load` that supports bitmaps produced by the `surface.save`. (which should arguably save to `SPRITEPK` or `SPRITERW` anyway).

Enjoy this glitchtastic browser, which proves the concept by smashing the contents of a .bmp file right into the display buffer without a care in the world. Hilarious results ensue:

image


Daft-Freak:
Getting the BMPs saved the right way up would be as simple as:
```diff
diff --git a/32blit/graphics/surface.cpp b/32blit/graphics/surface.cpp
index 53397032..a287ac65 100644
--- a/32blit/graphics/surface.cpp
+++ b/32blit/graphics/surface.cpp
@@ -153,7 +153,7 @@ namespace blit {
     head.data_offset = sizeof(head) + palette_size * 4;
 
     head.w = bounds.w;
-    head.h = bounds.h;
+    head.h = -bounds.h;
     head.bpp = pixel_stride * 8;
     head.image_size = data_size;
 
@@ -170,8 +170,7 @@ namespace blit {
     }
 
     for(int y = 0; y < bounds.h; y++) {
-        // flip y
-        auto in_offset = (bounds.h - 1 - y) * row_stride;
+        auto in_offset = y * row_stride;
 
         if(pixel_stride == 1)
           file.write(offset, row_stride, reinterpret_cast<char *>(data + in_offset));

```
(Learned that trick right after adding that...)

At that point, the only difference between BMP and SPRITERW is the header... (Well and the fact that you can easily open a BMP :smile:)

Gadgetoid:
That y flip trick works a treat, though I can't for love nor money get 160x120 and P-mode screenshots to display correctly.

PR #529, opened 2021-01-04
Surface load/save

Daft-Freak:
Adds SPRITERW support to `Surface::save` and BMP support to `Surface::load(filename)`. Also fixes loading non-paletted SPRITERW.


Gadgetoid:
You are my saviour, thank you. Implementing bitmap loading into the launcher was getting dicey.

PR #530, opened 2021-01-04
Fix mp_mode assignment typo

Daft-Freak:


Gadgetoid:
Ah! This is my kind of C++ :laughing: 

Thank you!

PR #531, opened 2021-01-04
Scrolling project credits into the launcher

lenardg:
I moved my scrolling credits into the Launcher app. The contents of the scrolling is the same as before. Currently launches with the menu button. 

I have not added a menu yet, because I did not want a menu with just one item on it :-) But later if we need a separate menu, we can add the scrolling as one item there. 

Please comment, this is certainly not intended as the finished product -- if you have any comments, please share :-)

### DIFF
--- a/launcher/CMakeLists.txt
+++ b/launcher/CMakeLists.txt
@@ -2,7 +2,7 @@ cmake_minimum_required(VERSION 3.9)
 project (launcher)
 include (../32blit.cmake)
 
-blit_executable(launcher launcher.cpp theme.cpp)
+blit_executable(launcher launcher.cpp theme.cpp credits.cpp)
 target_link_libraries(launcher LauncherShared)
 blit_assets_yaml(launcher assets.yml)
 blit_metadata(launcher metadata.yml)
--- a/launcher/contrib.hpp
+++ b/launcher/contrib.hpp
@@ -1,0 +1,38 @@
+// contrib.hpp
+// 32blit GitHub contributors
+//
+// AUTO GENERATED FILE, PLEASE DO NOT EDIT BY HAND
+//
+#pragma once
+
+static const char * contributors[] = {
+  "a2",
+  "ahnlak",
+  "ali1234",
+  "andreban",
+  "AndrewCapon",
+  "drisc",
+  "illbewithee",
+  "johnmccabe",
+  "LenardG",
+  "LordEidi",
+  "lowfatcode",
+  "Loxrie",
+  "lukesadlerhhl",
+  "mikerr",
+  "mylogon341",
+  "ntwyman",
+  "Ozzard",
+  "Pharap",
+  "shane-powell",
+  "tinwhisker",
+  "trollied",
+  "voidberg",
+  "ymauray",
+  "zenodante",
+  nullptr
+};
+static const char * specialthanks[] = {
+  "Daft-Freak",
+  nullptr
+};
--- a/launcher/credits.cpp
+++ b/launcher/credits.cpp
@@ -10,6 +10,7 @@ using namespace blit;
 
 #include "credits.hpp"
 #include "contrib.hpp"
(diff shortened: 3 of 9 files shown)


Gadgetoid:
I'd like to add Theme stuff into the menu at some point- with some sliders (or something) to set R/G/B values for the various (4 IIRC) colours of the launcher Theme.

Gadgetoid:
+ "gee" (or rabid-inventor) under Pirate Engineer for sure.

Really slick though. Nice work. I *also* like the black bars.

Gadgetoid:
You should probably also use `theme.color_text` for white text and `theme.color_accent` for the green; see:

https://github.com/pimoroni/32blit-beta/blob/a6f5afb987f8555d743310e70d1bead1303f0446/launcher/theme.hpp#L5-L10

This will mean those change with whatever colours the user picks, when theme customization is fully integrated. (right now you need to hexedit `.blit/Launcher/save0`)

Daft-Freak:
*insert joke about getting a special colour*

lowfatcode == Jon, right? Also, should we be asking people what name to use? (Guess they can just PR if they don't like it...)

Gadgetoid:
@Daft-Freak your name should either flash like a retro “Press Start” prompt or do the sine wave thing from rainbow ascent 🤣🤣

a2:
I realize this is self-serving but do we have a way to update the _launcher/credits.cpp_ file as new people contribute to the repo? Or can this be automated somehow 🤖?

lenardg:
What is your special colour @Daft-Freak ? 🤔

lenardg:
> I realize this is self-serving but do we have a way to update the _launcher/credits.cpp_ file as new people contribute to the repo? Or can this be automated somehow 🤖? 

There should be an API that could get the names from Github. But as some might prefer a nickname - like it was pointed out above - we then would need an additional mapping file as well 😀 

That said, I am a big fan of automation. 

@Gadgetoid ?

lenardg:
* Added a python script to update the contributors from Github (REST API). I did not add this to the build process, because it would mean every single time the project is built, it would need to make a REST API call. No more offline builds  ... besides, these do not change that often. So probably enough if @Gadgetoid runs it before every release, or something like that. 
* Added an ignore list to this Python script, so that Pimoroni crew will not end up on the contributors section.
* Added a nickname list to this Python script. This is where people can PR their nicknames if they are dissatisfied with the github username.
* Everyone above 100 contributions will end up on the Special Thanks list, which also gets a special visual upgrade in the credits :-) Go check it out.
* Started using theme colours

lenardg:
* Added rabid-inventor, as suggested above

Gadgetoid:
Automation is an awesome idea and I think it's entirely possible to add this to the release part of the GitHub Actions and basically forget about it. Will check it out!

Thank you!

lenardg:
In case there are changes, after running the script it would be great to check in any changes, so the file is up to date. Have never automated using GitHub, so I don't know how this is done :) 

Anyway, the file `contrib.hpp` is now part of the repository (is committed). I think it should be like this - the automation only updtes it, but is not like the assets that get generated at compile time.

Daft-Freak:
Haha, *has higher contribution count than @Gadgetoid* (Then again I do tend to commit small...)

Attempting to update the file during a release would be... interesting (you'd need to commit it, but releases are based on a tag... that would then be before that commit)

Hmm... could do with some nice fonts...

Gadgetoid:
> Haha, has higher contribution count than @Gadgetoid (Then again I do tend to commit small...)

Probably for the best, we've all seen what happens if I try to do anything. :laughing:

Gadgetoid:
We'll cross the release bridge when we come to it, but I suspect just generating the .hpp at release time is probably the sensible choice, and then stub it out with some default (jokey?) values for people building along at home. I don't think generated data necessarily needs source controlled.

PR #532, opened 2021-01-04
Added gitignore that ignores asset files for text example when compiled with VS

lenardg:
The VS build generates some assets for the _text_ example, that were not gitignored. Added a new gitignore file for the VS examples.

### DIFF
--- a/vs/examples/.gitignore
+++ b/vs/examples/.gitignore
@@ -1,0 +1,9 @@
+#
+# VS specific gitignore
+#
+
+#
+# Text project asset is named differently
+#
+font_asset.cpp
+font_asset.hpp


Daft-Freak:
Ah, different name so https://github.com/pimoroni/32blit-beta/blob/7ec17189544e75f7e17a135d38d13988c7bc2524/vs/.gitignore#L359 didn't catch it...

Gadgetoid:
Should probably just stick this in `32blit-beta/vs/.gitignore`, although this project using "font_asset" against the grain of others is probably going to quietly gnaw away at me :laughing:

lenardg:
OCD is a wonderful feature to have!! 🤣

Gadgetoid:
Will... uh... stick that on the TODO :laughing: 

Thanks!

PR #533, opened 2021-01-05
Use a slightly larger font for titles/description in the launcher

Daft-Freak:
Just dug this out of my pile of stashes... adds lowercase letters. (Maybe not the greatest font, but easily changeable)


Gadgetoid:
Haha darn it! You're all hitting me with the rebasing now, haha!

Gadgetoid:
*MUCH* better for readability and probably accessibility too. I experimented with our fat font and outline font but unfortunately they seem to suck a little bit.

lenardg:
Fat font was uuugly. Tried it for the credits, but did not like what I saw.

Gadgetoid:
:laughing: you're not wrong. Could probably do with embedding some nicer options into the engine. Not like I don't have a bunch of fonts hanging around from other display projects. And @Daft-Freak ... well has practically written the book on converting them now :laughing:

ahnlak:
I've found some nice fonts through GoogleFonts, but it's hard to know the best font size for the built-in ones at least.

PR #534, opened 2021-01-05
Rework toolchain PIC handling

Daft-Freak:
To make libraries work without needing to add the PIC flags manually. Also fixes undefined symbol errors due to linking the stdlibs in the wrong place.


Gadgetoid:
Can confirm, this changes Lua 

Fom: AAAH! I have no idea why this isn't working.

To: THINGS ARE HAPPENING!

That's... awesome! Thank you.

PR #535, opened 2021-01-06
Small game launch related fixes

Daft-Freak:
First one makes sure the save path is for the current game, instead of the previous one in `init`. Fixes a bug where if you exited a game with saves, the launcher would load that as it's theme...

Second one prevents games launched from flash from getting whatever that least launch path was... or junk...


Gadgetoid:
Haha, who'd have thunk the launcher theme support would find the bugs! Thank you.

PR #536, opened 2021-01-07
Feature/credits update

lenardg:
The credits now use the new font and also made the lines a little taller, looks better overall.


Gadgetoid:
Soo much better! Thank you.

Daft-Freak:
Oh no, now I'm in the worse font :laughing: 

I did make an outline version of the font... but ~1.5k for one name seems a little wasteful...
9x10outline_font

Gadgetoid:
1.5k on the launcher is hardly breaking the bank though :laughing: I'm sure we could find some other uses.

PR #538, opened 2021-01-08
Small credits stuff

Daft-Freak:
Update the list before releases, give me lowercase letters and other minor stuff.

Should replace the contrib list with a placeholder and ignore it I guess?


Gadgetoid:
> Should replace the contrib list with a placeholder and ignore it I guess?

Yes! Also thanks for picking up this and making it happen. Much appreciated! You deserve that nicer font, haha.

Might be worth adding a license/credit note for these fonts and moving all the launcher assets into the root assets dir. Some stuff people may well want to reuse and get better visibility of.

Daft-Freak:
Hmm, need to have _something_ in the placeholder, empty just makes the credits look a bit broken...

Licence for the fonts can be whatever, though I'm not seeing a licence for the existing assets anywhere...

What's the recommended way to use the "standard" assets anyway? I'd assume copying them, but I've seen a couple of repos with `../32blit-beta/assets/blah` (or similar) in their asset config... which explodes a little if you're using a different dir structure...


(Also, looking at the credits script, I hope the >= 100 contributions for "Special Thanks" doesn't encourage anyone to do a bunch of tiny commits...)

Gadgetoid:
Could probably raise the bar for "Special Thanks" I guess it should probably scale with time otherwise *everyone* (or at least everyone actively contributing) would eventually end up there. Definitely a "we'll cross that bridge when we come to it" thing.

There's no pattern for using standard assets, but I guess outside of `32blit-beta` you should *probably* copy them.

PR #540, opened 2021-01-08
Placeholder the credits

Daft-Freak:
One of the leftover bits from #538. I was going to `.gitignore` it, but that doesn't work...


Daft-Freak:
... and VS Code trimmed all the whitespace anyway...

Gadgetoid:
Thank you!

PR #541, opened 2021-01-09
hires -> lores mode change fix

Daft-Freak:
Moves the mode update even later (after the next render), which better matches what I intended to do in the first place. (Not flip in the new mode until something was rendered in it).

Fixes doing a hires -> lores switch.

Also changes a bunch of whitespace since VS Code refuses to leave that alone now...

### DIFF
--- a/32blit-stm32/Src/display.cpp
+++ b/32blit-stm32/Src/display.cpp
@@ -88,6 +88,12 @@ namespace display {
   }
 
   void enable_vblank_interrupt() {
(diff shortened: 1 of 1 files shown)


Gadgetoid:
Since finding "Hide whitespace changes" in the diff viewer, I've not worried about whitespace changes leaking into changesets. See: https://github.com/pimoroni/32blit-beta/pull/541/files?diff=split&w=1

Tidy!

Thanks for leaping in and fixing this.

PR #543, opened 2021-01-10
Misc BMP/screenshot/surface fixes

Daft-Freak:
Mostly discovered by all the random junk on my SD... First two are the most important, the other two just expand BMP support a little.

Not fixed:
- Having an image that doesn't fit in the static buffer on your SD card explodes.
- Paletted BMPs can't have alpha, which is not really a problem for screenshots, but the "alpha" field is written as 0 from most editors...

### DIFF
--- a/32blit/graphics/surface.cpp
+++ b/32blit/graphics/surface.cpp
@@ -59,7 +59,7 @@ namespace blit {
    *
    * \return `Surface` containing loaded data or `nullptr` if the image was invalid
    */
-  Surface *Surface::load(const packed_image *image, uint8_t *data) {
+  Surface *Surface::load(const packed_image *image, uint8_t *data, size_t data_size) {
     if(memcmp(image->type, "SPRITEPK", 8) != 0 && memcmp(image->type, "SPRITERW", 8) != 0 && memcmp(image->type, "SPRITERL", 8) != 0)
       return nullptr;
 
@@ -67,7 +67,7 @@ namespace blit {
       return nullptr;
 
     File file((const uint8_t *)image, image->byte_count);
-    return load_from_packed(file, data);
+    return load_from_packed(file, data, data_size, false);
   }
 
   /**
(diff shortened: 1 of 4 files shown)


Gadgetoid:
Ooh thank you!

Ooof regarding 0 alpha- basically initialising your image with a trash palette because you can’t tell if it’s *really* supposed to be 0 or just 0’d by an editor that’s only dealing with RGB?

Could *possibly* add an out for images that don’t fit in the buffer. Anything much larger is probably going to break anyway. Surface::load(string, buf) would need a Surface::load(string, buf, size) perhaps, and baulk with nullptr if w*h*d exceeds the given buffer size. Bit like “snprintf”?

Daft-Freak:
I think the alpha field is defined as "must be zero". I suppose since we already have to loop through the palette to swap R/B, we could check if every entry is 0 and set them to 255...

Also passing the size of the buffer is probably a good idea, most of the time you're passing in a static buffer so it would just be `load(..., buffer, sizeof(buffer))`... or just default it to 0.

Gadgetoid:
Fixes or... in the case of my code... horrible breaks:

```diff
diff --git a/32blit/graphics/surface.hpp b/32blit/graphics/surface.hpp
index 5ff0da1..96eda0d 100644
--- a/32blit/graphics/surface.hpp
+++ b/32blit/graphics/surface.hpp
@@ -124,7 +124,7 @@ namespace blit {
 
     static Surface *load(const packed_image *image, uint8_t *data, size_t data_size);
     static Surface *load(const packed_image *image) {return load(image, nullptr, 0);};
-    static Surface *load(const uint8_t *image, uint8_t *data, size_t data_size) {return load((packed_image *)image, nullptr, 0);};
+    static Surface *load(const uint8_t *image, uint8_t *data, size_t data_size) {return load((packed_image *)image, data, data_size);};
     static Surface *load(const uint8_t *image) {return load((packed_image *)image, nullptr, 0);};
 
     static Surface *load_read_only(const packed_image *image);
 ```

I can probably just apply this as I merge.

Daft-Freak:
Nope, I don't see a mistake there :wink:

Gadgetoid:
This will serve us right for not just removing one of those overloads altogether.

Gadgetoid:
Thank you!

PR #545, opened 2021-01-10
Automatically install firmware/launcher updates

Daft-Freak:
Checks for `firmware-update.blit` at init and if found, flashes deletes and starts it. Then does the same for `launcher.blit`.

Also cleans up any old firmware updaters so you don't end up with a screen full of them...

### DIFF
--- a/firmware/firmware.cpp
+++ b/firmware/firmware.cpp
@@ -405,16 +405,22 @@ static void start_launcher() {
 }
 
 // used for updates
-static void launch_and_delete(const char *path) {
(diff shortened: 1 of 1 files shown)


Gadgetoid:
Ha the flashing `launcher.blit` happens so quickly after updating the firmware that I barely see it happen at all. Makes the whole process feel like a one-step update, even though it isn't. Really slick. Thank you!

Gadgetoid:
Works great from shipped firmware and v0.1.7, happy days!

PR #546, opened 2021-01-11
Only delete old firmware when not launching something

Gadgetoid:
I was having an issue where running firmware-update-v0.1.8.blit, or `make firmware-update.flash` would jump my 32blit into a semi broken state.

I believe - when not launching `firmware-update.blit` from SD card - that the firmware updater was being deleted from flash before it had a chance to be run.

This patch changes that behaviour so old updates are only deleted when the reset_target is 0.

Users on v0.1.8 will have to copy `firmware-update.blit` to SD card and reset to update, either via `32blit firmware-update.blit` or just a file copy over MSD/direct onto the card.

### DIFF
--- a/firmware/firmware.cpp
+++ b/firmware/firmware.cpp
@@ -253,7 +253,7 @@ void scan_flash() {
         launcher_offset = offset;
 
       // remove old firmware updates
-      if(strcmp(game.title, "Firmware Updater") == 0) {
+      if(strcmp(game.title, "Firmware Updater") == 0 && persist.reset_target == 0) {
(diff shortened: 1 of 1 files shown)


Daft-Freak:
Hmm, yeah I can see why that would break. I forgot about the reset that happens for game -> game switches...

That 0 should be `prtFirmware` though, or it could be `offset != persist.last_game_offset`. (Which reminds me that `erase_flash_game` really needs that check)

Gadgetoid:
I tried `offset != persist.last_game_offset` and for some reason it didn't fix the problem. That might have been a me issue and not a firmware issue though.

PR #547, opened 2021-01-11
Assorted tweaks to firmware->launcher behaviour

Gadgetoid:
This drops some code that was rendered obsolete by #545 and switches launcher identification from `title` to `category`.

If you want to write a custom launcher, it *must* have a category of "launcher". Subsequent to this change the title can be anything, so swapping in new launcher should be slightly easier.

We might need a special case for "launcher" type apps, expanding on the "delete old launcher" behaviour to *always* delete and replace a launcher when any other launcher is flashed. While it would be fine to have *multiple* launchers flashed at once, it would always be the first one in Flash that wins so it's a little redundant.

### DIFF
--- a/firmware/firmware.cpp
+++ b/firmware/firmware.cpp
@@ -122,6 +122,7 @@ bool parse_flash_metadata(uint32_t offset, GameInfo &info) {
 bool parse_file_metadata(FIL &fh, GameInfo &info) {
   BlitGameHeader header;
   UINT bytes_read;
+  bool result = false;
   f_lseek(&fh, 0);
   f_read(&fh, &header, sizeof(header), &bytes_read);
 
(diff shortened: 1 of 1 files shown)


Daft-Freak:
Haha, that was on my TODO...

Gadgetoid:
Okay I added some cases for flash over CDC and flash from SD to catch a launcher-type app being flashed and remove the old one. I tested this by hex-editing `launcher.blit` to change the title to `launcher2` and `launcher3`, then reflashing them in various ways.

My main bugbear with this approach, and I think the firmware codebase has quite a bit of this, is that it's doing approximately the same thing in three locations.

Gadgetoid:
> Haha, that was on my TODO...

Hopefully I've achieved something approximately as good as what you would have written. ... which is more than can be said for that sentence.

Daft-Freak:
> My main bugbear with this approach, and I think the firmware codebase has quite a bit of this, is that it's doing approximately the same thing in three locations.

Ugh yeah, SD flash has the metadata to do it properly at the start (place the new game on top of the old one), while CDC flash has to wait until the end... and then the update flashing code is a partial copy of `launch_game_from_sd` because I needed to delete the file in the middle. (of course, not I have refactoring ideas...)

Gadgetoid:
Ooof refactoring :grimacing:  it's the work of saints, but don't let it keep you from anything more exciting :laughing: 

I'm just glad to have gone through the firmware and come out the other side mostly unscathed.

PR #548, opened 2021-01-11
Some firmware refactoring

Daft-Freak:
An attempt to de-duplicate and reorganise the flashing code in the firmware. There are a lot of things this could potentially break...

I've tested (at least I think):

SD flash installed
SD flash not installed
SD flash different version

CDC flash not installed
CDC flash installed
CDC flash different version
CDC flash launcher

CDC save

firmware update from SD
launcher update from SD

### DIFF
--- a/firmware-update/metadata.yml.in
+++ b/firmware-update/metadata.yml.in
@@ -7,4 +7,4 @@ icon:
   file: ../examples/no-icon.png
 version: ${FIRMWARE_VERSION}
 url: https://github.com/pimoroni/32blit-beta
-category: utility
+category: firmware
--- a/firmware/firmware.cpp
+++ b/firmware/firmware.cpp
@@ -3,7 +3,6 @@
 #include <cmath>
 #include "quadspi.h"
 #include "CDCCommandStream.h"
-#include "USBManager.h"
 #include "usbd_cdc_if.h"
 #include "file.hpp"
 #include "executable.hpp"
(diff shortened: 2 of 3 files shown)


Gadgetoid:
I think the other thing I'd change is making the category for "Firmware Updater" just "firmware" and scanning for that instead of the title. Probably more for consistency, but I guess it also permits a rename (appending version number?) and saves uh... a whopping 8 bytes?

Gadgetoid:
Okay really weird edge case here:

1. hex edit `launcher.blit` and change the version, I used `v1.0.1`
2. flash is to *SD card*
3. Check "root" and observe that the Launcher version v1.0.1 is present
4. reset
5. `launcher.blit` is deleted from the SD card, but *not* installed

I guess it should either
a. not delete `launcher.blit` at all it hasn't installed or
or
b. install it if it's newer, but then still a?

Daft-Freak:
I think that's because the CRC is the same, so it assumes it's already installed (version field is pretty much ignored). If it doesn't delete the launcher you could end up with it doing an auto-downgrade if you later installed a newer launcher through another method...

Gadgetoid:
Aaaah! That makes sense.

Gadgetoid:
You're right- leaving `launcher.blit` hanging around is probably a bad idea- if I delete the running launcher it will immediately install from that file, and who knows what might be there :laughing:

Gadgetoid:
This passed any test I could contrive for it, thank you! I hope this will pay off in future when someone has an idea for the firmware and no longer runs screaming from our finest organically grown code :laughing:

PR #549, opened 2021-01-12
Helpful things

Gadgetoid:
Helpers and additional operator support on some basic types. Probably need to flesh this out a lot more, but I'm adding these as my sprite-editor code demands it because I hate repeating myself.

Is there any drawback to this?

Also strikes me that Pen should cast to a `uint32` and just do a direct comparison, but for love nor money I couldn't get the thing to cast. Maybe the compiler is clever enough to optimize my long-winded comparison. Maybe I should just be casting and comparing pens in my code?

### DIFF
--- a/32blit/graphics/surface.hpp
+++ b/32blit/graphics/surface.hpp
@@ -74,6 +74,7 @@ namespace blit {
     uint8_t b = 0;
     uint8_t a = 0;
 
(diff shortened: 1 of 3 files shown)


Daft-Freak:
> Maybe the compiler is clever enough to optimize my long-winded comparison.

Yep, compiles to a single instruction. In fact, most operations on `Pen`s do.

Guess `Rect::size` should be closer to the four point methods... and the divide on size should have a `/=` to be consistent with `*`. (But that one's also a float... hmm)

PR #551, opened 2021-01-14
SD/SPI timeouts

Daft-Freak:
Removes a bunch of the timeouts from the SD code that didn't do anything useful and fixes the remaining ones to actually work.

Now repeatedly inserting a card doesn't hang sometimes... (If something goes wrong in mass storage mode, it'll probably still hang)


Gadgetoid:
Just did criminally horrible things to my SD card (pinging it across the room a few times, too) and haven't managed to lock it up prior to these fixes. Beforehand it was third time's a charm, so I'm pretty sure this helps!

Thank you.

PR #552, opened 2021-01-14
Display 128x128 spritesheets (or other images) at correct aspect

Gadgetoid:
Zoomed "view" somewhat sucks here, but I'd like to implement file association support so I can hijack 128x128 spritesheets and open them in the Sprite Editor. This is step 1.

### DIFF
--- a/launcher/launcher.cpp
+++ b/launcher/launcher.cpp
@@ -196,8 +196,7 @@ void load_file_list(std::string directory) {
       continue;
     }
 
-    // spritepk and spriterw are old, transitional long exts, sspk and ssrw are the file-association compatible short names
-    if(ext == "bmp" || ext == "spritepk" || ext == "spriterw" || ext == "sspk" || ext == "ssrw") {
+    if(ext == "bmp" || ext == "blim") {
       GameInfo game;
       game.type = GameType::screenshot;
       game.title = file.name.substr(0, file.name.length() - ext.length() - 1);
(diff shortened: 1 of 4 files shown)


launcher/launcher.cpp
```diff
@@ -196,12 +196,16 @@ void load_file_list(std::string directory) {
       continue;
     }
 
-    if(ext == "bmp" || ext == "spritepk" || ext == "spriterw") {
+    // spritepk and spriterw are old, transitional long exts, sspk and ssrw are the file-association compatible short names
+    if(ext == "bmp" || ext == "spritepk" || ext == "spriterw" || ext == "sspk" || ext == "ssrw") {
```
Daft-Freak, line 200: I don't think spritepk was even referenced anywhere until now... and spriterw is only a few days old, don't think much would break if you just dropped those.

Gadgetoid: True, I'm forging new ground actually using these features :laughing:

Gadgetoid: Tbf I'm not entirely keen on "sspk" and "ssrw" but it's something. It should just be "ss32" or something, and the format is described in the file. Then `.save()` can take `packed` and `rle` args for when we're feeling really bored and want to implement them.

Daft-Freak: Yeah, all the info is already in the header for loading... though putting packing code in `save` would end up including it in the firmware... (Solve that one later I guess)

PR #553, opened 2021-01-15
Added basic highscore tracking for scrolly-tile

ThePythonator:
Daft-Freak mentioned this a while ago, that highscore tracking would be cool.
This adds basic highscore tracking to scroll-tile (ignoring the level seed).

### DIFF
--- a/examples/scrolly-tile/scrolly-tile.cpp
+++ b/examples/scrolly-tile/scrolly-tile.cpp
@@ -67,6 +67,11 @@ uint32_t current_row = 0;
 Timer state_update;
 Point tile_offset(0, 0);
 
+struct SaveData {
+  uint32_t highscore;
+} saveData;
+
+uint32_t highscore = 0;
 
 Vec2 player_position(80.0f, SCREEN_H - PLAYER_H);
 Vec2 player_velocity(0.0f, 0.0f);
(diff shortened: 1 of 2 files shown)


examples/scrolly-tile/scrolly-tile.cpp
```diff
@@ -107,6 +112,10 @@ using tile_callback = uint8_t (*)(uint8_t tile, uint8_t x, uint8_t y, void *args
 uint32_t prng_lfsr = 0;
 const uint16_t prng_tap = 0x74b8;
 
+uint32_t max(uint32_t a, uint32_t b) {
+  return a > b ? a : b;
+}
+
```
Daft-Freak, line 118: Think you want `std::max` from `#include <algorithm>`

ThePythonator: Didn't know that existed! 😄

Gadgetoid: You can probably drop this, plus the "highscore" variable altogether and just do:

```c++
if (player_progress > saveData.highscore) {
    saveData.highscore = player_progress;
}
```

No need to keep another copy of the score, the struct has you covered.

Daft-Freak: Hmm, yeah... should've read the rest! Anyway, that should really be `save_data` for consistency.

ThePythonator: Need to update it when displaying the highscore as well then

ThePythonator: I'll fix this tomorrow I think

Gadgetoid:
Awesome, thank you! Will aim to get this tested & merged ASAP since I want to tag a new release SOON. Somewhat caught up in Sprite Editor :laughing:

Gadgetoid:
Why did I make this awful hard game? :laughing: 

Thank you! Works a treat.

PR #555, opened 2021-01-15
Clear category if no BLITTYPE metadata

Daft-Freak:
Prevents random old stuff found after the launcher being used as the launcher.


Gadgetoid:
I *knew* I wasn't going mad from too many late evenings programming. Thank you!

PR #556, opened 2021-01-17
Update Windows-VisualStudio.md

dstretch:
Added the location of the Embedded and IoT development tools option as I couldn't find it and spent ages looking!  Hopefully this will save someone else some time in the future.


Gadgetoid:
Thank you! You've saved some poor future soul those same ages.

PR #557, opened 2021-01-18
additional step for macOS

titimoby:
Two elements needed to be installed.
I added the brew commands in the md file.


Daft-Freak:
The fun thing about these is they're only _sometimes_ needed, if pip manages to install a prebuilt PIL/freetype-py then everything just works...

Gadgetoid:
And you've hit upon everything I hate about `pip` in a single sentence. :laughing: 

Thank you!

PR #558, opened 2021-01-18
Use HSE as clock source

Daft-Freak:
Hasn't exploded.

Seems more accurate than HSI, at least comparing Super Mario Land vs the same emulator on a PC and a real GB is the same speed now. Most clocks are the same (I think?) ADC is 2MHz lower.

Not that I have any idea what I'm doing with these...


Gadgetoid:
These values check out when I plug them into STM32 CubeMX. I think this is a holdout from when we had no external OSC- that was added a bunch of prototypes ago for USB. Clock config just never caught up!

Thank you!

PR #560, opened 2021-01-18
Flash tmp area/cache

Daft-Freak:
Reserves some space at the end of flash for temporary files and adds some API for using it (only handles one file in there, could be improved).

```c++
File f("path/to/thing", OpenMode::read | OpenMode::cached);
// file has been copied to flash if possible
// cache space is reserved until close
```

Also, stops audio while ext flash is unavailable to prevent games using audio callbacks from crashing.

(WIP?, at least I don't feel like I've spent enough time on this...)


Gadgetoid:
Release is tagged so I feel secure unleashing this on unsuspecting people now! Thank you. No doubt *something* will surface.

Issue #344, opened 2020-09-11
Random API inconsistencies


Daft-Freak:
Some little inconsistencies I keep noticing... then forgetting about:
 - `struct packed_image`
 - `AudioChannel::callback_waveBufferRefresh`
 - Some enum values are lowercase (`ScreenMode`, `OpenMode`, `TextAlign`), most are UPPERCASE. Looks like most of the lowercase ones are mine... Oops.
 - `blit::debug(std::string)` `blit::debugf(const char *, ...)`

Daft-Freak:
This one's in the API:
`set_screen_palette(const Pen *colours, int num_cols)` - both the implementations copy to `screen.palette`, could be replaced with something like `update_screen_palette(int num_cols)` (default to 256?). That gets screen palette updates closer to any other surface.

Gadgetoid:
Assigning @lowfatcode here (for when he gets out of the other end of the mad chaos in the warehouse) to have a little sweep over the API. Along with #354

Daft-Freak:
`TileMap`(tilemap-test, docs) and `Map` (flight, platformer, raycaster).

`Map` has layers, but uses std::vectors for storage (huge compile times for flight, more RAM/copying). MapLayer can be passed to `mode7`.

`TileMap` requires the bounds to be a power of two and can repeat/define a "default" tile.

Daft-Freak:
Ugh, I accidentally made the enum problem worse: https://github.com/32blit/32blit-sdk/blob/d7e545ddd17919fe2a190f069c38c99e0c9cb743/32blit/graphics/tilemap.hpp#L13-L16

PR #562, opened 2021-01-20
Rework audio wave channel

Daft-Freak:
- Rename the callback for more consistency (#344)
- Pass the channel to the callback, helps if you're using more than one channel... and just makes sense
- Stop doing << 8 on samples. Now you can use 16-bit samples without reducing the volume 256x
- Make the callback optional to allow using a fixed pattern.
- `wave_callback_arg` is now a generic `user_data` pointer.

Changes for users are...

Old:
```c++
channels[0].wave_callback_arg = some_ptr;
channels[0].callback_waveBufferRefresh = &callback;

//...

void callback(void *data) {
  //...
  for(int i = 0; i < 64; i++)
    channels[0].wave_buffer[i] = some_sample;
}
```
New:
```c++
channels[0].user_data = some_ptr;
channels[0].wave_buffer_callback = &callback;

//...

void callback(AudioChannel &channel) {
  auto data = channel->user_data;
  //...
  for(int i = 0; i < 64; i++)
    channel.wave_buffer[i] = some_sample;
}
```

Also, if you're using 8-bit samples you need to multiply them by 256 now. (If you're using 16-bit samples you can stop dividing your volume by 256).

This will also cause API breakage, but will only affect games that use the WAVE waveform (callback arg will be wrong).

### DIFF
--- a/32blit/audio/audio.cpp
+++ b/32blit/audio/audio.cpp
@@ -123,10 +123,11 @@ namespace blit {
         }
 
         if(channel.waveforms & Waveform::WAVE) {
-          channel_sample += channel.wave_buffer[channel.wave_buf_pos] << 8;
-          if (++channel.wave_buf_pos == 64) { // If the position is at the end, reset and hit up callback for more.
+          channel_sample += channel.wave_buffer[channel.wave_buf_pos];
+          if (++channel.wave_buf_pos == 64) {
             channel.wave_buf_pos = 0;
-            (*channel.callback_waveBufferRefresh)(channel.wave_callback_arg);
+            if(channel.wave_buffer_callback)
+                channel.wave_buffer_callback(channel);
           }
           waveform_count++;
         }
--- a/32blit/audio/audio.hpp
+++ b/32blit/audio/audio.hpp
@@ -2,9 +2,9 @@
 
 #include <cstdint>
 
-namespace blit {  
+namespace blit {
 
-  // The duration a note is played is determined by the amount of attack, 
+  // The duration a note is played is determined by the amount of attack,
   // decay, and release, combined with the length of the note as defined by
   // the user.
   //
@@ -14,14 +14,14 @@ namespace blit {
   // - Release: number of milliseconds it takes for a note to reduce to zero volume after it has ended
   //
   // Attack (750ms) - Decay (500ms) -------- Sustain ----- Release (250ms)
-  // 
+  //
   //                +         +                                  +    +
   //                |         |                                  |    |
   //                |         |                                  |    |
   //                |         |                                  |    |
   //                v         v                                  v    v
   // 0ms               1000ms              2000ms              3000ms              4000ms
-  //                                                                                  
+  //
   // |              XXXX |                   |                   |                   |
   // |             X    X|XX                 |                   |                   |
   // |            X      |  XXX              |                   |                   |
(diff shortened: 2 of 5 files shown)


zenodante:
Should we also make some short playback buffer on 32blit hardware instead of generating 1 sample per interupt? Currently a 22k Hz interrupt for sound system is too frequent. A ~50Hz interupt for filling playback DMA call would make the sound playback more accurate and stable.

Gadgetoid:
@zenodante probably - yes - although we'd need - I guess - to size the wave buffer appropriately (441 samples?) and the *output* buffer appropriately.

IIRC an earlier build I had a buffer set up with a memory -> peripheral DMA, and would run the audio engine on the DMA-half-complete callback.

My reasoning for switching to a 22050Hz native sample rate with the audio engine invoked on every sample (IIRC) is due to wanting to mirror the capabilities of retro audio engines in the hands of expert composers. IE: you can mash the audio channel settings at high frequencies to create new effects and sounds that the sound system designer may not have anticipated.

That said I'm not sure this is how - for example - the more advanced C64 SID tunes actually worked. Is a 50Hz "Vblank" interrupt equivalent actually... fine?

This then begs the question- how does the end-user control the audio engine. If we're dealing with a sample buffer and a 50Hz update rate then a majority of changes to the engine in `update` would be moot. You'd almost have to treat audio like a `render` operation and do it there... maybe that's okay?

Daft-Freak:
Currently you can only really have 100Hz updates anyway. You could try to do precise timing within `update`, but not really _between_ updates since there's all the input/i2c/usb stuff going on there...

(I guess for more advanced stuff you'd want the game to be able to refill the output buffer directly, instead of using a single WAVE channel and a double-copy.)

zenodante:
> @zenodante probably - yes - although we'd need - I guess - to size the wave buffer appropriately (441 samples?) and the _output_ buffer appropriately.
> 
> IIRC an earlier build I had a buffer set up with a memory -> peripheral DMA, and would run the audio engine on the DMA-half-complete callback.
> 
> My reasoning for switching to a 22050Hz native sample rate with the audio engine invoked on every sample (IIRC) is due to wanting to mirror the capabilities of retro audio engines in the hands of expert composers. IE: you can mash the audio channel settings at high frequencies to create new effects and sounds that the sound system designer may not have anticipated.
> 
> That said I'm not sure this is how - for example - the more advanced C64 SID tunes actually worked. Is a 50Hz "Vblank" interrupt equivalent actually... fine?
> 
> This then begs the question- how does the end-user control the audio engine. If we're dealing with a sample buffer and a 50Hz update rate then a majority of changes to the engine in `update` would be moot. You'd almost have to treat audio like a `render` operation and do it there... maybe that's okay?

If we use the timer event to provide the DMA tick and use double buffer for DMA transfer to DAC, we could set the buffer size to any number you would like. 100 means (22050/100 = 220.5 hz buffer update rate) and we will call the sound generation function 100 times to fill the next buffer in interrupt. As the buffer refreashing is done in the DMA finished interrupt, user would not notice any difference. (Just like Daft Freak mentioned, user could only update the channel statement in less than ~100Hz)

It would reduce the interrupt frequency, increase system performance and make the code more I-cache friendly. (Otherwise the sound signal function would be moved into I-cache 22k times per second and replaced by game code after return back from interrupt)

Daft-Freak:
Hmm, you can use the callback on a WAVE channel as as fast-ish timer to mess with audio stuff...

Anyway, using a buffer or not for the output wouldn't really affect this PR as it's just changing the user-visible part. (and SDL already works like that...)

Gadgetoid:
True, getting hung up on a separate issue here- discussion can continue here -> #568

PR #563, opened 2021-01-21
Added Waveform Demo using 32Blit Audio API

ThePythonator:
Created basic demo which generates 8 different sound effects when the corresponding button is pressed (d-pad and ABXY).

If the comments are insufficient please let me know.

Two of the sound effects sound rather similar but I couldn't come up with another one, suggestions are appreciated.


Daft-Freak:
Secret other example list: https://github.com/pimoroni/32blit-beta/blob/master/site/examples/index.html#L46

ThePythonator:
Thanks @Daft-Freak , didn't spot that 😀

ThePythonator:
If there's anything else, please let me know, although I'll also have a quick search through...

Gadgetoid:
I just spent far too long playing with this. Love it.

:laughing:

PR #564, opened 2021-01-22
Added center() function to blit::Rect

ahnlak:
Minor enhancement, after I found myself writing code to get the middle of a Rect for the 42nd time.


Gadgetoid:
I dig this! Thank you.

PR #565, opened 2021-01-23
Fix flash cache fallback

Daft-Freak:
Apparently I didn't test the fallback/error handling... and my emulator hits both of these cases.


Gadgetoid:
:grin: 

Thank you!

PR #566, opened 2021-01-24
Move flash_end to the right place

Daft-Freak:
Doesn't fix anything, but does prevent metadata overwriting ITCM if you attempt to use it. (Making that actually work for games is more work...)


Gadgetoid:
Now I'm curious as to what you've been working on to run into this :laughing: 

Thank you!

PR #571, opened 2021-01-27
Support flash progress at 160x120 res

Gadgetoid:
Uses `screen.bounds` to resize the flash progress bar for 160x120 resolution.

I don't know how this will interact with palette-mode display buffers, but I'm guessing... poorly? Will test.


Gadgetoid:
Looks like Palette mode just yields... SURPRISE COLOURS. I'm... okay with that... for now. Short of somehow detecting a couple of unused colours and swapping them out on the fly, there's naught much we can do. Though maybe this sort of business is a job for a very small, low bit-depth LTDC layer.

PR #572, opened 2021-01-29
Optimise SD writes too

Daft-Freak:
Seems to reduce the setup time of "sd-test" by ~10ms on my "slow" card. (Doesn't seem to affect my "fast" card, it must be slow at writing).

(PR 0x100! :smile:)


Gadgetoid:
Thank you! Did get around to testing this a few days ago but forgot to merge. All a-go-go with house sprucing-up.

PR #573, opened 2021-01-29
CDC stream adjustment/cleanup

Daft-Freak:
A result of trying to figure out why https://github.com/ali1234/32blit-webserial-test didn't work.

One problem was that it would end up stuck in the reloc parsing with 2 bytes in the buffer when it needed 4. This makes it so that you can return from StreamData with a few bytes in the buffer and they will be preserved on the next call.

Also deletes a bunch of stream stuff that we're not using.

### DIFF
--- a/32blit-stm32/Inc/CDCDataStream.h
+++ b/32blit-stm32/Inc/CDCDataStream.h
@@ -13,39 +13,6 @@
 #include <streambuf>
 #include <istream>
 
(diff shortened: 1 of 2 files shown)


ali1234:
This works for me - I can flash from Chrome now. :)

I have removed a lot of the debugging from the above repo so now it sends the whole flash data in one `write` and it still works.

Gadgetoid:
Good enough for me, thanks both!

PR #576, opened 2021-02-01
Launch command

Daft-Freak:
Adds a `LNCH` command which allows ~the device to serve lunch~ launching flies in the same way that the launcher does. Also sends back an OK at the end of flash/save.

This is working towards making auto-launch optional and work for files too. (Need to do some tool bits...)

### DIFF
--- a/firmware/firmware.cpp
+++ b/firmware/firmware.cpp
@@ -24,6 +24,7 @@ extern CDCCommandStream g_commandStream;
 
 FlashLoader flashLoader;
 CDCEraseHandler cdc_erase_handler;
+CDCLaunchHandler cdc_launch_handler;
 
 struct GameInfo {
   char title[25], author[17];
@@ -632,6 +633,8 @@ void init() {
 
   g_commandStream.AddCommandHandler(CDCCommandHandler::CDCFourCCMake<'E', 'R', 'S', 'E'>::value, &cdc_erase_handler);
 
+  g_commandStream.AddCommandHandler(CDCCommandHandler::CDCFourCCMake<'L', 'N', 'C', 'H'>::value, &cdc_launch_handler);
+
   // check for updates
   if(::file_exists("firmware-update.blit")) {
     // TODO: -vx.x.x?
@@ -755,6 +758,26 @@ CDCCommandHandler::StreamResult CDCEraseHandler::StreamData(CDCDataStream &dataS
   erase_flash_game(offset);
 
   return srFinish;
+}
+
+CDCCommandHandler::StreamResult CDCLaunchHandler::StreamData(CDCDataStream &dataStream) {
+  uint8_t byte;
+  while(dataStream.Get(byte)) {
+    if(path_off == MAX_FILENAME) {
+      path_off = 0;
+      return srError;
+    }
+
+    path[path_off++] = byte;
+
+    if(byte == 0) {
+      path_off = 0;
+      launch_file_from_sd(path);
+      return srFinish;
+    }
+  }
+
+  return srContinue;
 }
 
 //////////////////////////////////////////////////////////////////////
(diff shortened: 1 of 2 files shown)


Gadgetoid:
Patched a LNCH command into the tools to test this, seems to work fine! Worth adding the ability to launch from a block/offset, for completeness? Though I don't know how the tool would even handle that.

```diff
diff --git a/src/ttblit/core/blitserial.py b/src/ttblit/core/blitserial.py
index cb4ff2c..6e4c694 100644
--- a/src/ttblit/core/blitserial.py
+++ b/src/ttblit/core/blitserial.py
@@ -120,6 +120,12 @@ class BlitSerial(serial.Serial):
                 sent_byte_count += chunk_size
                 progress.update(chunk_size)
 
+    @firmware_command
+    def launch(self, file_path):
+        command = f'32BLLNCH{file_path}\x00'
+        self.reset_output_buffer()
+        self.write(command.encode('ascii'))
+
     @firmware_command
     def erase(self, offset):
         self.write(b'32BLERSE\x00')
diff --git a/src/ttblit/tool/flasher.py b/src/ttblit/tool/flasher.py
index 444f1cf..1e46e5e 100644
--- a/src/ttblit/tool/flasher.py
+++ b/src/ttblit/tool/flasher.py
@@ -21,7 +21,7 @@ def serial_command(fn):
     @click.option('ports', '--port', multiple=True, help='Serial port')
     @functools.wraps(fn)
     def _decorated(ports, **kwargs):
-        if not ports:
+        if not ports or ports[0].lower() == 'auto':
             ports = BlitSerial.find_comport()
 
         for port in ports:
@@ -31,6 +31,26 @@ def serial_command(fn):
     return _decorated
 
 
+@flash_cli.command()
+@serial_command
+@click.argument('file_path', type=pathlib.Path, required=True)
+def launch(blitserial, file_path):
+    blitserial.launch(file_path)
+
+
+@flash_cli.command()
+@serial_command
+@click.argument('file', type=pathlib.Path, required=True)
+@click.argument('directory', type=str, default='sd:/')
+def install(blitserial, file, directory):
+    blitserial.reset_to_firmware()
+    if directory.startswith('flash:'):
+        blitserial.send_file(file, 'flash')
+    else:
+        directory = directory.replace('sd:', '')
+        blitserial.send_file(file, 'sd', directory=directory)
+
+
 @flash_cli.command()
 @serial_command
 @click.option('--file', type=pathlib.Path, required=True, help='File to save')
 ```

**Comment by Daft-Freak** [comment]
It uses the same function as the `launch` API, so `flash:/[block]` should work.

I was thinking of something like this at the end of sending a file:
```py
        response = serial.read(8)

        if response == b'32BL__OK':
            # get block for flash
            if dest == 'flash':
                block, = struct.unpack('<H', serial.read(2))
                print(block)

            # do launch if requested
```

to replace the current always auto-launching on flash. (`--launch` option to flash/save maybe?)



(Hmm, that line about "auto" in your diff makes me think that `--port all` is gone in the refactor, though it's been a while since I last used that...)

Gadgetoid:
Yes I think `--port all` has gone. BlitSerial `find_comport` - called by default when no port is specified - seems to return *all* matching devices though:

```python
    @classmethod
    def find_comport(cls):
        ret = []
        for comport in serial.tools.list_ports.comports():
            if comport.vid == 0x0483 and comport.pid == 0x5740:
                logging.info(f'Found 32Blit on {comport.device}')
                ret.append(comport.device)

        if ret:
            return ret

        raise RuntimeError('Unable to find 32Blit')
    ```

Which would suggest `all`  is implicit since anything wrapped in the `serial_command` decorator runs for every port in the list.

Daft-Freak:
I don't imagine there are going to be that many people having multiple blits plugged in anyway... and we're getting off topic again :smile:

Gadgetoid:
Forgot to merge this :eyes: 

Thank you!

PR #577, opened 2021-02-01
Sort-of sleep mode

Daft-Freak:
Just fades out audio/the backlight after a period of inactivity. This was an experiment since there seems to have been a few mentions of sleep/low-power/screen-off recently...

Should maybe do something with the accelerometer?


Gadgetoid:
That's really slick! A gentle fade-out too. Love it! Thank you.

Tackling a real system sleep is on my TODO for sometime in the future, the sleep states are scary!

wdevore:
It would be nice if I could turn it off while I have it plugged. I don't code on it everyday and sometimes for weeks. As it is now I need to plug it in every so often just to charge it.

PR #579, opened 2021-02-03
Use find_package() instead of include()

ali1234:
This makes it so that users of 32blit.cmake (now 32blit-config.cmake) use:

    find_package(32BLIT PATHS path/to/sdk)

Instead of:

    include(path/to/sdk/32blit.cmake)

This removes the need to test whether the include actually exists, and is all around more in keeping with how CMake is supposed to work.

The path can now be set on command line using `-D32BLIT_DIR=whatever` (instead of `32BLIT_PATH`). CMake checks this variable automatically.

Note that when doing an emscripten build you must specify an absolute path for `32BLIT_DIR`. When doing SDL or STM32 builds, a relative path is fine.

Backwards compatibility is maintained with a stub `32blit.cmake` that prints a warning and includes `32blit-config.cmake`.

This will allow https://github.com/32blit/32blit-boilerplate/issues/12 to be easily fixed.


ali1234:
PS I didn't update any docs.

Daft-Freak:
There's a windows SDL thing [here](https://github.com/pimoroni/32blit-beta/blob/master/32blit-sdl/CMakeLists.txt#L33), that needs adjusting. Also the "I'm too lazy to type a path twice" thing [here](https://github.com/pimoroni/32blit-beta/blob/master/32blit.toolchain#L101).

Can probably just find/replace the docs, don't think there's much about 32BLIT_PATH/out-of-tree projects there...

32blit.cmake
```diff
@@ -1,61 +1,2 @@
...
-
-	if (${CMAKE_SYSTEM_NAME} STREQUAL Generic)
-		set(FLASH_PORT "AUTO" CACHE STRING "Port to use for flash")
-
-		include(${CMAKE_CURRENT_LIST_DIR}/32blit-stm32/executable.cmake)
-	else()
-		add_subdirectory(${CMAKE_CURRENT_LIST_DIR}/32blit-sdl 32blit-sdl)
-	endif()
-endif ()
+message("Please update your CMake project to use find_package() instead of include().")
```
Daft-Freak, line 1: Might want to expand the warning a bit, as it's not just a case of replacing include with find_package.

Daft-Freak:
Think I tried going all the way and making the libraries installable at one point, but handling BlitHalSDL's dependencies was... painful.

ali1234:
Inside the emscripten block we can just use `32BLIT_DIR` instead. This variable will always be set if the package has been found, and we will never reach this point if it hasn't.

In the toolchain I am not so sure of the side effects, but setting `32BLIT_DIR` here should be okay, except for the `CMAKE_CURRENT_LIST_DIR` being relative. But that only affects emscripten anyway, and you won't use the toolchain for that.

Gadgetoid:
Ah I was going to ask why the change to `32BLIT_DIR` from `32BLIT_PATH`, but now I understand that variable is automatically set.

If a project sets `32BLIT_PATH` and includes `32blit.cmake`, will it still build? Just with a deprecation warning?

ali1234:
32BLIT_DIR is an automatic variable like SDL_DIR etc. User can set it before running find_package to tell it where to look, and afterwards it is guaranteed to be set to the right path. So by using find_package in the backwards-compat, it is guaranteed that 32BLIT_DIR will always be set regardless of how the user project works.

The variable is currently only used in one place inside the SDK, inside the Emscripten stuff to help find the SDL libraries. Getting Emscripten to work at all is quite advanced so if you are using it, you probably know what you are doing.

Daft-Freak:
The use of `32BLIT_PATH/DIR` is to help Windows builds find SDL, Emscripten gets it automagically (`-s USE_SDL=2`)... and I've just checked that it still finds SDL on Windows (with and without updating to find_package).

Gadgetoid:
Thanks all, merging this ahead of the next release!

PR #580, opened 2021-02-04
Add piano to emscripten example list

Daft-Freak:
It wasn't there.


Gadgetoid:
My bad, I always miss *something*. Thank you!

PR #581, opened 2021-02-04
Emscripten install improvement and persistent saves

Daft-Freak:
Fixes the the install step for Emscripten to install the actual game wasm/js instead of just the shell. Then adds a minimal custom shell and makes saves persist.

Includes a workaround for an old SDL issue I forgot to fix. (I'll probably merge [the fix](https://github.com/emscripten-ports/SDL2/pull/136) soon, but would still need to add a new tag/update Emscripten/wait for an Emscripten release/...)

The sync in the save code feels a bit like the wrong place, but `close_file` doesn't know if the file was used for writing... or if it was written to somewhere that needs syncing.

### DIFF
--- a/32blit-sdl/CMakeLists.txt
+++ b/32blit-sdl/CMakeLists.txt
@@ -176,6 +176,10 @@ function(blit_executable NAME SOURCES)
 			SUFFIX ".html"
 			LINK_FLAGS "-s ENVIRONMENT=web -s SDL2_IMAGE_FORMATS=['jpg']"
 		)
+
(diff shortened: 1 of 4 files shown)


Gadgetoid:
The save sync is commented and ifdef'd so it looks pretty harmless. The fact that we require only a handful of little tweaks like this to build to Emscripten is pretty flippin' remarkable. I hope it can be encouraged as a way to share/demo or even release games.

Thank you!

Daft-Freak:
Having an SDL backend is useful :)

PR #583, opened 2021-02-05
Firmware-related docs updates

Daft-Freak:
Add some stuff about firmware-update and launchers.

Things that probably need mentioned somewhere:
- WSL python tools needing a manual `--port /dev/ttySx` (matching the right COMx)
- Installing from releases


Gadgetoid:
Thank you!

PR #585, opened 2021-02-08
Fix "sleep" if the stick is the only thing being used

Daft-Freak:
Tried to play 32blox with the stick. Same thing would probably happen if a game was entirely accel based...


Gadgetoid:
Ooof, this would be tricker to do for accel but perhaps the chip - I can't recall - has some integrated threshold-based motion detection we can use to avoid having to save a previous 3D vector and compare the two.

PR #587, opened 2021-02-09
Improves the doom-fire demo

andreban:
- Updates the algorithm to update / render in a single loop.
- Adds wind to doom fire and binds left/right DPAD keys.
- Button X is used to stop / start the fire.


Gadgetoid:
Love it! Thank you.

PR #588, opened 2021-02-09
Install libraries into mac bundle when not using frameworks

Daft-Freak:
This doesn't seem to explode any more and results in boilerplate-based builds including the SDL libs (using brew to install).

`fixup_bundle` could also be used to handle copying all needed dlls for Windows... but that's currently not working for a bunch of other reasons... (Mostly path related *sigh*)


Gadgetoid:
Oof, very handy! Thank you. Ready-to-ship build output would be ace across all build types.

PR #589, opened 2021-02-10
Add a command line option to set the position of the SDL window

ali1234:
This is useful for testing multiplayer games, as you can run two
copies with arbitrary window positions.


Gadgetoid:
Handy! Thank you.

PR #591, opened 2021-02-11
More metadata author consistency

Daft-Freak:
Improves... this:
Screenshot_2021-02-11 32blit Games

Also changes the one example that uses "Pimoroni" instead of "pimoroni".


Gadgetoid:
Haha, awesome!

PR #592, opened 2021-02-12
Actions repo name fix

Daft-Freak:
Avoid using the repo name in paths so that actions builds don't fail if your repo has a different name.

(Guess who hasn't renamed their 32blit-beta repo :smile: )


Gadgetoid:
Ha! Whoops. Yeah this is better. For some reason paths hurt my brain in Actions. Thank you!

PR #593, opened 2021-02-12
Adjust NOISE waveform

Daft-Freak:
Adjust the LFSR tap so that 0x8000 doesn't end up always 0 and fix the overflow check to actually happen on overflow. Without this frequency doesn't have much of an effect and some values result in no sound at all.

(The waveform isn't off-centre now and it... sounds better? :shrug: )


Gadgetoid:
I concur, noise sounds ace now and I better merge this before I lose hours playing with noise piano :laughing:  Thank you!

PR #594, opened 2021-02-13
Some MinGW/SDL related fixes

Daft-Freak:
Found this old branch with some attempts at fixing the MinGW fails in #483. It doesn't entirely fix that and fixes nothing with the current SDL version we're using... but is more correct (I think).


Gadgetoid:
I'm taking it on faith that this is a step forward since I'm not tooled up for windows-anything at the moment. :laughing: Thank you!

PR #595, opened 2021-02-13
Make most menu item lists const

Daft-Freak:
Another patch I found in my giant pile of branches. Saves a little RAM.

(I have a another patch that makes all the USB descriptors const too https://github.com/Daft-Freak/32blit-beta/commit/b5494a36202c954ce55f4c6a3ff6e770d9b09455 which removes most of the firmware's `.data`)


Gadgetoid:
Thank you!

PR #599, opened 2021-02-14
Rework SDL scaling

Daft-Freak:
I realised that you can pretty easily get access to SDL events in a game with something like:
```c++
#ifndef TARGET_32BLIT_HW
#include <SDL.h>

int sdl_event_watch(void *user_data, SDL_Event *event) {
    switch(event->type) {
        //...
    }
    return 1;
}

#endif

void init() {
    //...

#ifndef TARGET_32BLIT_HW
    SDL_AddEventWatch(sdl_event_watch, nullptr);
#endif
}
```

So, I attempted to do something with the mouse... coords were all wrong due to the scaling. Then I realised that _almost_ all the scaling stuff could be replaced with `SDL_RenderSetLogicalSize`/`SDL_RenderSetIntegerScale`, which also scales input. (The `Stretch` mode is a bit nasty, I'm just overriding the scale every resize... but only the `KeepPixels` mode is actually used.)

### DIFF
--- a/32blit-sdl/Input.cpp
+++ b/32blit-sdl/Input.cpp
@@ -65,29 +65,25 @@ int Input::find_button(int button) {
 	else return iter->second;
 }
 
-Input::Input(SDL_Window *window, System *target) : target(target) {
-	int w, h;
-	SDL_GetWindowSize(window, &w, &h);
-	resize(w, h);
-}
-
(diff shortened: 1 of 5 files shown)


Gadgetoid:
Works for me. Thank you!

PR #600, opened 2021-02-15
Handle more files in the launcher without crashing

Daft-Freak:
Allows the launcher to list more associated files (tested to about 700-800), with the side effect that it will now crash on a larger directory even if the files aren't associated.

(Without pre-reserving the vector, the exponential growth causes it to run out of memory part way through the loop. Using https://github.com/Daft-Freak/32blit-beta/commit/6ae6dba6397d508d0586ae058f6756fa2f1b8ae7 to actually detect out of memory.)


Gadgetoid:
Thank you!

Issue #598, opened 2021-02-14
multiplayer_is_connected() returns true and then send_message() hangs if blit is connected to something other than another blit


ali1234:
Same code as #597

Run it on 32blit while connected to Linux PC on USB.

Result: It hard locks after about 12 messages. If you then unplug the USB cable, the system menu pops up but buttons don't respond. You have to press reset to clear the error.

Since there is no API to tell whether the other side of the connection is a blit or something else, the only way to find out is by sending a message and waiting for a response. Since the cable might not be connected at game start, you will want to resend the detection message periodically. That guarantees a crash when connected to PC...

ali1234:
PS I did not check with the fix for 597 applied.

ali1234:
Fixed in #602

Issue #597, opened 2021-02-14
Multiplayer can only receive one message per render()


ali1234:
This code sends one message per update().

```
#include <cstring>

#include "multiplayer.hpp"

#include "engine/multiplayer.hpp"

using namespace blit;

uint32_t sent = 0;
uint32_t received = 0;
uint32_t frames = 0;

void on_message(const uint8_t *data, uint16_t len) {
    received++;
}

void init() {
    message_received = on_message;
    enable_multiplayer();
}

void render(uint32_t time_ms) {
    char msg[100];
    sprintf(msg, "sent: %d\nreceived %d\nframes: %d", sent, received, frames);
    screen.pen = Pen(0, 0, 0);
    screen.clear();
    screen.alpha = 255;
    screen.pen = Pen(255, 255, 255);
    screen.text(msg, minimal_font, Point(0, 30));
    frames++;
}

void update(uint32_t time_ms) {
    auto out = "Hi!";
    send_message((uint8_t *)out, strlen(out));
    sent++;
}

```

Result:

![image](https://user-images.githubusercontent.com/651970/107876517-5131ec80-6ebe-11eb-9dcb-0061a080f1aa.png)

Daft-Freak:
Hmm, only affects SDL due to it only doing updates every 20ms (https://github.com/32blit/32blit-sdk/blob/master/32blit-sdl/System.cpp#L199)... and Multiplayer::update only handling one message before it returns. 

Throwing a loop in there helps. https://github.com/Daft-Freak/32blit-beta/commit/3801a950b9f8720061f8dc9f247d5ffb8dd96b69

ali1234:
Fixed in #602

PR #602, opened 2021-02-15
Multiplayer improvements

Daft-Freak:
This allows SDL to receive messages faster than one every 20ms(#597) and adds a handshake to the STM32 backend so that `is_multiplayer_connected` actually means that two devices are connected and in "multiplayer mode" (#598).


Gadgetoid:
Thank you!

PR #603, opened 2021-02-16
Replace prng_lfsr with prng_xorshift in audio.cpp

sh1boot:
It's a better PRNG and still very simple.


Gadgetoid:
Thank you. Interestingly the low-pass approach has all but eliminated any audible perception of pitch in the noise. Instead, the noise appears to get louder as the cut-off frequency increases. You can see the increase in "loudness" in waveform view:

image

You can very easily see the frequency components change, though. Here's the lowest note in the `piano` example, which I modified to emit noise:

image

And the highest:

image

I think this might be a typical problem with cutoff frequency where the oscillator volume isn't attenuated to compensate, but now we're getting really complicated trying to find a correct approach, when the crude approach gave the right "sound". This is pushing the limits of my musical and signal knowledge, though.

sh1boot:
Oh, I guess that makes sense.

The low-pass filter does nothing but selectively(ish) attenuate the signal, whereas sample-and-hold keeps exactly the same amplitude but changes the frequency response with time distortion (and in-fills the high part of the spectrum with reflections rather than distinct noise).

When I [plotted both](https://colab.research.google.com/drive/1KSJh8XqGlM7SDNXBco1854o330tlF4yw) I saw that the sample-and-hold solution actually grew in amplitude as the cut-off frequency fell, which I couldn't understand so I added a fiddle to make it easier to compare and then never tried to figure out what was happening.

The proper scale factor seems to be `sqrt(f_c/f_s)` and I'd propose to use this in the low-pass filter implementation, but I'd worry about overflow.


In my experience noise comes out _way_ louder than, eg., triangles at the same peak amplitude, so should the noise waveform be scaled down by a dozen dB or so before being used so that its perceived loudness is in line with the other waveform functions?  That'd probably fix the overflow concern.

As for why bother; it's mostly academic, but technically the sample-and-hold solution is made up of noise up to the specified frequency, and then forward and backward reflections of those frequencies from that point on.  I don't know if that's perceptible at all.  I would guess not.  The other problem is that its pass-band is actually the green function here:

image

That is, it has a series of notch filters at all the multiples of the sample frequency.  Again, I can't hear it, but ... _technically_.

sh1boot:
Tossed out the filter part because complexity (and also not a feature of the pull request).  Then because I can't help but fiddle I tweaked the generator to approximate a normal distribution because that's more correcterer.

Trying to understand why I believed this to be the case, I learned that whatever distribution you use, if you suppose that you generated it at a much higher frequency and then filtered it down to your sampling frequency then as that scale factor increases the distribution function for individual samples tends towards a normal distribution.  So the idea of calculating your samples on a normal distribution rather than on a uniform distribution is a short-cut to pretending that you've generated white noise out to infinity and then averaged that down to your sampling frequency.  Then that concept is reversed again because the easiest way to make a very coarse approximation of a normal distribution is to average several uniform random values.

_Or_, another rationale is that for equal power, more of your samples stay closer to 0 -- so clipping occurs less frequently and filtering comes out a bit tidier.

Gadgetoid:
Wow. It's... beautiful...
image

me: sheesh that's gotta take a load of code....
*looks at code* ... wat?

PR #604, opened 2021-02-16
Improve fade in from "sleep"

Daft-Freak:
This fixes two (small) issues with the fade from "sleeping"
 - If you held a button while fading in, the screen would stay black
 - If you tried to cancel during the fade out, it would fade in from black

It would also increase the timeout... if I hadn't accidentally done that in the previous PR :facepalm:


Gadgetoid:
Ha good timing, I only just noticed that bug today! Thank you.

PR #605, opened 2021-02-18
No auto-launch on CDC flash

Daft-Freak:
Removes the auto-launch after flashing something through CDC. However, if the flashed file is a launcher we still have to run it as we probably just deleted the running one... (Similar thing for firmware-updates)


Daft-Freak:
This PR and https://github.com/32blit/32blit-tools/pull/72 depend on each other, at least partially.

Gadgetoid:
Works a treat! Thank you.

PR #609, opened 2021-02-19
Basic test/demo of the metadata API

Gadgetoid:
Grabs the metadata for the running game and displays it, whee!

I wonder if it's worth being able to get a pointer to the image data, too? Would a game ever want to display its own splash image or icon?


Daft-Freak:
> I wonder if it's worth being able to get a pointer to the image data, too? 

That would require the images to be embedded in SDL builds first...

PR #610, opened 2021-02-19
Fix VS releases actually being debug builds

Daft-Freak:
https://github.com/32blit/32blit-boilerplate/pull/17, but for the SDK.


Gadgetoid:
Thank you- got some projects to patch!

PR #611, opened 2021-02-21
Document Windows builds without needing WSL or a full Visual Studio install

Daft-Freak:
This is the thing I keep mentioning whenever a Windows user has setup problems... Finally got around to working out what was required. This was tested on a "clean" Windows VM.

The basic Windows instructions are now very similar to the Linux/macOS setup. I was tempted to replace every occurrence of `make x` in the docs with `cmake --build . --target x` since that works with every CMake generator. (Instead of "follow these docs, but use nmake instead of make")

I guess the "you need a cross-compile env" bits in 32blit.md and 32blit-firmware.md could link to this too.

(Apparently the default windows Python install has `py` in PATH and not `python` or `python3`...)

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,6 +22,7 @@ Refer to the OS specific docs:
 * [Linux](docs/Linux.md)
 * [Linux - ChromeOS](docs/ChromeOS.md)
 * [Windows - Visual Studio](docs/Windows-VisualStudio.md)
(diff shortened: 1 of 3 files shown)


Gadgetoid:
Thank you- at some point I'm going to have to commission a fresh windows install and try this. Also need a mac... :laughing:

PR #612, opened 2021-02-22
Split api_version to handle breaking changes

Daft-Freak:
Should be avoided, but handling it instead of crashing is probably still a good idea. Also, it's not like we were going to need 4 billion API versions...

version_major = breaking changes
version_minor = non-breaking API additions.

### DIFF
--- a/32blit-stm32/startup_user.cpp
+++ b/32blit-stm32/startup_user.cpp
@@ -6,17 +6,20 @@ extern void update(uint32_t time);
 extern void render(uint32_t time);
 
 extern "C" bool cpp_do_init() {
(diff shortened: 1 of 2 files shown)


Gadgetoid:
I dunno the pace of development is pretty fast :laughing:

Good shout, though, will try and get this in for the pending release.

Though right now testing is fun since we're not even versioning the API.

Daft-Freak:
Next: making the launcher/firmware do something when a game doesn't launch due to incompatibility...

Gadgetoid:
That *would* be sensible :laughing:

PR #614, opened 2021-02-23
Fix MinGW SDL DLLs

Daft-Freak:
If SDL2_DIR didn't have a trailing slash some of the find_*s would fail, resulting in MinGW not copying some of the dlls.

The MinGW toolchain defaults SDL2_DIR to something without a trailing slash...


Gadgetoid:
Thank you!

PR #615, opened 2021-02-24
Actually init API versions/unbreak everything

Daft-Freak:
Someone forgot that the API section isn't initialised... oops.


Gadgetoid:
Welp I totally missed it, too! :facepalm: 

Thank you!

PR #616, opened 2021-02-24
Launch/API fixes

Daft-Freak:
- Fixed an incompatible launcher resulting in a reset loop (now handled the same as no launcher)
- Added a message when a game fails to launch
- Fixed flashing a game when the same game was running (it would get erased and crash when resuming it)

The special case of "flashing a launcher always resets" is now "flashing the running game always resets".

Tested with an incompatible launcher/game and flashing the same game twice.

Need to fix the tools to not error out when the device resets after flashing...

### DIFF
--- a/32blit-stm32/Inc/32blit.h
+++ b/32blit-stm32/Inc/32blit.h
@@ -33,7 +33,7 @@ extern void blit_process_input();
 
 // Switching execution.
 // Address is relative to the start of flash, ignored if switching to firmware
-extern void blit_switch_execution(uint32_t address, bool force_game);
+extern bool blit_switch_execution(uint32_t address, bool force_game);
 extern bool blit_user_code_running();
 extern "C" void blit_reset_with_error();
 extern void blit_enable_user_code();
--- a/32blit-stm32/Src/32blit.cpp
+++ b/32blit-stm32/Src/32blit.cpp
@@ -669,7 +669,7 @@ void blit_process_input() {
 typedef  void (*pFunction)(void);
 pFunction JumpToApplication;
 
-void blit_switch_execution(uint32_t address, bool force_game)
+bool blit_switch_execution(uint32_t address, bool force_game)
 {
   if(blit_user_code_running() && !force_game)
     persist.reset_target = prtFirmware;
@@ -693,6 +693,7 @@ void blit_switch_execution(uint32_t address, bool force_game)
     // TODO: may be possible to return to the menu without a hard reset but currently flashing doesn't work
     SCB_CleanDCache();
     NVIC_SystemReset();
+    return true; // can't get here
   }
 
 	// switch to user app in external flash
@@ -714,20 +715,21 @@ void blit_switch_execution(uint32_t address, bool force_game)
       if(!init(address)) {
         user_render = nullptr;
         user_tick = nullptr;
-        // this would just be a return, but qspi is already mapped by this point
+        // don't try to auto-launch this game again
         persist.reset_target = prtFirmware;
-        SCB_CleanDCache();
-        NVIC_SystemReset();
-        return;
+
+        qspi_disable_memorymapped_mode();
+
+        return false;
       }
 
       blit::render = user_render;
       do_tick = user_tick;
-      return;
+      return true;
     }
     // anything flashed at a non-zero offset should have a valid header
     else if(address != 0)
-      return;
+      return false;
   }
 
   // old-style soft-reset to app with linked HAL
@@ -778,6 +780,8 @@ void blit_switch_execution(uint32_t address, bool force_game)
 	while(1)
 	{
 	}
+
+  return true;
 }
 
 bool blit_user_code_running() {
--- a/firmware/firmware.cpp
+++ b/firmware/firmware.cpp
@@ -454,8 +454,7 @@ static bool launch_game_from_sd(const char *path, bool auto_delete = false) {
     if(auto_delete)
       ::remove_file(path);
 
-    blit_switch_execution(launch_offset, true);
-    return true;
+    return blit_switch_execution(launch_offset, true);
   }
 
   blit_enable_user_code();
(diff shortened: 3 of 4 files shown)


Gadgetoid:
Jeepers you're quick! I was only just starting to notice the flashing-the-running-game crash as I was working on the audio demo last night. Will give it a try, thank you.

Gadgetoid:
Any reason you don't move the handling of flashed data until after the response is sent?

```patch
diff --git a/firmware/firmware.cpp b/firmware/firmware.cpp
index e1015f6..fb4904b 100644
--- a/firmware/firmware.cpp
+++ b/firmware/firmware.cpp
@@ -988,8 +988,6 @@ CDCCommandHandler::StreamResult FlashLoader::StreamData(CDCDataStream &dataStrea
           progress.update(m_uParseIndex + 1);
 
           if(bEOS) {
-            handle_data_end(result != srError);
-
             if(result != srError) {
               while(CDC_Transmit_HS((uint8_t *)"32BL__OK", 8) == USBD_BUSY){}
               if(dest == Destination::Flash) {
@@ -1001,6 +999,7 @@ CDCCommandHandler::StreamResult FlashLoader::StreamData(CDCDataStream &dataStrea
               result = srFinish;
             }
             progress.hide();
+            handle_data_end(result != srError);
           }
 
           m_uParseIndex++;
```

Gadgetoid:
Another approach might be something like this, but - since the "Erasing Flash Blocks" step on the 32blit currently freezes up the tool with no feedback about what's happening and this ruins your clean separation of concerns - maybe not.

```patch
diff --git a/firmware/firmware.cpp b/firmware/firmware.cpp
index e1015f6..d7b5b8a 100644
--- a/firmware/firmware.cpp
+++ b/firmware/firmware.cpp
@@ -989,7 +989,6 @@ CDCCommandHandler::StreamResult FlashLoader::StreamData(CDCDataStream &dataStrea
 
           if(bEOS) {
             handle_data_end(result != srError);
-
             if(result != srError) {
               while(CDC_Transmit_HS((uint8_t *)"32BL__OK", 8) == USBD_BUSY){}
               if(dest == Destination::Flash) {
@@ -1001,6 +1000,7 @@ CDCCommandHandler::StreamResult FlashLoader::StreamData(CDCDataStream &dataStrea
               result = srFinish;
             }
             progress.hide();
+            if(result == srFinish) blit_enable_user_code();
           }
 
           m_uParseIndex++;
@@ -1056,7 +1056,6 @@ void FlashLoader::handle_data_end(bool success) {
   // done recieving data
   if(dest == Destination::SD) {
     f_close(&file);
-    blit_enable_user_code();
   } else {
     // flash
     if(success) {
@@ -1083,7 +1082,6 @@ void FlashLoader::handle_data_end(bool success) {
         qspi_enable_memorymapped_mode();
         flash_mapped = false;
       }
-      blit_enable_user_code();
     }
   }
 }
 ```

**Comment by Daft-Freak** [comment]
Only noticed it myself when I went to test "super blit kart"...

First idea looks okay, it has finished flashing/saving at that point though it's relying on the erase taking time to allow the response to get there. (I think?)

Second one just moves switching to the (old) game to the other side of sending the response. Don't think the game code gets a chance to run in there anyway...

**Comment by Gadgetoid** [comment]
I *think* - at least on my Linux setup - that as long as the response is output to serial, then it will be read at the receiving end (since it's in the receive buffer) even though the port has *technically* closed. Re-flashing `doom-fire` over and over again is surely not the most scientific test but I was unable to provoke a failure- even If I delete `doom-fire` first, so that there's no erase step at all.

**Comment by Daft-Freak** [comment]
The only problem I can find is if you try to flash multiple "large" games too quickly it gets a bit confused. (Sometimes crashes, sometimes fails with random parse errors).

Bit of a corner case, especially considering that used to be a lot more broken...

**Comment by Gadgetoid** [comment]
Does make me wonder if we need an up-front handshake before sending the rest of the game data, but that's now getting into rewrite-all-the-things territory to save a corner case. Something a big ol' `time.sleep()` could fix... albeit not exactly gracefully.

Worth noting that our internal commissioning tools basically do exactly this, though, with lots of fault tolerance:

```sh
echo "Ensure 32Blit is out of DFU mode"
while [ 1 ]; do
    serial_port=`readlink -f /dev/serial/by-path/platform-fd500000.pcie-pci-0000:01:00.0-usb-0:$1:1.0`
    32blit flash --port $serial_port info
    if [ $? -eq 0 ]; then
        break
    fi
    sleep 0.2
done
for f in bin/*.blit; do
    echo "Flashing $f"
    sleep 0.9
    status=1
    while [ $status -eq 1 ]; do
        serial_port=`readlink -f /dev/serial/by-path/platform-fd500000.pcie-pci-0000:01:00.0-usb-0:$1:1.0`
        32blit flash --port $serial_port flash --file $f
        status=$?
        sleep 0.2
    done
    serial_port=`readlink -f /dev/serial/by-path/platform-fd500000.pcie-pci-0000:01:00.0-usb-0:$1:1.0`
    32blit flash --port $serial_port reset
done
```

Though this is a problem that could be solved much more elegantly by tricking the blit into flashing a complete binary of games.

Gadgetoid:
Flashing a game while the launcher is running works successfully, but since the launcher is no longer quit and re-launched it doesn't update the list of installed games.

An *extremely* dirty solution would be to call `blit::init();` again, but since this is written only to be run *once* bad times will probably ensue.

I dare say we need a `blit::register_signal_handler(function)` or maybe just special case handling to force a launcher reset if a game is flashed.

I supose suspend/resume and other "signals" could also be handled via a word or two in the persistent state- set the signal, call `update` once. Feels hacky, too. Not sure quite where to go with this.

Daft-Freak:
Hmm, yeah I had noticed that too... Calling init again wouldn't even be that bad as the "init" exposed to the firmware includes global init so it's _nearly_ a reset. (+ `init_api_shared` covers almost everything else).

Avoiding resets on game -> game switches is something I want to try at some point...

Gadgetoid:
So far so good anyway, thank you!

PR #618, opened 2021-02-25
A few small fixes

Daft-Freak:
First one fixes `vec.transform(Mat4::identity())` having an effect. Second one makes it so you can do:
```c++
static constexpr blit::Size size{96, 96};
uint8_t data[size.area()];

blit::Surface surface{data, blit::PixelFormat::P, size};
```
(From my minimap code)

There are probably a lot more methods that could be `const`/`constexpr`.

### DIFF
--- a/32blit/graphics/surface.cpp
+++ b/32blit/graphics/surface.cpp
@@ -299,10 +299,10 @@ namespace blit {
     if (dr.empty())
       return; // after clipping there is nothing to draw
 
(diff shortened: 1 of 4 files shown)


Daft-Freak:
Should've called this "kart fixes"... added a few more including a fix for massive scaled sprites (probably also an optimisation...)

Gadgetoid:
Thank you! Interesting point on the efficiency of using int vs byte.

PR #623, opened 2021-02-28
Handle metadata with quotes in 32blit-sdl

Daft-Freak:
I just exploded my build a bit...

Another case of needing to do a bit more work because the magic command that fixes it was only added "recently".

(This only works if `metadata.cmake` has its quotes escaped... which it doesn't.)


Daft-Freak:
Even without fixing the tools, this turns the compile failure into a CMake warning and an incorrect string...

Gadgetoid:
Thank you!

PR #624, opened 2021-03-01
More multiplayer improvements

Daft-Freak:
Fixes for SDL connection handling and easier bridging to the STM32 side.

Fun test I came up with for this (assuming no --connect/--listen):

- Start game 1. Starts listening.
- Start game 2. Connects to 1, 1 stops listening.
- Start game 3. Starts listening.
- Start game 4. Connects to 3, 3 stops listening.
- You now have two independent pairs of games open.
- Close 3. 4 disconnects.
- Close 2. 1 starts listening, 4 connects to 1, 1 stops listening again.
- Still have a connection, but neither of the ones we started with!

### DIFF
--- a/32blit-sdl/Multiplayer.cpp
+++ b/32blit-sdl/Multiplayer.cpp
@@ -26,10 +26,8 @@ void Multiplayer::update() {
     if(!socket && !listen_socket) {
         // attempt to reconnect
         auto now = SDL_GetTicks();
-        if(have_ip && (now - last_connect_time) > retry_interval) {
-            socket = SDLNet_TCP_Open(&ip);
-            if(socket)
-              SDLNet_TCP_AddSocket(sock_set, socket);
+        if((now - last_connect_time) > retry_interval) {
+            setup();
 
             last_connect_time = now;
         }
@@ -145,7 +143,7 @@ void Multiplayer::send_message(const uint8_t *data, uint16_t length) {
 void Multiplayer::setup() {
     const uint16_t port = 0x32B1;
 
-    have_ip = false;
+    IPaddress ip;
 
     // try connecting first for auto
     if(mode != Mode::Listen) {
@@ -167,18 +165,18 @@ void Multiplayer::setup() {
         listen_socket = SDLNet_TCP_Open(&ip);
     }
 
-    // we at least have a valid address now
-    have_ip = true;
-
     if(!socket && !listen_socket) {
         std::cerr << "Failed to open socket: " << SDLNet_GetError() << std::endl;
         return;
     }
 
-    if(listen_socket)
+    if(listen_socket) {
         SDLNet_TCP_AddSocket(sock_set, listen_socket);
-    else
+        mode = Mode::Listen;
+    } else {
         SDLNet_TCP_AddSocket(sock_set, socket);
+        mode = Mode::Connect;
+    }
 }
 
 void Multiplayer::disconnect() {
--- a/32blit-sdl/Multiplayer.cpp
+++ b/32blit-sdl/Multiplayer.cpp
@@ -78,12 +78,19 @@ void Multiplayer::update() {
             continue;
 
           if(memcmp(head_buf, "32BLUSER", 8) == 0) {
-            read = SDLNet_TCP_Recv(socket, head_buf, 2);
+            // get the length
+            int off = 0;
 
-            if(read <= 0) {
-              disconnect();
-              return;
-            }
+            do {
+              read = SDLNet_TCP_Recv(socket, head_buf + off, 2 - off);
+
+              if(read <= 0) {
+                disconnect();
+                return;
+              }
+
+              off += read;
+            } while(off != 2);
 
             recv_len = head_buf[0] | (head_buf[1] << 8);
             recv_buf = new uint8_t[recv_len];
--- a/32blit-sdl/Multiplayer.cpp
+++ b/32blit-sdl/Multiplayer.cpp
@@ -52,6 +52,8 @@ void Multiplayer::update() {
 
               SDLNet_TCP_AddSocket(sock_set, socket);
 
+              SDLNet_TCP_Send(socket, "32BLMLTI\1", 9);
+
               // stop listening now
               stop_listening();
           }
@@ -81,6 +83,14 @@ void Multiplayer::update() {
             recv_len = head[0] | (head[1] << 8);
             recv_buf = new uint8_t[recv_len];
             recv_off = 0;
+          } else if(memcmp(head, "32BLMLTI", 8) == 0) {
+            // handle the handshake packet
+            SDLNet_TCP_Recv(socket, head, 1);
+            handshake = head[0] == 1;
+
+            if(mode == Mode::Connect)
+              SDLNet_TCP_Send(socket, "32BLMLTI\1", 9);
+
           } else {
             std::cerr << "Unexpected header: " << std::string(reinterpret_cast<char *>(head), 8) << std::endl;
           }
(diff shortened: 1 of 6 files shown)


ali1234:
This still doesn't quite work for SDL to USB. The SDL client connects but does not send `32BLMLTI`, instead it goes directly to sending `32BLUSER` messages:

```
client connect TCP:6
pairing clients TCP:6, USB:4
comms USB:4, TCP:6: starting
comms TCP:6, USB:4: starting
comms TCP:6, USB:4:  b'32BLUSER\x1c\x004\x124\x12\x00\x00\x00\x00\x00\x00\x00\x00\xb6U\x00\x000\xac2\x05\xd2\x7f\x00\x00.\xa2\x17\x8c'
comms TCP:6, USB:4:  b'32BLUSER\x1c\x004\x124\x12\x00\x00\x00\x00\x00\x00\x00\x00\xb6U\x00\x000\xac2\x05\xd2\x7f\x00\x00.\xa2\x17\x8c'
comms TCP:6, USB:4:  b'32BLUSER\x1c\x004\x124\x12\x00\x00\x00\x00\x00\x00\x00\x00\xb6U\x00\x000\xac2\x05\xd2\x7f\x00\x00.\xa2\x17\x8c'
```

ali1234:
This seems to be because only the SDL listening client will send 32BLMLTI (and perhaps only the USB client in host mode?) - so if we connect a connecting SDL client and a USB device client, it seems like both will just sit and wait for the other to send the first handshake.

Daft-Freak:
Hmm, yeah no hosts, no handshakes... For the STM32 code only sending handshakes from the (USB) host avoids the hang you get when nothing is reading the other end.

Sending a MLTI packet to one of them should be enough to get it going. (With the newest commit to stop an infinite loop of replies...)

ali1234:
Injecting 32BLMLTI works for USB devices, but still does not work for SDL.

ali1234:
I have pushed my latest bridge code here https://github.com/ali1234/32blit-mpbridge

anyio_serial bug has been fixed so it is safe to install now.

ali1234:
I determined that both sides see a connection. But for some reason the SDL client never receives any messages, even though I can see USB sending them. I think it is still somehow desyncing...

ali1234:
Okay, the reason for this is that the SDL code expects to always read a whole 8 bytes of header in one go. This does not happen when bridging to serial. The serial read unblocks as soon as it has one character, so then the SDL client gets the first character of the header in a read, followed by the rest of the message in a second read.

ali1234:
I did a hacky fix to force the SDL code to read 8 bytes in a loop and now it works: https://youtu.be/Popj8ZSsi4c

It crashed shortly after this, at the end of the match (ie after two rounds). I am not sure why yet.

Daft-Freak:
Added header buffering, but the '2' from the handshake is going missing...

```
pairing clients TCP:3, USB:0
comms USB:0, TCP:3: starting
comms TCP:3, USB:0: starting
comms _____ -> USB:0:  b'32BLMLTI\x01'
comms USB:0 -> TCP:3:  b'3'
comms USB:0 -> TCP:3:  b'BLMLTI\x02'
comms USB:0 -> TCP:3:  b'3'
comms USB:0 -> TCP:3:  b'2BLUSER\x12\x00This is a message!'
comms USB:0 -> TCP:3:  b'3'
comms USB:0 -> TCP:3:  b'2BLUSER\x12\x00This is a message!'
```
(Got the messages though)

ali1234:
Hmm I've seen it eat the 3 but never the 2... it might be that bytes get lost if you don't start reading from the serial port fast enough after opening it... but losing a byte in the middle is weird.

ali1234:
> (This is all wrong because I made the mistake of believing the SDL_net docs, which say that it waits for there to be `maxlen` available... but it actually just calls `recv`... _sigh_)

Ah yes you made the same mistake as me this morning. You were reading the docs for SDL 1.x. SDL 2.x is different, but for some reason the docs aren't on libsdl.org, or at least Google won't find them.

Daft-Freak:
Hmm, first connection with the bridge seems to always work but then restarting the SDL side results in the missing '2' (or sometimes '3'). Unplugging the device and plugging it back in also works without bytes going missing.

ali1234:
I think the USB code is now leaking buffers. If I send a lot of very large messages it eventually starts writing over the screen, and then crashes shortly after.

Daft-Freak:
Found something in the FIFO code (caused by the fix in #573, oops). I can now spam 16k packets without killing the firmware.

ali1234:
I left USB running all night in loopback mode and again it desynced during initial game sync. This is the last thing that got sent:
```
comms USB:0 -> USB:0:  b'3'
comms USB:0 -> USB:0:  b'2BLUSER\x0c\x004\x124\x12\x00\x00\x00\x00\x07\x00\x00\x00'
comms USB:0 -> USB:0:  b'3'
comms USB:0 -> USB:0:  b'2BLUSER\x0c\x004\x124\x12\x00\x00\x00\x00\x00\x00\x00\x00'
comms USB:0 -> USB:0:  b'32BLUSER\x0c\x004\x124\x12\x00\x00\x00\x00\x01\x00\x00\x0032BLUSER\x1c\x004\x124\x12\x00\x00\x00\x00\x02\x00\x00\x00\x15E\x1e+&N\xd6$\x8a\xb7\xdb>\x07\xb5w.'
```
Each message has 8 bytes of fixed data (0x12341234, 0x00000000) then a type field:
1. Game over (0x00000007)
2. Ping (0x00000000)
3. Pong (0x00000001)
4. Challenge request (0x00000002, followed by 4x uint32_t random numbers)

Here is the final message fully decoded (it appears to be correct in that no bytes were lost or added):

```
33 32 42 4c 55 53 45 52 0c 00 34 12 34 12 00 00 00 00 01 00 00 00
3  2  B  L  U  S  E  R    12   0x12341234  0x00000000  0x00000001 (pong)

33 32 42 4c 55 53 45 52 1c 00 34 12 34 12 00 00 00 00 02 00 00 00 15 45 1e 2b 26 4e d6 24 8a b7 db 3e 07 b5 77 2e
3  2  B  L  U  S  E  R    28   0x12341234  0x00000000  0x00000002  0xRRRRRRRR  0xRRRRRRRR  0xRRRRRRRR  0xRRRRRRRR (challenge)
```

The last block of data sent includes two messages. Is it possible that the USB parser can't handle receiving two messages in one read? This would be unlikely to happen during gameplay because the only packets sent during that phase are over 400 bytes.

ali1234:
Looks like this is indeed the problem. To reproduce:

1. Get updated multiplayer example from https://github.com/32blit/32blit-sdk/pull/627
2. Change line 59 to `bool button_pressed = !(buttons & Button::A);` (ie make it send messages all the time unless you hold A)
3. Run `mpbridge --loop`
4. Run modded multiplayer example.

Result: On TCP client it works fine. sent always == recv. On USB, it loses about 1% of messages, apparently when two arrive at the same time.

ali1234:
Caught it losing a message:

IMG_20210306_121108

Daft-Freak:
Yep, parser was just throwing away anything after the handler returned. Modified the multiplayer example to send two messages each time. Every other one was consistently lost.

ali1234:
Unfortunately this still doesn't quite work. If I send four messages all at once, the first two messages are received, and then nothing until ~160.

image

Daft-Freak:
Last commit added a new bug when the length gets split across buffers, oops. Also fixed some bugs in handling the command getting split.

First bug resulted in the third message being parsed as having a length of 0x1212 instead of 0x12.

ali1234:
This is now stable enough to play several rounds in a row repeatably. There are still some desyncs but I think they are bugs in my code.

Gadgetoid:
Tested with the aid of #627 - thanks both!

Looks like a `--help` is in order for SDL builds. Since we have metadata baked into SDL maybe also a `--version`.

PR #625, opened 2021-03-01
Use .mp4 for video extension

Daft-Freak:
This stops ffmpeg spamming MPEG related errors at me. (Yes, I'm actually using the video recording stuff...)


Gadgetoid:
The extra step required to actually use the videos anywhere has kept me using OBS, but I've at least *tried* to use the record functionality. Always thought it was a useful addition so I'm glad *someone's* using it.

Daft-Freak:
Yeah, me being too lazy to open/configure OBS for a short video (when it's not failing to load with "no supported GPU"). Usually have to re-encode either way for Discord's tiny size limit though.

Hmm, may have a problem if I ever get sound in my games...

PR #626, opened 2021-03-02
Tilemap optimisations

Daft-Freak:
Reduces "Super Blit Kart" `render` times by 8ms and makes the track rendering > 2.5x faster. First commit also means you can put something in the tile index you have defined as empty for other uses.

(I also have some patches for sprite stretch blits, but I think those need more work/testing)

### DIFF
--- a/32blit/graphics/tilemap.cpp
+++ b/32blit/graphics/tilemap.cpp
@@ -207,7 +207,7 @@ namespace blit {
         wc += dwc;
         doff++;
         c--;
-      } while(c && (wc.x >> fix_shift + 3) == wcx >> 3 && (wc.y >> fix_shift + 3) == wcy >> 3);
+      } while(c && (wc.x >> (fix_shift + 3)) == wcx >> 3 && (wc.y >> (fix_shift + 3)) == wcy >> 3);
 
     } while (c);
   }
--- a/32blit/graphics/tilemap.cpp
+++ b/32blit/graphics/tilemap.cpp
@@ -24,8 +24,8 @@ namespace blit {
    * TODO: Document
    */
   int32_t TileMap::offset(const Point &p) {
-    int32_t cx = ((uint16_t)p.x) & (bounds.w - 1);
-    int32_t cy = ((uint16_t)p.y) & (bounds.h - 1);
+    int32_t cx = ((uint32_t)p.x) & (bounds.w - 1);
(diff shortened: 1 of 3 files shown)


32blit/graphics/tilemap.cpp
```diff
@@ -144,50 +144,72 @@ namespace blit {
...
 
-      wc += dwc;
-      doff++;
-    } while (--c);
+      // skip to next tile
+      do {
+        wc += dwc;
+        doff++;
+        c--;
+      } while(c && (wc.x >> fix_shift + 3) == wcx >> 3 && (wc.y >> fix_shift + 3) == wcy >> 3);
```
Daft-Freak, line 210: Whoops, warning on every compiler...

Daft-Freak:
Hmm, _very slightly_ worse for drawing a non-scaled fullscreen hires tilemap (< a millisecond). Though that's already a suboptimal case... (looking up the tile for every pixel when it isn't going to change for 8 most of the time)

More numbers:


  | master | kart-opt-tilemap
-- | -- | --
hires | 13ms | 13.2ms
hires half scale | 13ms | 7.3ms
hires quarter scale | 12.8ms | 4.4ms
kart mode7 | 12ms | 4.5ms

[Special case for translation only](https://github.com/32blit/32blit-sdk/commit/bdbffd2409cf019632429f7316adb533acbc77c5) : ~5ms

Gadgetoid:
Thank you!

I think the benefits to mode7 far outweigh any minor drawbacks, though I notice this still has optimised cases for fixed-size spritesheets and it would be interesting to see what happens if arbitrary limits like those are lifted. I think some chatter on Discord suggested this change gives enough headroom for that not to be a problem?

I've a strong preference for simplicity over optimisation *but* if we're not hitting draw times that get us comfortably within 50fps.... well it doesn't matter how easy to use something is if it tanks your framerate.

The lack of FPS counter on Tilemap Test/Demo is conspicuous when testing this change :laughing:  but I can't see any perceptible drawbacks in our - granted, very basic - examples. Nice work!

Daft-Freak:
At least for the mode7 case, removing the "power of two bounds" and "16 column spritesheet" restrictions cost ~.2ms each. I think it had a much larger impact on the non-scaled case though. (Since it's then affecting _every pixel_).


(I kind of like the 16 col restriction though as it makes the max-size spritesheet square)

PR #627, opened 2021-03-06
Make the multiplayer example a bit more interesting

ali1234:
Adds a sequence number to each message, implements scrolling, and
shows a count of messages sent and received.

### DIFF
--- a/examples/multiplayer/multiplayer.cpp
+++ b/examples/multiplayer/multiplayer.cpp
@@ -1,4 +1,5 @@
 #include <cstring>
+#include <deque>
 
 #include "multiplayer.hpp"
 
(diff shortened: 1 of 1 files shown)


Daft-Freak:
Yeah, this was _very_ minimal... [Might want to squash a few of the warnings though](https://github.com/32blit/32blit-sdk/pull/627/files) (The usual "int32_t isn't int sometimes" and "VS hates everything"...)

Oh, and I was about to replace the manual `button_pressed` with `buttons.pressed`. (Which probably didn't exist when I wrote the example.)

Gadgetoid:
Extremely useful, thank you!

PR #628, opened 2021-03-08
MP3 looping and restart/play from start

Daft-Freak:
Adds a `restart` method to `MP3Stream` (resets decoder to start) and `loop`/`from_start` flags to `play`.


Gadgetoid:
Nice! Thank you.

Can't wait to... hear... this in action!

PR #629, opened 2021-03-08
Minor SDL multiplayer fixes

Daft-Freak:
... the one thing I didn't test with the last multiplayer PR was _not_ using multiplayer. This should stop the SDL code from attempting to (re-)connect even if multiplayer is disabled. Also improves a bit of the error handling.


Gadgetoid:
D'oh! Also did not test. :facepalm:  Thank you!

PR #630, opened 2021-03-08
Add a wrapper for the microsecond timer

Daft-Freak:
I keep `#include`-ing api_private.hpp to use `get_us_timer` for profiling. This adds a `blit::now_us` wrapper for when you just want to time one thing precisely without setting up the profiler.

(Sometimes isn't any more precise than a millisecond... *looks at Emscripten/Firefox*)

### DIFF
--- a/32blit/engine/engine.cpp
+++ b/32blit/engine/engine.cpp
@@ -27,6 +27,17 @@ namespace blit {
     return api.now();
   }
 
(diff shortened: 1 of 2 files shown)


Gadgetoid:
Timing sidechannel attack mitigations strike again aaaaaa!

Thank you!

PR #632, opened 2021-03-09
Patch launcher reset after flash

### LINKED ISSUE #631: Game metadata corrupt in launcher after flashing a blit
To reproduce:

1. Be in the launcher.
2. `32blit install` any blit with metadata.
3. Select the thing you just flashed in the launcher.

Result: The metadata for the newly flashed blit is corrupted in the launcher. The image can be blank or corrupted, and the rest of the metadata may be missing, including the game name which might be replaced by something like "2".

Attempting to launch the blit from the launcher at this stage may lead to the game and/or launcher crashing.

Restarting the launcher by selecting "exit game" from the menu will clear the problem, and the blit will be displayed correctly and loading will work.

Gadgetoid:
Tries to fix #631 but has its own drawbacks.

Still crashes after about 45 game flashes, but ensures the launcher list is re-loaded immediately after the flash and avoids listing a game that has been deleted.

### DIFF
--- a/firmware/firmware.cpp
+++ b/firmware/firmware.cpp
@@ -614,6 +614,10 @@ static const uint8_t *flash_to_tmp(const std::string &filename, uint32_t &size) 
   return (const uint8_t *)(qspi_flash_address + flash_offset);
 }
 
+static bool blit_is_launcher_running() {
+  return launcher_offset == persist.last_game_offset;
+}
+
 static void tmp_file_closed(const uint8_t *ptr) {
   cached_file_in_tmp = false;
 }
@@ -1083,6 +1087,12 @@ void FlashLoader::handle_data_end(bool success) {
         flash_mapped = false;
       }
       blit_enable_user_code();
+
+      if(blit_is_launcher_running()) {
+        auto game_header = (BlitGameHeader *) (qspi_flash_address + persist.last_game_offset);
+        auto init = (BlitInitFunction)((uint8_t *)game_header->init + persist.last_game_offset);
+        init(persist.last_game_offset);
+      }
     }
   }
 }


firmware/firmware.cpp
```diff
@@ -614,6 +614,10 @@ static const uint8_t *flash_to_tmp(const std::string &filename, uint32_t &size)
   return (const uint8_t *)(qspi_flash_address + flash_offset);
 }
 
+static bool blit_is_launcher_running() {
+  return launcher_offset == persist.last_game_offset;
+}
+
```
Daft-Freak, line 620: Oh, entirely forgot that the firmware had `launcher_offset`... much shorter than my version, which was:
```c++
bool blit_is_launcher_running() {
  auto meta = blit_get_running_game_metadata();

  if(memcmp(meta + 1, "BLITTYPE", 8) == 0) {
    auto type_meta = reinterpret_cast<RawTypeMetadata *>(reinterpret_cast<char *>(meta) + sizeof(*meta) + 8);

    return strcmp(type_meta->category, "launcher") == 0;
  }

  return false;
}
```

firmware/firmware.cpp
```diff
@@ -1083,6 +1087,12 @@ void FlashLoader::handle_data_end(bool success) {
         flash_mapped = false;
       }
       blit_enable_user_code();
+
+      if(blit_is_launcher_running()) {
```
Daft-Freak, line 1091: Only difference in my local code is a comment: `// soft-reset launcher if it's running to update list`

Daft-Freak:
Can't get mine to crash (as usual *sigh*), but can confirm that the second patch does nothing. Just to be sure I applied this to set the background to red if `initialized` was true in `init`:
```diff
diff --git a/launcher/launcher.cpp b/launcher/launcher.cpp
index 850413f1..8c9ca593 100644
--- a/launcher/launcher.cpp
+++ b/launcher/launcher.cpp
@@ -28,6 +28,9 @@ constexpr uint32_t qspi_flash_sector_size = 64 * 1024;
 
 static Screen currentScreen = Screen::main;
 
+bool initialized = false;
+bool was_init = false;
+
 bool sd_detected = true;
 Vec2 file_list_scroll_offset(10.0f, 0.0f);
 Point game_info_offset(120, 20);
@@ -359,6 +362,9 @@ void init() {
   init_lists();
 
   credits::prepare();
+
+  was_init = initialized;
+  initialized = true;
 }
 
 void swoosh(uint32_t time, float t1, float t2, float s1, float s2, int t0, int offset_y=120, int size=60, int alpha=45) {
@@ -393,7 +399,7 @@ void swoosh(uint32_t time, float t1, float t2, float s1, float s2, int t0, int o
 void render(uint32_t time) {
   screen.sprites = spritesheet;
 
-  screen.pen = theme.color_background;
+  screen.pen = was_init ? Pen(255, 0, 0) : theme.color_background;
   screen.clear();
 
   if(currentScreen != Screen::screenshot) {

```

Didn't get a red background. (Wouldn't have been surprised if some of the init code didn't do what I thought it did though...)

(Got almost 300 flashes before stopping the flash loop)

PR #634, opened 2021-03-10
Add missing #pragma once to save.hpp

Daft-Freak:
Whoops, just got some redeclaration errors here...


Gadgetoid:
There can be only one.


Thank you!

PR #635, opened 2021-03-10
Waste less memory on buffers in MP3Stream

Daft-Freak:
~22k -> ~4.7k.

Avoid the file buffer if unneeded (buffer files) and reduce the size of the output buffer (can't remember why it was so big).

Just need to test more on device...


Daft-Freak:
Okay tested buffer/SD, loop/no-loop and nothing exploded. (Even with the worst case of 44100Hz stereo on my slower card)

Gadgetoid:
Less buf is best buf :muscle:

Thank you.

PR #636, opened 2021-03-11
Add a helper to load a TileMap from the TMX struct

Daft-Freak:
Seems like something that should exist...


Gadgetoid:
Hard agree, thank you!

PR #637, opened 2021-03-11
Misc blend changes

Daft-Freak:
Mostly opening this for the first patch, which un-breaks screenshots after I got the alpha channel backwards somehow.

Second patch just avoids some work for alpha == 0 (which ends up as 1 after the "fast" `alpha` function). 

Third is for bugs in `RGBA_RGBA` that were pointed out recently.


Gadgetoid:
Ah! I had not even noticed the conspicuous lack of screenshots in the browser!

Thank you.

PR #638, opened 2021-03-11
Add --help and --info commands to SDL builds

Gadgetoid:
Adds `--help` to document the growing list of commands for SDL builds:

image

Additionally I've added an `--info` command to output the game metadata:

image

And maybe we need `--credits` too?


Gadgetoid:
Now with extra `--credits`

image

Daft-Freak:
That's... not the nicest `#include`... Also requires you to run the contrib script...

Gadgetoid:
It *could* be better. Hopefully GitHub Actions releases can take care of running the contrib script, but maybe that header should be... elsewhere. Or integrated into CMake and .gitignored for changes.

Daft-Freak:
For SDK releases it'll get generated already, but not games. Annoyingly .gitignoring an existing file does nothing. If it was always generated that wouldn't be a problem, but then you're calling GitHub APIs whenever you do a clean build... unless you do "bad CMake" and generate it in the source dir, which probably isn't _that bad_ in this case. Hmm, but then it wouldn't update.

I'm beginning to think the least problematic solution is to just commit the list, run the script before releases and commit an update if there is one...

Gadgetoid:
Yeah I hold with your least problematic solution. It's another thing I have to remember, but forgetting to bump the credits (which don't change super often anyway) may not be the worst crime.

If we were being overcomplicated then I guess the release workflow could check the updated file back in, or raise an automatic PR.

Daft-Freak:
Hmm, could have PR builds check if the file needs updating, then new people get to add themselves...

Gadgetoid:
I think I'm going for a mixture of "let releases and CI continue as normal" and "try and remind myself to commit a somewhat up to date version of the file periodically". :laughing:

PR #639, opened 2021-03-12
Some MP3 buffering fixes

Daft-Freak:
Smaller buffer exposed some existing bugs. Fixes an underrun right at the end of the file preventing the end from being detected and the decode loop using the worst-case output size instead of calculating it, resulting in half-filled buffers most of the time.

(Guess that was why the buffers were so big...)

### DIFF
--- a/32blit/audio/mp3-stream.cpp
+++ b/32blit/audio/mp3-stream.cpp
@@ -92,7 +92,8 @@ namespace blit {
   }
 
   void MP3Stream::pause() {
-    blit::channels[channel].off();
+    if(channel != -1)
+      blit::channels[channel].off();
   }
 
   void MP3Stream::restart() {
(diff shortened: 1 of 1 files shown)


Daft-Freak:
Another fix for a bug that was... interesting to track down.

Calling `MP3Stream::pause` before playing for the first time would attempt to call `.off()` on channel -1. `MP3Stream` calls `pause` at the start of `play` (and `restart`), so this always happens.

The `adsr_phase` member of the "-1th" channel _happens_ to be exactly where `user_code_disabled` is in a release build of the firmware. `off` sets `adsr_phase` to `ADSRPhase::Off` (which has a value of 4), resulting in the firmware now thinking that "user code" is now disabled... except it's still running so nothing happens yet.

... until you open the menu. `blit_menu_render` doesn't draw the game if `user_code_disabled` is set... and closing the menu also checks it and doesn't restore the game render/tick functions.

So yeah, an MP3 bug broke the firmware menu.

Gadgetoid:
That's a heck of the bug! One of the joys of trying to do anything sufficiently complicated on a micro I guess.

PR #640, opened 2021-03-12
Silence git errors when getting version

Daft-Freak:
Prevents the "fatal: not a git repository" message when building from a zip


Gadgetoid:
me: people use repository zips?

also me: OOOH releases!

Thank you!

PR #642, opened 2021-03-16
Ignore API versioning in the firmware updater

Daft-Freak:
Because of the bug in <0.1.11 where the API version was uninitialised, the latest firmware updater doesn't run... unless you already have that firmware. (Also makes sense for any future unavoidable breaking changes.)

The updater doesn't even use almost all of the API anyway. (`buttons` from `blit::tick` and `set_screen_mode` from the init code. `set_screen_mode` is also the first function)

Tested that this runs all the way back to v0.1.2 (first version with .blit). (Excluding any other issues, like the one that deletes the updater)

### DIFF
--- a/32blit-stm32/startup_user.cpp
+++ b/32blit-stm32/startup_user.cpp
@@ -6,11 +6,13 @@ extern void update(uint32_t time);
 extern void render(uint32_t time);
 
 extern "C" bool cpp_do_init() {
+#ifndef IGNORE_API_VERSION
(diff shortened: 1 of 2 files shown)


Gadgetoid:
This may no-doubt prove useful for future utilities which leverage the SDK but not the API. Though it's anyone's guess what those might be... maybe I'll finally get around to writing a bulk flash eraser :laughing:

PR #644, opened 2021-03-16
Add toggleable frame-time meter to launcher

Gadgetoid:
Shows the framerate in milliseconds using a bargraph (1ms = 1 dot) and the frame time in microseconds.

Enter the credits screen and press Y to toggle.

Fixes a couple of instances where the background swooshes were tanking framerate unnecessarily:

* Behind the credits screen
* Behind the screenshot browser

There's no real reason to include a frame time meter in the launcher (edit: except it's a critical part of the 32blit UX so it's probably a good idea to stay on top of anything that might be tanking the framerate), so I'm flagging this TODO: maybe find a way to make this a system feature so we don't have to copypasta it into every darn game.

### DIFF
--- a/launcher/launcher.cpp
+++ b/launcher/launcher.cpp
@@ -396,7 +396,7 @@ void render(uint32_t time) {
   screen.pen = theme.color_background;
   screen.clear();
 
-  if(currentScreen != Screen::screenshot) {
+  if(currentScreen != Screen::screenshot && currentScreen != Screen::credits && selected_game.type != GameType::screenshot) {
     screen.pen = Pen(255, 255, 255);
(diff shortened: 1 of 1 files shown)


Gadgetoid:
With some metrics to tell me how big an impact the Swoosh effect was making on framerate, I've taken some steps to:

1. Make it look less glitchy
2. Optimise the result with `h_span`

As a result the frame time for the effect hasn't changed much, but the vertical bars are gone.

Note: The vertical bars were actually an aesthetic choice and not an optimisation, even though they saved considerable time by skipping pixels. They were intended to prevent the impression of left to right movement in the swooshes, which I found really nauseating. My brain must have locked in on the effect when it was more pronounced, however, since I can no longer see it.

PR #646, opened 2021-03-18
Add a CLAMP_TO_EDGE repeat mode to TileMap

Daft-Freak:
Clamps to the tile at the edge of the map. Almost called it `CLAMP` but `NONE`/`DEFAULT_FILL` also clamp (hmm, kind of like `CLAMP_TO_BORDER` in most graphics APIs).

Though while doing this, noticed that `DEFAULT_FILL` doesn't do what I expected it to do (`default_tile_id` is an index into the map instead of the tileset...)

Also, enum case inconsistency...

### DIFF
--- a/32blit/graphics/tilemap.cpp
+++ b/32blit/graphics/tilemap.cpp
@@ -61,6 +61,15 @@ namespace blit {
       if (repeat_mode == REPEAT)
         return cx + cy * bounds.w;
 
+      if(repeat_mode == CLAMP_TO_EDGE) {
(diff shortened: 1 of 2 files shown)


Gadgetoid:
Have I finally completely lost the plot, or is there nowhere that `repeat_mode` is actually set to a specific value?

Daft-Freak:
Hmm, it isn't initialised anywhere. I completely missed that...

Gadgetoid:
Thank you!

PR #652, opened 2021-04-11
Disable sound when the menu is open

Daft-Freak:
Since `update` isn't getting called you end up with broken audio most of the time (unless you're streaming entirely in the callback). I _think_ this shouldn't break the other user of `sound::enabled` ("user code" disabling for flash stuff)


Gadgetoid:
Thank you!

PR #654, opened 2021-04-12
Rework TMX struct

Daft-Freak:
https://github.com/32blit/32blit-tools/pull/79

Also makes `load_tmx` reject unhandled maps instead of drawing a broken mess.


Gadgetoid:
*grits teeth*

Thank you! Breaking changes in release candidates hooray! :laughing:

PR #655, opened 2021-04-12
Fixing memory leak in Surface

LordEidi:
This is an ugly hack but fixes the memory leak in Surface for now. A refactoring would be the better option.

### DIFF
--- a/32blit/graphics/surface.cpp
+++ b/32blit/graphics/surface.cpp
@@ -51,6 +51,12 @@ namespace blit {
   Surface::Surface(uint8_t *data, const PixelFormat &format, const Size &bounds) : data(data), bounds(bounds), format(format) {
     init();
   }
+  
+  Surface::~Surface()
+  {
+      if (_bNewData) delete[] data;
+      if (_bNewPalette) delete[] palette;
+  }
 
   /**
    * Loads a packed or raw image asset into a `Surface`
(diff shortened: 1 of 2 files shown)


Daft-Freak:
Hmm, another one in `load_from_bmp`: https://github.com/32blit/32blit-sdk/blob/946baad41a9771f5242dd1d5306b364769f5979d/32blit/graphics/surface.cpp#L822-L825

Also, tooManyCamels, not_enough_snakes :smile:

Gadgetoid:
Quite amazed this hasn't been conflicted yet. I guess the engine doesn't get quite the TLC the ports do!

Seems the reason this failed for Emscripten has been lost to the sands of time. Could use a rebase and some gentle nudging into `snake_case` as suggested.

@LordEidi got any time/interest in doing this, or should we grab and tidy up?

Daft-Freak:
Until recently (in uh, "SDK time") changing the layout of `Surface` would break the firmware API. That's possibly why nothing much has changed here.

As an additional comment, the new variables should probably be up with the other variables.

(This was on my list of things to look at, which I should _probably_ be less secretive about...)

LordEidi:
Sorry, wasn't on Github lately (Codeberg is a nice place to be :) ).

Re your question regarding tidying up. We moved away from loading our data through blit::Surface and implemented an Aseprite exporter which generates code (two arrays: one image, one palette and a struct for the rest) which is then directly compiled into the binary. We feed that data structure into the Surface during execution and haven't had any trouble with that so far.

Long story short: I could help you brushing up this pull request. But I haven't followed the 32blit SDK changes lately as closely as I used to. I would need a bit of guidance what you need to accept the change.

PR #656, opened 2021-04-14
Removed out-dated template directory, to avoid confusion

ahnlak:
The old `template` directory is wildly out of date, having evolved into https://github.com/32blit/32blit-boilerplate - leaving it lying around has already caused some confusion with new users not realising that it's outdated, so it probably makes sense to get rid of it completely.


Daft-Freak:
There's a reference to the template in the VS docs... but the .sln file is the one thing the boilerplate doesn't have. (Not that I'm complaining, I think everyone knows my opinion on having two build systems by now :smile:)

ahnlak:
I noticed it was mentioned (in passing) in a couple of the older docs; I should probably run through the docs for this PR too - the `sln` file in `template` is probably so out of date that it causes more problems than it's worth and I *think* we reached the conclusion that VS / CMake was the "official" way?

(that last bit may have just been wishful thinking, though...)

Gadgetoid:
nuked from orbit. Thank you!

I agree Visual Studio sln based builds are bit of an odd duck. I'm not going to defend them, but I wouldn't begrudge anyone who does.

PR #657, opened 2021-04-14
Added --launch_path option to SDL builds

ahnlak:
Added a command line option to SDL builds to make it easier to test Blits that handle file associations; returns the provided file from the `get_launch_path()` function as it does when launching an associated file on the 32blit itself.

TL;DR - saw the `// TODO` comment, and did :-)

### DIFF
--- a/32blit-sdl/UserCode.hpp
+++ b/32blit-sdl/UserCode.hpp
@@ -11,3 +11,6 @@ extern const char *metadata_description;
 extern const char *metadata_version;
 extern const char *metadata_url;
 extern const char *metadata_category;
+
+/* Added a command line ability to specify a launch_path parameter. */
+extern const char *launch_path;
--- a/32blit-sdl/UserCode.hpp
+++ b/32blit-sdl/UserCode.hpp
@@ -11,6 +11,3 @@ extern const char *metadata_description;
 extern const char *metadata_version;
 extern const char *metadata_url;
 extern const char *metadata_category;
-
-/* Added a command line ability to specify a launch_path parameter. */
-extern const char *launch_path;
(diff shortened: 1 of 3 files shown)


32blit-sdl/UserCode.hpp
```diff
@@ -11,3 +11,6 @@ extern const char *metadata_description;
 extern const char *metadata_version;
 extern const char *metadata_url;
 extern const char *metadata_category;
+
+/* Added a command line ability to specify a launch_path parameter. */
+extern const char *launch_path;
```
Daft-Freak, line 16: Hmm, doesn't feel like the right place for this... everything declared else declared here is defined in the "game".

Daft-Freak:
*adds TODO to not forget*

*forgets anyway*

Who needs local testing anyway? :smile:

32blit-sdl/UserCode.hpp
ahnlak: I sort of agree, but I couldn't work out a less "wrong" header file to put it in. I suppose I could have just extern'd it in System.cpp but that would have felt wrong too.

ahnlak:
I took that TODO as an invitation to contribute :smiley:

32blit-sdl/UserCode.hpp
Gadgetoid: But `System.cpp` *includes* `UserCode.hpp` so it's just one level of wrongness indirection :laughing: I'd toss it in `System.cpp` just so fewer files are touched by the change.

ahnlak: Consider it done ;)

Gadgetoid:
Thank you! Might have to try this with my sprite editor.

ahnlak:
Funnily enough I only ended up adding it when I started accidentally writing a BASIC interpreter...

PR #658, opened 2021-04-14
Don't trim whitespace from patches

Daft-Freak:
Corrupting patches when you save them is more than a little annoying. Especially when you're using  `git add –patch` and trying to edit...


Gadgetoid:
Ugh if I had a dollar for every time I had to fix a patch... thank you!

PR #659, opened 2021-04-15
Mostly cleanup patches

Daft-Freak:
Noticed some of these while doing other stuff and a lot more after finding that the toolchain had `-Wno-unused-variable` set.

Last commit fixes the initial flash of uninitialised data after being powered off (by clearing LTDC ram when `persist` needs reset). It does add a warning about writing out of bounds that I can't seem to avoid.


Gadgetoid:
Thank you!

PR #660, opened 2021-04-18
More docs improvements

Daft-Freak:
Main improvement is prioritising the "generic" Windows setup over the VS one, as it covers all the dependencies. This makes the OS setup docs all cover the same thing: installing dependencies and command-line builds.

Also more specific Linux requirements and not listing "building" under "prerequisites" in most of the docs.


Gadgetoid:
Thank you!

PR #661, opened 2021-04-20
Less resets for launcher/game switching

Daft-Freak:
First two commits improve firmware updates (would show the "no launcher" screen while cleaning up). The rest avoid resetting while launching a game (restoring the behaviour before the launcher split) and exiting a game (a TODO since the firmware split).

(Bug fixed by the last commit only exists with #652 + this)


Gadgetoid:
This is pretty awesome and I need to make the time to try it out! Thank you!

Gadgetoid:
Hasn't exploded for me! Perhaps... time... for a release? :laughing: 

Thank you!

PR #662, opened 2021-05-04
Power/sleep stuff

Daft-Freak:
An attempt at sleeping between updates (only gets to sleep for max 45us due to audio interrupts...) and some improvements to "Power Off" (similar to "idle sleep" now, but enters shipping mode when not charging/full for max battery saving).

I was going to open this with the reset one, but they ended up conflicting... and then I got distracted...


Gadgetoid:
> and then I got distracted...

Relatable.

Thank you, I'll check this out.

Gadgetoid:
No issues so far, onward and upward! Thank you.

PR #664, opened 2021-05-12
Fix LTDC active area

Daft-Freak:
It was 321x241. Also calculates the accumulated values to be more clear.

(Found while experimenting with a Pico connected to the LCD connector...)


Gadgetoid:
> (Found while experimenting with a Pico connected to the LCD connector...)

:eyes:

Gadgetoid:
Magic-number-b-gone! Thank you.

I think I might have introduced the 321x241 bug when I was fixing the gap on one side, maybe, possibly...

PR #665, opened 2021-05-20
Fix the VS device build

Daft-Freak:
VS adds some overrides that break the build... Didn't spot that when I wrote this section. Another way to fix this would be to include a mostly-configured `CMakeSettings.json` in the boilerplate.

Still possibly not the most reliable config as it depends on VS shipping a compatible toolchain...


Gadgetoid:
Thank you- it's high time I set up a Windows box and gave ol' VS a try. It's been quite a while!

PR #666, opened 2021-05-22
Add release firmware install instructions

Daft-Freak:
This at least covers the thing users want at the top. The rest may still need cleaning up...


Gadgetoid:
Thank you! This is definitely helpful.

PR #667, opened 2021-05-22
Fix some paths

Daft-Freak:
CMake finds most of these for us as part of its compiler detection(https://github.com/Kitware/CMake/blob/v3.16.9/Modules/CMakeFindBinUtils.cmake#L112). Using find_program for the other two avoids a mix of full paths and names in the cache. 

Which helps if there's weird stuff going on with PATH... *looks at VS*


Gadgetoid:
VS, always the troublemaker!

Thank you!

PR #668, opened 2021-05-23
Update Windows-WSL.md

ThatOtherOtherBob:
Clarifying pip3 isn't the continuation of a long line wrapping and adding another line needed before my PC could install gcc.

### DIFF
--- a/docs/Windows-WSL.md
+++ b/docs/Windows-WSL.md
@@ -30,10 +30,11 @@ After that, proceed to the Microsoft Store to download Ubuntu for WSL.
 
 ### Installing requirements inside WSL
 
-The following requirements enable cross-compile to 32Blit via ARM GCC:
+The following 3 commands enable cross-compile to 32Blit via ARM GCC:
 (These are similar to the [Linux requirements](Linux.md#prerequisites), see those for more details)
 
 ```
+sudo apt update
 sudo apt install gcc gcc-arm-none-eabi unzip cmake make python3 python3-pip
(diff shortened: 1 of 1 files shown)


Gadgetoid:
Might be worth splitting these into different code-fences so it's clearer at a glance. IE:

The following commands enable cross-compile to 32Blit via ARM GCC:
(These are similar to the [Linux requirements](Linux.md#prerequisites), see those for more details)

```
sudo apt update
sudo apt install gcc gcc-arm-none-eabi unzip cmake make python3 python3-pip
```
And you should also install the 32Blit tools:

```
pip3 install 32blit
```




(Note: "construct" and "bitstring" are automatically installed as part of "32blit" now so we can simplify that line)

Daft-Freak:
So the Linux docs have a blank line between the two here, also doesn't `apt update` before the install... and has a separate bit for the unneeded deps. Guess there's more tidying to be done...

Gadgetoid:
I think it's probably on us to grab the gist of this change and improve the documentation. Probably ties in with #691 and #692 but I pretty much ditched Windows so I'll have to provision a VM or something to kick the tyres on the docs and wrap these up.

Daft-Freak:
Running a windows VM to test what is essentially a linux VM seems a bit silly. I have a windows setup handy... but I don't think I have WSL set up on it... Not sure how much this ties in with the other two, those are actually running on windows.

(These changes look reasonable, but we should maybe make the WSL docs as close to the linux ones as possible)

PR #672, opened 2021-05-25
Link the firmware against the right stdlibs

Daft-Freak:
Toolchain used `NONPIC_STDLIBS`, CMakeLists used `NOPIC_STDLIBS`, resulting in the firmware using the system libraries. If you're using Ubuntu these are similar enough that this doesn't explode.


Gadgetoid:
Seems to save a few k, too :thinking:

Issue #287, opened 2020-03-09
2 noob issues and fixes


mgarcia-org:
in file: https://github.com/pimoroni/32blit-beta/blob/master/docs/Linux.md

On the "sudo apt install ...." line, it should include:
python3-setuptools 

And , line "pip3 install construct bitstring" requires sudo

mgarcia-org:
The last one also on tools doc in src (updated 11 days ago):
https://github.com/pimoroni/32blit-tools/blob/master/src/README.md

Which is duplicated in root folder (updated 17 days ago):
https://github.com/pimoroni/32blit-tools/blob/master/README.md

Gadgetoid:
Every time I stick "sudo" before a command in documentation, someone drives by to say I should leave the bare command and add something like "You might need sudo." 😆 

I'm not *quite* sure how we should approach this, but it probably needs standardised somehow. Some Python setups will happily install packages for just the current user- but the root/user environments don't cross pollinate so it can - and does, in my case - confuse the heck out of people.

Oh Python 🙄 

I think you're right re `python3-setuptools` though.

Daft-Freak:
I'm going to close this as `sudo pip install` is even less likely to be the right thing to do these days...

(Though we might need a new issue for updating the tools instructions in general...)

merwok:
These days, installing setuptools is mostly not required: most projects are downloaded as wheels, that don’t have build steps, and those that use sdists have metadata that specifies their build backend (setuptools or other) to the installer tool.

PR #676, opened 2021-05-28
Add python3-setuptools to apt line

Daft-Freak:
Small fix for an older issue. (#287)

This is only a "recommends" in Debian stable/Ubuntu LTS so may not get installed automatically.

(Didn't do anything about `sudo pip` as I'm a little against that...)


Daft-Freak:
Oh, will conflict with #668

Gadgetoid:
Thank you- I think #668 needs some TLC anyway so it should be fiiinnne... *hides*

Issue #519, opened 2020-12-31
Timer has a PAUSED state and is_paused() check but no pause functionality.


Loxrie:
I've been working on adding this to the logic and my changes work though I'm still testing.  However I'm not sure it's in keeping with the style of a lot of 32blit code as it does add some internal state checks to start() and pause():

vis: https://github.com/Loxrie/32blit-beta/blob/feature/pause/32blit/engine/timer.cpp

What would be your preference for this kind of code?  

- start() should be left well alone and pause() should be accompanied by a resume() function.  State checks removed. Consumer of Timer responsible for ensuring state transitions are acceptable.
- The state checks are OK.

I'm unsure of the best approach especially with regards to potential bloat over time and any impacts on speed.

PR #677, opened 2021-06-01
Timer/Tween improvements

Daft-Freak:
Fixes:
- Deleting a `Timer` or anything containing a timer causing a use-after-free.
- Restarting a non-infinite timer not resetting the loop count and turning it into an infinite timer
- Half implemented pause (#519)
- Makes the stored loop count the same size as the argument to `init` (int32 vs int16)

WIP because:
- Switched the function pointer to `std::function`, which makes the `user_data` pointer a bit redundant as you can use bind/lambdas to pass stuff in
- Moved most of `init` to the constructor, but didn't remove `init` because breakage
- Need to update `Tween`

### DIFF
--- a/32blit/engine/timer.cpp
+++ b/32blit/engine/timer.cpp
@@ -41,14 +41,12 @@ namespace blit {
     if(state == UNINITIALISED)
       timers.push_back(this);
 
-    // reset loop counter if we're restarting
-    if(state == STOPPED || state == FINISHED)
-      loop_count = 0;
-
     if(state == PAUSED)
       started = blit::now() - (paused - started); // Modify start time based on when timer was paused.
-    else
+    else {
       started = blit::now();
+      loop_count = 0;
+    }
 
     this->state = RUNNING;
   }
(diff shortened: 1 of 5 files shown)


Gadgetoid:
I'm sure `user_data` is only used by Lua at this point so we can either:

a. Leave it, maybe some people like *the old way* and might use it :laughing: 
b. Drop it, because I can't imagine anyone has picked it up yet

The `init` function was intended to provide some way to change the speed/loop count of a timer on the fly. However it feels like there should be a more explicit way to achieve this since... calling `init` again feels wrong. I guess functionality could be duplicated verbatim in the constructor, and `init()` marked deprecated for a grace period.

Daft-Freak:
I really just didn't want to revert a commit you just added :smile:

Using `init` to change stuff was a bit broken anyway, it would add the timer to `timers` again (fixed here as well). Also all the struct members are public so you can change everything anyway... (possibly some of them shouldn't be)

Gadgetoid:
> I really just didn't want to revert a commit you just added

Haaaaaaaaa I'm not precious if you have a better approach. I've already got a pending change to Lua, switching it over to sweet sweet lambdas. Makes my code look like I know what I'm doing :eyes: .

Gadgetoid:
Thank you! Better bosh my patch up to 32blit-lua!

PR #679, opened 2021-06-04
32blit.hpp cleanup

Daft-Freak:
Noticed this needed some updating...

Last two commits should probably be the other way around.


Gadgetoid:
Ooof, wading through the sticky mud in the depths of the 32blit code swamp there! ... 

Thank you :laughing:

PR #680, opened 2021-06-04
Do a full scan_flash before launching after flash

Daft-Freak:
Fixes flashing something with file associations while it's running. (It would try to launch the old one the next time you opened a file)

I was trying to avoid work by only setting the launcher offset instead of this... resulting in this fun corner-case.


Gadgetoid:
:grimacing: whoops! :laughing: thank you! Everything's too complicated for our own good now... but sheesh is it ever incredible.

PR #682, opened 2021-06-07
Bump Emscripten version

Daft-Freak:
Should fix the "unexpected hash" errors caused by https://dl.bintray.com/homebrew/mirror/jpeg-9c.tar.gz returning "Forbidden!" again...

TODO: boilerplate too

### DIFF
--- a/.github/workflows/Emscripten.yml
+++ b/.github/workflows/Emscripten.yml
@@ -48,7 +48,7 @@ jobs:
     - name: Configure CMake
       shell: bash
       working-directory: ${{runner.workspace}}/build
-      run: emcmake cmake $GITHUB_WORKSPACE -D32BLIT_DIR=$GITHUB_WORKSPACE -DCMAKE_BUILD_TYPE=$BUILD_TYPE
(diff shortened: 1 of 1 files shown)


Daft-Freak:
Hmm, seems a lot slower...

Daft-Freak:
Possibly because `Release` is now `-O3`. https://github.com/emscripten-core/emscripten/blob/main/ChangeLog.md#2012-01092021

Gadgetoid:
12m :grimacing:

I guess we still have the option of overriding that back to -O2. If it ain't broke...

Daft-Freak:
Tried to get ccache working (support added recently) didn't succeed, so forced it back to -O2 for now.

(Also tried just overriding for linking, which had about half the effect of doing it everywhere)

Gadgetoid:
Was worth a try I suppose! Thank you.

Looks like users will build with the default options, too, so we don't lose anything by speeding up CI.

Daft-Freak:
> TODO: boilerplate too

...since this just got me again: https://github.com/32blit/32blit-boilerplate/pull/21

PR #683, opened 2021-06-07
Check that calling the tools didn't fail

Daft-Freak:
If the tool does fail we'll either fail to include the generated file on the next line, or ignore the error and continue using stale config.

Error you get without this if metadata/assets.yml is missing:
```
...
FileNotFoundError: [Errno 2] No such file or directory: '/batt/repos/.../metadata.yml'
CMake Error at /batt/32blit/32blit-stm32/executable.cmake:60 (include):
  include could not find load file:

    /batt/repos/.../build/metadata.cmake
Call Stack (most recent call first):
  CMakeLists.txt:6 (blit_metadata)
```


Gadgetoid:
Ooof, handy, thank you!

PR #684, opened 2021-06-10
Use transparent_index

Daft-Freak:
Makes `Surface::transparent_index` actually do something. Always uses the value from the dest surface, though I guess P->P blits don't make sense if both don't use the same palette anyway.


Gadgetoid:
Ran through the examples and nothing is on fire... except doom fire. Thank you!

PR #686, opened 2021-06-18
Metadata for files

Daft-Freak:
Loads metadata for a file from a `.blmeta` file next to it. (Needs a way to generate that file, like [this](https://github.com/Daft-Freak/32blit-tools/commit/eff26443678cf57b1fd1cfc52b4176466ffac803))

[Test file for a Tetris rom](https://github.com/32blit/32blit-sdk/files/6677730/Tetris.gb.zip)

### DIFF
--- a/launcher/launcher.cpp
+++ b/launcher/launcher.cpp
@@ -116,6 +116,10 @@ void load_directory_list(std::string directory) {
 
 bool parse_file_metadata(const std::string &filename, BlitGameMetadata &metadata, bool unpack_images = false) {
   blit::File f(filename);
+
+  if(!f.is_open())
+    return false;
+
   uint32_t offset = 0;
 
   uint8_t buf[sizeof(BlitGameHeader)];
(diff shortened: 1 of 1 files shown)


Gadgetoid:
I don't know if it's a bug or a feature, but creating "snake.lua" and "snake.lua.blmeta" gives me my "Susi Snake" entry in "ROOT" complete with metadata, yay! *However* it removes the "Launches with: Lua" text. I think it may be useful to preserve that since it's virtually impossible to tell that you're running a Lua app (does it matter!?) otherwise.

Daft-Freak:
Hmm yeah, did wonder if there should be _something_ displayed for "actually launching something else to run this". The "Launches with: ..." text was using the description field, might make it too long if we just append that back on...

Gadgetoid:
This has been working well for me anyway, so I think we can forge ahead and figure out how the launching application might be identified in future. Thank you!

PR #689, opened 2021-07-07
File fixes

Daft-Freak:
Trying to take more screenshots for BlitHub, found a leak when you exit a game with files open... and only because I couldn't enable storage mode.

This closes all files on a game switch, which shouldn't break anything as the firmware doesn't keep any files open or go through the API `open_file` function. Also fixes the flash cache to actually set the "is something in there" flag... and resets that on game switch as well.


Daft-Freak:
Oops, based on the FatFS PR... guess that means I've tested that a lot :smile:

Gadgetoid:
That's a very *me* last commit. Heck of a PR, though, thank you! I wonder how many more little gotchas are left in the codebase.

PR #690, opened 2021-07-08
Make Rect.size const

Daft-Freak:
Tiny change to match everything else. (... and fix my build after const-ing some things)


Gadgetoid:
Thank you!

PR #691, opened 2021-07-15
Revise Windows docs, prefer CMake

Gadgetoid:


docs/Windows-VisualStudio.md
```diff
@@ -100,7 +103,8 @@ There is also a skeleton game project created for you at https://github.com/32bl
       "inheritEnvironments": [ "gcc-arm" ],
       "variables": [],
       "intelliSenseMode": "linux-gcc-arm",
-      "cmakeToolchain": "[path...]/32blit-sdk/32blit.toolchain"
+      "cmakeToolchain": "[path...]/32blit-sdk/32blit.toolchain",
+      "environments": [ { "PATH": "${env.PATH}" } ]
```
Daft-Freak, line 107: While that does fix the build, it usually breaks configuring as it resets the PATH set by the "gcc-arm" environment containing the cross tools... (Works after configuring as all the paths are cached)

docs/Windows-VisualStudio.md
```diff
@@ -1,27 +1,24 @@
...
 
 You will need Visual Studio 2019 (preferably version 16.4).
 
-Make sure you install C++ desktop development support.
-
-You will also need to download SDL2 development libraries from the [SDL homepage](https://www.libsdl.org/download-2.0.php). Here find the latest version of the VC development libraries (at the time of this writing SDL2-devel-2.0.10-VC.zip). Additionally, download SDL2_image from [here](https://www.libsdl.org/projects/SDL_image/) (SDL2_image-devel-2.0.5-VC.zip) and SDL2_net from [here](https://www.libsdl.org/projects/SDL_net/) (SDL2_net-devel-2.0.1-VC.zip).
-
-Place these in the `vs\sdl\` folder. You will need to merge the include/lib directories. If you are using CMake/Open Folder, these are downloaded automatically.
-
+Make sure you install "Desktop development with C++". If you plan to build for 33blit, you'll also need "Linux development with C++", making sure "C++ CMake tools for Linux" and "Embedded and IoT development tools" are selected in the "Optional" section.
```
Daft-Freak, line 21: Ah, the new improved 33blit :smile:

docs/Windows-VisualStudio.md
Gadgetoid: I wonder if it can do something like append to the path rather than completely trounce it. I assume you've probably bashed your head against this quite a bit! :grimacing:

docs/Windows-VisualStudio.md
Gadgetoid: One more bit is nothing to sniff at! Though memory alignment of smaller types is quite complicated so you're encouraged to use `uint33_t` everywhere.

docs/Windows-VisualStudio.md
Gadgetoid: That said, I can't seem to get it to break configuration. Even after deleting the cache.

Daft-Freak: Yeah I kept trying to workaround it in `32blit setup`... and eventually got annoyed enough to remember my MS account and report it.

Daft-Freak: > That said, I can't seem to get it to break configuration. Even after deleting the cache.

Possibly finding some other installed toolchain instead of the VS supplied one? It definitely doesn't find the one shipped with VS with that override.

Gadgetoid: Oooh that's something to check. Might be prudent to remove the ARM GCC toolchain since that's in my path.

I strongly suspect there wont be a fix for this imminently and we'd have to bless the command-line procedure for actual on-device builds. Which isn't so bad.

Or, I guess, if it's gonna detect ARM GCC and work... that's also fine?

Daft-Freak: Hah, maybe it'll work in VS 2022, who knows... There's always VS Code (and probably most other editors) to make the "command-line" builds less... command-line.

Guess the problem with an "external" toolchain is it might stop working, as finding that instead of the VS one probably isn't supposed to work. Might mess up intellisense or something :shrug:

Gadgetoid: You're right- it finds the ARM GCC toolchain I installed. Makes sense!

This... sort-of... fix does the trick in both cases:

```json
"environments": [ { "PATH": "${env.ProgramFiles(x86)}/Microsoft Visual Studio/2019/Community/Linux/gcc_arm/bin/;${env.PATH}" } ]
```

It's ugly, and I'd feel a lot happier if I could find a way to get the *full* path to the currently running Visual Studio but there doesn't seem to be a way. Also not sure if I'm subtly breaking something by supplying the wrong executable, my install isn't exactly clean!

Daft-Freak:
Guess this should mention somewhere that `32blit setup` does all the config stuff for you if you select the VS option. (Oops, completely forgot to document that tool anywhere)

Gadgetoid:
I never got as far as to get `32blit setup` working when I did my Windows install. I guess I should have tried that first :laughing:

Daft-Freak:
I think the changes other than the PATH thing still make sense. (I don't know if the bug got fixed, but it got closed as "lower priority" so I'm assuming it hasn't). I did just delete the entire "option 1" section though.

Guess I have to make that "33blit" myself now. :laughing:

docs/Windows-VisualStudio.md
Daft-Freak: In VS2022 it seems that the "IoT" configs are gone, so the only way to get this to work is to install the Arm toolchain and manually add a config similar to the one above. (Including the PATH hack...)

Anyway, I was in the area so I've rebased this.

PR #692, opened 2021-07-15
Tweak Windows CLI docs

Gadgetoid:


docs/Windows.md
```diff
@@ -3,25 +3,55 @@
...
 - [Prerequisites](#prerequisites)
-- [Using Visual Studio](#using-visual-studio)
+- [Setting Up](#setting-up)
 - [Building & Running on 32Blit](#building--running-on-32blit)
 - [Building & Running Locally](#building--running-locally)
   - [Build Everything](#build-everything)
+- [Troubleshooting](#troubleshooting)
+  - [Multiple definition of `cleanup_glue`, `_reclaim_reent` etc](#multiple-definition-of-cleanup_glue-_reclaim_reent-etc)
+
+To build using Visual Studio [see here](Windows-VisualStudio.md). The rest of these instructions cover command-line builds.
```
Daft-Freak, line 13: Hmm, I had that after the prerequisites section because the VS docs don't mention installing python/the 32blit tools (and git?).

docs/Windows.md
```diff
@@ -57,27 +106,35 @@ cmake -G"NMake Makefiles" ..
...
 ```
 
-When the build completes you should be able to run any example.
+When the build completes you should be able to run/install any example.
+
+## Troubleshooting
+
+### Multiple definition of `cleanup_glue`, `_reclaim_reent` etc
+
+If you're building *for* 32blit using the ARM GCC toolchain, and it fails with a wall of text full of multiple definition errors, you've probably installed GCC 10. You should use the `gcc-arm-none-eabi-9-2020-q2-update-win32.exe` package.
```
Daft-Freak, line 140: This _should_ be fixed now.

docs/Windows.md
```diff
@@ -3,25 +3,55 @@
...
+    "fontSize": 10,
+    "guid": "{5ee0706e-b015-46b2-98a3-2122a8e627d3}",
+    "historySize": 9001,
+    "icon": "C:\\Program Files (x86)\\Microsoft Visual Studio\\2019\\Community\\Common7\\IDE\\Assets\\VisualStudio.70x70.contrast-standard_scale-80.png",
+    "name": "Developer Command Prompt for VS2019",
+    "padding": "0, 0, 0, 0",
+    "snapOnInput": true,
+    "startingDirectory": "%USERPROFILE%"
+}
+```
```
Daft-Freak, line 44: Used
```json
            {       
                "commandline": "cmd.exe /k \"C://Program Files (x86)//Microsoft Visual Studio//2022//BuildTools//Common7//Tools//VsDevCmd.bat\"",
                "guid": "{5ee0706e-b015-46b2-98a3-2122a8e627d3}",
                "name": "Developer Command Prompt for VS2022",
                "startingDirectory": "%USERPROFILE%"
            }
```

for VS 2022 build tools. (Couldn't find anything to use as an icon and the rest seems optional)

docs/Windows.md
```diff
@@ -3,25 +3,55 @@
...
- - [Git for Windows](https://git-scm.com/download/win)
- - [Python](https://www.python.org/downloads/)
- - At least the [Visual Studio build tools](https://visualstudio.microsoft.com/downloads/#build-tools-for-visual-studio-2019) (For local builds, select "C++ build tools"). This includes a copy of CMake. A full Visual Studio install can also be used.
- - The [GNU Arm Embedded Toolchain](https://developer.arm.com/tools-and-software/open-source-software/developer-tools/gnu-toolchain/gnu-rm/downloads) (For 32blit device builds, use the 9-2020-q2-update version). Make sure to select the "Add path to environment variable" at the end of the setup.
+Install the 32blit tools:
 
-Now open "Developer Command Prompt for VS 2019" and install the 32blit tools:
 ```
-py -m pip install 32blit
+pip install 32blit
```
Daft-Freak, line 53: My python install (3.10.2) doesn't seem to have `pip` (or `python`) only `py`.

docs/Windows.md
```diff
@@ -31,20 +61,39 @@ WARNING: The script 32blit.exe is installed in 'C:\Users\[Name]\AppData\Local\Pr
...
+
+### On 32Blit
+
+If you want to run code on 32Blit, you should now refer to [Building & Running On 32Blit](32blit.md). On Windows you will need to add `-G"NMake Makefiles"` to your cmake commands, for example:
 
-If you want to run code on 32Blit, you should now refer to [Building & Running On 32Blit](32blit.md). You will need to add `-G"NMake Makefiles"` to your cmake commands and use `nmake` instead of `make`. For example:
 ```
 mkdir build.stm32
 cd build.stm32
 cmake -G"NMake Makefiles" -DCMAKE_TOOLCHAIN_FILE=../32blit.toolchain ..
```
Daft-Freak, line 79: The CMake we get from VS should be new enough for
`cmake -B build.stm32 -G"NMake Makefiles" -DCMAKE_TOOLCHAIN_FILE=./32blit.toolchain`
instead of this block.

(and then `cmake --build build.stm32 [...]`)

... but I guess If we're doing that we should update other places too...

docs/Windows.md
Daft-Freak: Yep, using 10.3.1 works fine

Daft-Freak (commented):
Tested this out with up-to-date VS 2022. Mostly worked out.

Daft-Freak:
Dropped the windows terminal (developer cmd prompt seems to be there by default now) and arm toolchain troubleshooting (fixed) parts.

docs/Windows.md
Daft-Freak: This is still true, should drop this change too...
snakes

PR #694, opened 2021-07-16
Rix reading battery fault status

Daft-Freak:
Attempting to read the fault register and the status register in one read would result in the fault register always reading as 0. We're still not doing anything with the fault status, but at least it reads correctly now.

Also some cleanup.


Gadgetoid:
Good catch, thank you! Battery status and info has always been one of the things that stymies me!

PR #695, opened 2021-07-27
Skip time spent with the menu open

Daft-Freak:
This avoids the issue where if you open the menu for a while and close it the game appears to hang/skip forward as it gets a huge number of `update` calls. (The same thing could happen to the menu, but it doesn't do much in update so was a lot harder to notice)

Also skips "menu time" for timers/tweens (tested with tween-demo). Didn't increase the API version despite adding a new field as nothing breaks with a mismatch (may sometimes trigger the skip code incorrectly, but only once and at the start of the game where it wouldn't call `update` anyway).

(This has been broken since 0b260b43b7ab37067746d9cbc8764734d67535ff)

### DIFF
--- a/32blit-stm32/Src/32blit.cpp
+++ b/32blit-stm32/Src/32blit.cpp
@@ -410,6 +410,8 @@ void blit_init() {
 
     blit::api.get_metadata = ::get_metadata;
 
+    blit::api.tick_function_changed = false;
+
   display::init();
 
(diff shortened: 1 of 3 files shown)


Gadgetoid:
The number of examples that just shrug and keep going beneath the menu is ... :laughing: can't beat the ol' update-in-render approach.

Thank you.

PR #696, opened 2021-07-27
Remove "Exit Game" from fw menu in Launcher

Gadgetoid:
If the running executable category is "launcher" this will hide the "Exit Game" menu item, since we're not in a game anyway.

Additionally moves SystemMenu from "32blit-stm32" to "firmware" where it belongs. (Though, to be fair, the line between the HAL and the firmware is... hazy)

### DIFF
--- a/32blit-stm32/Src/SystemMenu/firmware_menu.cpp
+++ b/32blit-stm32/Src/SystemMenu/firmware_menu.cpp
@@ -47,14 +47,14 @@ static Menu::Item firmware_menu_items[] {
     {BACKLIGHT, "Backlight"},
     {VOLUME, "Volume"},
     {SCREENSHOT, "Take Screenshot"},
-    {Menu::Separator,nullptr},
+    {Menu::Separator, nullptr},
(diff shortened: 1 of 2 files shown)


Daft-Freak:
But `32blit-stm32` uses the menu classes (and `firmware` doesn't), so now you can't link it to anything other than `firmware` without missing symbols... (not that `blit_executable_int_flash` has _any_ other users) Yeah, there's not really a line there, but I have tried to avoid making dependencies in the wrong direction.

Hmm, I have used that menu item as a "Restart Launcher" option in the past, probably not recently though...

Gadgetoid:
Ugh, you're right, all the menu setup is in `32blit.cpp`. Aaaaaah, it hurts :laughing: 

I guess at least:

```c++
item.label = strcmp(metadata.category, "launcher") == 0 ? nullptr : "Exit Game"
```

Could become:

```c++
item.label = strcmp(metadata.category, "launcher") == 0 ? "Restart Launcher" : "Exit Game"
```

Which makes it much clearer about what the menu item is doing.

PR #697, opened 2021-08-04
Non-pico patches from the pico branch

Daft-Freak:
An assortment of patches from the pico branch that aren't really pico related.

- A few example changes
- Small optimisations that are more noticeable on the pico (saves ~20ms in `hsv_to_rgb` there, the examples that do that per-pixel are still slow though)
- Half-working CDC host with TinyUSB-based device (will connect and receive, sending from the host hangs though.)


Gadgetoid:
Wow- the button handling in particle was *old*!

Thank you!

PR #700, opened 2021-08-08
Fix SDL get_metadata description

Daft-Freak:
Oops.

Happened to notice while ~~completely breaking~~ tweaking "the cupboard".


Gadgetoid:
Aaaaah, yes, I see it now! Thank you.

PR #701, opened 2021-08-10
Make geometry into a game

Gadgetoid:
* Add crude sound effects
* Add score and lives
* Add energy bar, movement/shooting costs energy
* Add collission with asteroids, damage costs energ
* Damage to 0 energy loses a life
* Invincible after losing a life until 1st shot
* Score/lives reset after 0


guru:
Request for some tweaks to make it MVP complete:
- Colour code boulders based on area.
- Cycle spectrum when they're small enough to 'vape'
- Big text for score
- Lives as smaller colourcycling/spinning ships top right?

PR #702, opened 2021-08-12
More non-pico pico branch patches

Daft-Freak:
First commit saves 512 bytes of RAM by moving the `nodes` array in `Surface::polygon` to the stack. Second one avoids constant pens taking up RAM by making the constructors `constexpr`. (Lost a whole 12 bytes to the pens in `watermark` before this!)

Last one replaces 96-100 calls to `pixel` per-tile in rainbow ascent with two `h_span`s and a `rectangle`. (This was actually copied from my lua port as it had a noticeable effect on the pico)


Daft-Freak:
Apparently I forgot to press the create button... there are now another four commits tidying up some fatfs code and moving it up so it can be reused.

Gadgetoid:
Awesome, thank you! I'll have to dig up the changes I made to port Scrolly Tile to 120 width (it's *very* opinionated about being 16 tiles wide).

PR #703, opened 2021-08-13
Pico support

Daft-Freak:
This should be at the point where it doesn't affect anything non-pico related. A significant chunk of this is messing around with st7789, some adding things that get removed again.

Also, partial support for RGB565 is (re-)added. They can be used for drawing to and blitting RGB/paletted surfaces to. Blitting from RGB565 to anything (including itself) is not supported.

Still a few things a bit unfinished:
- random needs a seed
- storage stuff (fat with 4k blocks in the last 1/4 of flash, requires manual format, doesn't prevent the device and USB accessing at the same time)
- some VGA stuff (haven't tested it in a while)
- maybe don't always overclock to 250MHz?
- needs a rebase
- USB descriptors are mostly TinyUSB placeholders


ali1234:
What hardware can I use to test this?

Gadgetoid:
Words cannot elucidate how awesome this is. That was an exciting button to push!

Thank you for tidying your random pet project up into something merge-worthy. The results really are truly incredible and I never thought I'd see the 32blit SDK run *so well* on another, slower platform.

Now... bring on the ports :laughing:

PR #704, opened 2021-08-13
Raycster: improve sprite rendering and more

Gadgetoid:
- Rip out scanline rendering of sprites and use stretch_blit_sprite and screen.clip instead, this isn't really faster but it's a little less painful to look at.
- Remove tile visibility map, it did nothing much but introduce pop-up bugs
- Add some very minor optimisations
- Delete some dead code
- Add and load "wasp" sprite


Gadgetoid:
* Add wasp sprites from @ali1234
* Merge bug spray entities with "map_sprites" entities
* Handle spraypaint as a sprite, so it's clipped behind walls
* Handle collission between sprites/walls/wasps
* Wasps follow player, collide with walls, take damage
* Spray damages wasps
* Wasps damage player
* "Game Over" condition
* Add player velocity to spray


# TODO
* [ ] Bug spray doesn't feel like it's coming from the can, doesn't start small enough and doesn't swell into clouds quite so well as it should
* [ ] Add bug spray sprite directly into spritesheet
* [ ] Fix floating point crashes
* [ ] Wasp damage/death animation
* [ ] "Wasps left" indicator?

Gadgetoid:
I would have totally got 'round to that, if I could just stop being ruthlessly obsessed with my blog CSS and HTML for ten minutes :laughing: 

Thank you!

Daft-Freak:
Hah, I'm definitely not using 32blit "maintenance" as an excuse to avoid fixing my own stuff.

BTW, I ~played some games~ did some testing with the current code to maybe move towards doing an actual release.

Gadgetoid:
I think we are rather overdue a release! I'll reserve some time one evening to try and draft up some notes.

Daft-Freak:
Release notes are the main reason I didn't do it myself :laughing: 

Also assuming you want to merge this before someone looks at #744

Gadgetoid:
Ooof I thought about that last week - not sure if we:

Add unit tests for actual proper coverage and spit the SDK wholly from the distribution of 32blit games.

OR

Add a list or manifest or submodules (or whatever other technique is in vogue at the moment) of projects to be cloned by the workflow and built against the SDK automatically as a means of CI.

The latter feels like it would be reproducing your build box, could be non-required test and used to detect the impact of (some) breaking changes more directly.

Kinda feels like we should do both :skull:

Daft-Freak:
I was mostly thinking that the examples should demonstrate some SDK functionality in a simple enough way. TBH they're a large amount of the "documentation" right now.

raycaster (and voxel) don't seem to be a good fit for this. (Also voxel has a giant data file that slows down the build a bit)

Anyway, should probably discuss this over there.

Gadgetoid:
Can progress this in some future project, it's been in PR hell for too long! Thanks for the assist.

PR #706, opened 2021-08-20
pico: Tweak ST7789 gamma and content-adaptive brightness

Gadgetoid:
Switches to manufacturer supplied gamma settings, which are clearly better with more brightness/contrast.

Turns on content-adaptive brightness, which seems to give a significant boost in display brightness. May need testing with various games to see if there are any artifacts or drawbacks - other than potential additional power usage - of using this feature.

### DIFF
--- a/32blit-pico/st7789.cpp
+++ b/32blit-pico/st7789.cpp
@@ -160,8 +160,10 @@ namespace pimoroni {
         command(reg::VRHS, 1, "\x12");
         command(reg::VDVS, 1, "\x20");
         command(reg::PWCTRL1, 2, "\xa4\xa1");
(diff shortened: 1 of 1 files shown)


Daft-Freak:
Hmm, it's definitely... bright. Looking at a random ST7789 datasheet, looks like that `CAB` (or `WRCACE `) setting is for "medium colour enhancement" and adaptive brightness is off?

Gadgetoid:
Oof, yeah I think I turned off content-adaptive brightness because it appeared to do nothing while "Colour Enhancement" appears to make it... B R I G H T. Possibly too bright. Pure gamma settings on their own still feel a little dim, however, despite being much nicer than the previous setup.

It doesn't help that this is a little subjective.

Daft-Freak:
Hmm, hardware-test/serial-debug (HSV rainbow background) look a bit bad with "enhancement" on. (too bright at the top)

PR #707, opened 2021-08-22
Set image-rendering on Emscripten canvas

Daft-Freak:
Avoids blurriness with hidpi screens. The shell used for the example browser already did this. (Specified twice for browser support reasons)


Gadgetoid:
I hadn't come across this somehow! Thank you.

PR #714, opened 2021-09-15
Allow building for pico by specifying a toolchain, same as 32blit

ali1234:
To build a project for pico just do:

    cmake -DCMAKE_TOOLCHAIN_FILE=../../32blit-sdk/pico.toolchain ..

You may optionally override PICO_BOARD and PICO_SDK_PATH

### DIFF
--- a/pico.toolchain
+++ b/pico.toolchain
@@ -1,0 +1,106 @@
+# todo there is probably a more "cmake" way of doing this going thru the standard path with our "PICO" platform
+#  i.e. CMake<Lang>Information and whatnot
+
+# Toolchain file is processed multiple times, however, it cannot access CMake cache on some runs.
+# We store the search path in an environment variable so that we can always access it.
+if (NOT "${PICO_TOOLCHAIN_PATH}" STREQUAL "")
+    set(ENV{PICO_TOOLCHAIN_PATH} "${PICO_TOOLCHAIN_PATH}")
+endif ()
+
+# Find the compiler executable and store its path in a cache entry ${compiler_path}.
+# If not found, issue a fatal message and stop processing. PICO_TOOLCHAIN_PATH can be provided from
+# commandline as additional search path.
+function(pico_find_compiler compiler_path compiler_exe)
+    # Search user provided path first.
+    find_program(
+            ${compiler_path} ${compiler_exe}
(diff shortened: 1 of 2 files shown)


ali1234:
How it works:

`32blit-pico/CMakeLists.txt` includes those two weird cmake files that set up the SDK. They claim that they must be included before `project()` however this does not seem to be the case because...

`pico.toolchain` is `pico_arm_gcc.cmake` and `find_compiler.cmake` from the pico-sdk. These normally have to be included before project... because they are actually toolchain files. Using them as such makes all this work. I have simply taken those two files and turned them into one file (ie pasting one into the other where it would normally import it). Then, at the end, I added a couple of defaults for the board and path variables.

pico.toolchain
```diff
@@ -0,0 +1,106 @@
...
+set(CMAKE_CXX_COMPILER_FORCED TRUE)
+
+# Add target system root to cmake find path.
+get_filename_component(PICO_COMPILER_DIR "${PICO_COMPILER_CC}" DIRECTORY)
+get_filename_component(CMAKE_FIND_ROOT_PATH "${PICO_COMPILER_DIR}" DIRECTORY)
+
+# Look for includes and libraries only in the target system prefix.
+set(CMAKE_FIND_ROOT_PATH_MODE_INCLUDE ONLY)
+set(CMAKE_FIND_ROOT_PATH_MODE_LIBRARY ONLY)
+#set(CMAKE_FIND_ROOT_PATH_MODE_PACKAGE ONLY)
```
ali1234, line 81: I also commented this line because it prevents the 32blit-sdk from being found. This is the same as the problem with emscripten.

Daft-Freak:
Hmm, hopefully the Pico SDK doesn't change in a way that breaks this... I guess this is more how CMake is supposed to be used, but also not how the Pico SDK wants you to use it :shrug: 


Anyway, having to do less special stuff for one platform is nice and this didn't appear to break the old setup.

ali1234:
I think the only reason this works for CI is because they have include guards on everything. So it will all get included multiple times if you do it the recommended way, but it is still in mostly the right order.

It probably will break, that's why this is a draft. Maybe we could add it as a submodule or use FetchContent to get a specific revision.

PR #715, opened 2021-09-15
Implement forced 16 bit screenmodes

ali1234:
You can set ScreenMode::lores_565 or ScreenMode::hires_565 to be
guaranteed a 16 bit framebuffer on every platform. On Pico, this
does exactly the same as the normal screenmodes.

### DIFF
--- a/32blit-pico/main.cpp
+++ b/32blit-pico/main.cpp
@@ -64,6 +64,7 @@ static volatile bool do_render = true;
 static Surface &set_screen_mode(ScreenMode mode) {
   switch(mode) {
     case ScreenMode::lores:
+    case ScreenMode::lores_565:
       screen = lores_screen;
       // window
 #ifdef DISPLAY_ST7789
@@ -75,6 +76,7 @@ static Surface &set_screen_mode(ScreenMode mode) {
       break;
 
     case ScreenMode::hires:
+    case ScreenMode::hires_565:
 #if defined(DISPLAY_ST7789) && ALLOW_HIRES
       if(have_vsync)
         do_render = true;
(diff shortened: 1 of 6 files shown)


ali1234:
Why: 565 screenmodes have better performance because they are word aligned. The whole pixel can be read or written in a single memory access.

This is a bit hacky though. For some reason the RGB565 mode is BGR in SDL. On Blit it is RGB.

I have converted a few if/else blocks into switch statements. And I duplicated the flip functions in the 32blit HAL. I assume those are supposed to be as fast as possible.

ali1234:
Also most of these files have a mixture of tabs and spaces, so I have not attempted to format the new code to match.

ali1234:
Also I haven't tested the lores modes.

Daft-Freak:
> Also most of these files have a mixture of tabs and spaces, so I have not attempted to format the new code to match.

I think most of the 32blit code is two spaces (and some of the code is definitely going from 2 -> 4 here)

Looks like there's a R/B swap bit for DMA2D if that helps with getting it consistent (`DMA2D_OPFCCR_RBS`)

Daft-Freak:
Okay yeah, I see what you mean about it being RGB on the blit. I think the blit display is getting R/B swapped _somewhere_ because the LTDC RGB888 mode is actually BGR... (Thought it was `st7272a_set_bgr`, but that's never called).

The important thing is that this all seems to work though.

ali1234:
Yes, on blit I was surprised that it "just works" where as on SDL I had to tell it the buffer is BGR, even though on SDL the 24 bit buffer is declared as RGB.

ali1234:
Replaced by https://github.com/32blit/32blit-sdk/pull/730

PR #718, opened 2021-10-04
Pico ST7789 refactoring

Daft-Freak:
Some of this is for supporting the other displays, but the rest results in more constants, more metadata and less hardcoded pins. (And also fixes it being a class that can only have one instance because reasons)


Gadgetoid:
Thank you! That all ended up as a bit of a hack.

PR #719, opened 2021-10-07
Docs: minor tweaks & a docs landing page

Gadgetoid:


Daft-Freak:
Hmm, think I saw a palette demo in PicoDVI...

PR #727, opened 2021-11-03
Reset sleep timer/wake up when receiving CDC data

Daft-Freak:
Random late late night "that'll be easy" thought... and it was as soon as I actually flashed the modified firmware... :facepalm:  :yawning_face:


Gadgetoid:
Thank you. I've had these random late night thoughts and they never end well :grimacing:

PR #728, opened 2021-11-06
Set 32BLIT_DIR in the Pico toolchain too

Daft-Freak:
For consistency with 32blit.toolchain. More importantly, this makes the command in the boilerplate/docs actually work.


Gadgetoid:
Ah I think this caught me out recently, too. Just added `-D32BLIT_DIR` on auto-pilot. Thank you!

PR #729, opened 2021-11-11
Introducing game controller rumble in SDL builds

LordEidi:
This is a first shot at having the game controllers which can rumble react to blit::vibration in the SDL port.

### DIFF
--- a/32blit-sdl/Input.cpp
+++ b/32blit-sdl/Input.cpp
@@ -68,6 +68,25 @@ int Input::find_button(int button) {
 }
 
 Input::Input(System *target) : target(target) {
+  // Open all joysticks as game controllers
+  for(int n = 0; n < SDL_NumJoysticks(); n++) {
+
+    if(SDL_IsGameController(n)) {
+
+      auto gc = SDL_GameControllerOpen(n);
+      if(gc != nullptr) {
+        _add_controller(gc);
+      }
+    }
+  }
+}
+
+Input::~Input() {
+  for (auto gc : game_controllers)
+  {
+    _remove_controller(gc.gc_id);
+  }
+  game_controllers.clear();
 }
 
 bool Input::handle_mouse(int button, bool state, int x, int y) {
@@ -132,6 +151,34 @@ bool Input::handle_controller_motion(int axis, int value) {
 	return false;
 }
 
+void Input::handle_controller_added(Sint32 joystick_index) {
+
+  SDL_GameController* gc = SDL_GameControllerOpen(joystick_index);
+  if (gc != nullptr) {
+    _add_controller(gc);
+  }
+}
+
+void Input::handle_controller_removed(Sint32 joystick_index) {
+
+  auto gc_removed = SDL_GameControllerFromInstanceID(joystick_index);
+  if(gc_removed != nullptr) {
+    _remove_controller(gc_removed);
+  }
+}
+
+void Input::rumble_controllers(const float& volume) {
+
+  if(volume > 0) {
+    for (auto gc : game_controllers) {
+      if(gc.can_rumble) {
+        auto frequency = volume * 0xFFFF;
+        SDL_GameControllerRumble(gc.gc_id, frequency, frequency, 50);
+      }
+    }
+  }
+}
+
 void Input::_virtual_tilt(int x, int y, int half_w, int half_h) {
 	float z = 80.0f;
 	x = x - half_w;
@@ -148,4 +195,22 @@ void Input::_virtual_analog(int x, int y, int half_w, int half_h) {
 	float jy = (float)y / half_h;
 	target->set_joystick(0, jx);
 	target->set_joystick(1, jy);
+}
+
+void Input::_add_controller(SDL_GameController* gc) {
+  // welcome rumble to test if it can rumble
+  auto can_rumble = SDL_GameControllerRumble(gc, 0xFFFF, 0xFFFF, 200);
+
+  GameController gcs = {gc, can_rumble == 0};
+  game_controllers.push_back(gcs);
+}
+
+void Input::_remove_controller(SDL_GameController* gc) {
+  game_controllers.erase(
+    std::remove_if(game_controllers.begin(), game_controllers.end(), [&](GameController const & controller) {
+        return controller.gc_id == gc;
+    }),
+    game_controllers.end());
+
+  SDL_GameControllerClose(gc);
 }
--- a/32blit-sdl/Input.hpp
+++ b/32blit-sdl/Input.hpp
@@ -1,5 +1,6 @@
 #include <map>
 #include <vector>
+#include <algorithm>
 class System;
 
 struct game_controller
@@ -13,9 +14,7 @@ class Input {
 		static std::map<int, int> keys;
 		static std::map<int, int> buttons;
 
-        std::vector<game_controller> game_controllers;
-
-        static int find_key(int key);
+    static int find_key(int key);
 		static int find_button(int button);
 
 		Input(System *system);
--- a/32blit-sdl/Input.hpp
+++ b/32blit-sdl/Input.hpp
@@ -1,12 +1,21 @@
 #include <map>
+#include <vector>
 class System;
+
+struct game_controller
+    {
+        SDL_GameController* gc_id;
+        bool can_rumble;
+    };
 
 class Input {
 	public:
 		static std::map<int, int> keys;
 		static std::map<int, int> buttons;
 
-		static int find_key(int key);
+        std::vector<game_controller> game_controllers;
+
+        static int find_key(int key);
 		static int find_button(int button);
 
 		Input(System *system);
--- a/32blit-sdl/Input.hpp
+++ b/32blit-sdl/Input.hpp
@@ -3,11 +3,11 @@
 #include <algorithm>
 class System;
 
-struct game_controller
-    {
-        SDL_GameController* gc_id;
-        bool can_rumble;
-    };
+struct GameController
+{
+  SDL_GameController* gc_id;
+  bool can_rumble;
+};
 
 class Input {
 	public:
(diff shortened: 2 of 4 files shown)


LordEidi:
There are a few points which probably will need some discussion:

- Currently there is no update in-game. If a controller is added or removed during the lifetime of the application, the corresponding SDL events won't be handled. But I will add this once we agreed on the rest.
- There is a new struct which holds the pointer to the game controller, as well as a boolean which stores the info re rumble ability of the controller. There is no need to regularly call rumble if the controller is not able to do so.
- I had to split the initialisation code in 32blit-sdl/Main.cpp around line 241 since I put the controller storage into 32blit-sdl/Input.hpp. While it would not necessarily be needed to first initialise the Input class it probably would make sense to move the game controller registration (and de-reg) code into the Input class so that we could call the same code from within the SDL Event handler.
- There is still a bit of fprint sprayed through the code. That can be removed later on and was mostly used for debugging purposes.
- The duration of the rumble is always 20ms. Mostly because the SDL API and the 32blit API are not compatible. The one asks for a duration and the other just has a value for the rumble frequency. Telling the SDL API to rumble for 20ms with the frequency written into the blit::vibration solves the problem somewhat, that we call the rumble function for as long as the blit::vibration is set without the need to have some complex logic to sync the SDL duration with the 32blit loops.

Daft-Freak:
Yeah, probably makes sense to put all the controller setup in the `Input` class. In the SDL code all the duration does is set the rumble to zero at a later point using the exact same API exposed to users, so you can probably just set the duration to something longer than the update period and pretend it doesn't exist.

Also, SDL just got a HasRumble query... https://github.com/libsdl-org/SDL/pull/4943 (Not that we can use it yet)

... and no PR would be complete without me pointing out that the indentation is looking a bit wonky :smile:

32blit-sdl/Main.cpp
```diff
@@ -233,13 +233,39 @@ int main(int argc, char *argv[]) {
...
 	// Open all joysticks as game controllers
-	for(int n=0; n<SDL_NumJoysticks(); n++) {
-		SDL_GameControllerOpen(n);
+	for(int n = 0; n < SDL_NumJoysticks(); n++) {
+
+        if(SDL_IsGameController(n)) {
+
+            SDL_GameController* gc = SDL_GameControllerOpen(n);
+
+            if (gc != nullptr && SDL_GameControllerGetAttached(gc) == SDL_TRUE) {
```
Daft-Freak, line 248: I'm not sure open can be successful if the controller isn't connected?

32blit-sdl/Input.hpp
```diff
@@ -1,12 +1,21 @@
 #include <map>
+#include <vector>
 class System;
 
+struct game_controller
```
Daft-Freak, line 5: Should be CamelCase I think. (Maybe even private to `Input`)

32blit-sdl/Main.cpp
```diff
@@ -233,13 +233,39 @@ int main(int argc, char *argv[]) {
...
+                printf("SDL_GetError() = %s\n", SDL_GetError());
+            }
+        }
 	}
 
-	blit_system = new System();
-	blit_input = new Input(blit_system);
+    if(!blit_input->game_controllers.empty()) {
+
+        SDL_GameControllerEventState(SDL_ENABLE);
```
Daft-Freak, line 265: Hmm, enabled should be the default.

32blit-sdl/Main.cpp
```diff
@@ -270,6 +296,13 @@ int main(int argc, char *argv[]) {
 	delete blit_capture;
 #endif
 
+    // delete game controllers
+    for (auto gc : blit_input->game_controllers)
+    {
+        SDL_GameControllerClose(gc.gc_id);
+    }
+    blit_input->game_controllers.clear();
+
```
Daft-Freak, line 305: I'd say that should be in `Input`'s destructor... but it's currently not getting `delete`d. Something else to fix I guess...

32blit-sdl/System.cpp
```diff
@@ -269,6 +272,16 @@ void System::loop()
 	blit::joystick.y = shadow_joystick[1];
 	SDL_UnlockMutex(m_input);
 	blit::tick(::now());
+
+    if(blit::vibration > 0) {
```
Daft-Freak, line 276: Doesn't this mean you can't turn vibration off? Though the duration would make it auto-disable anyway... This block could be an `update_rumble` method on `Input` or something similar.

LordEidi: Nopes, no need to. The code only rumbles for 20ms max anyway. So it only rumbles until the next cycle. And then either is restarted for another 20ms or silenced automatically.

32blit-sdl/Main.cpp
LordEidi: Funnily, this is taken from some SDL Lib example. But I get your point, the second part of the if could probably be omited.

32blit-sdl/Input.hpp
LordEidi: Thought about making it private as well. But since not all functions are within the Input class, it currently needs to be public.

32blit-sdl/Main.cpp
LordEidi: Taken from an example as well. Not sure it always is. Documentation is not 100% clear on that. Would leave this as it is.

32blit-sdl/Main.cpp
LordEidi: Indeed. Leaving there until fixed, or moving to the Input class and calling it from here?

Daft-Freak: I'd just stick a delete with the other three below, fixing the (tiny) leak too. (Not that it matters much as we're exiting anyway...)

LordEidi:
> Yeah, probably makes sense to put all the controller setup in the `Input` class. In the SDL code all the duration does is set the rumble to zero at a later point using the exact same API exposed to users, so you can probably just set the duration to something longer than the update period and pretend it doesn't exist.

re Input class. OK, I am moving the code into the Input class.

re duration. Not sure I get what you mean. But that is what is happening now. Rumble is set for 20ms which is one cycle. And is then restarted for another 20ms if still asked for by the game. No need to stop, you won't have any chance to cancel those 20ms anyway.

> Also, SDL just got a HasRumble query... [libsdl-org/SDL#4943](https://github.com/libsdl-org/SDL/pull/4943) (Not that we can use it yet)

Thanks for the hint. The documentation regarding controllers is a bit difficult btw.

> ... and no PR would be complete without me pointing out that the indentation is looking a bit wonky smile

Must be the problem with the diff between me writing and me moving the changes into another repository... Looking into it, but I hope you can life with it anyway.

Daft-Freak:
For the duration I was mostly pointing out that setting the rumble again before the duration is finished is okay, because SDL itself does that to stop after the duration is finished. (It might be a good idea to set a duration slightly higher than the update interval, since there's no guarantee that the 20ms timer is _actually_ 20ms.)

For the indentation, it's apparently _supposed_ to be 2 spaces (https://github.com/32blit/32blit-sdk/blob/master/.editorconfig#L11-L12). But the SDL code is mostly tabs... Though the new code here is neither of those.

32blit-sdl/Main.cpp
Daft-Freak: I'm pretty sure that there are only a few SDL events that are disabled by default, and they're not controller events... also it worked without that before.

32blit-sdl/Input.hpp
LordEidi: And re CamelCase, nopes, see File.hpp

LordEidi:
What should I take as duration? 50ms? Not noticeable by the user, but long enough for the cycle duration?

32blit-sdl/Input.hpp
Daft-Freak: The struct in that file is CamelCase though? I know there's `packed_image`, but I think that was missed back when everything was changed to CamelCase...

Daft-Freak:
Something like that yeah, enough to cover any timing variations but doesn't need to be too big.

LordEidi:
Hopefully I addressed all points:

- changed 20ms to 50ms.
- moved all controller handling into the Input class.
- added a destructor to Input and calling delete from Main.
- added correct event handling for removed and added controllers.
- configured 1 tab to 2 spaces.
- renamed struct to CamelCase.
- removed unwanted / not needed code.

Daft-Freak:
Well, it works... when you use a controller that actually has rumble :facepalm:  (turns out most of the controllers I have lying around don't...)

Don't see anything non-trivial to complain about so I'll just ignore any little formatting things and leave the rest to someone who can press the merge button :smile:

LordEidi:
Thanks @Daft-Freak for your support!

32blit-sdl/Input.cpp
```diff
@@ -149,3 +196,21 @@ void Input::_virtual_analog(int x, int y, int half_w, int half_h) {
 	target->set_joystick(0, jx);
 	target->set_joystick(1, jy);
 }
+
+void Input::_add_controller(SDL_GameController* gc) {
+  // welcome rumble to test if it can rumble
+  auto can_rumble = SDL_GameControllerRumble(gc, 0xFFFF, 0xFFFF, 200);
```
cgutman, line 202: If you don't actually want the welcome rumble, you can probe for rumble functionality using short low-intensity rumble like `SDL_GameControllerRumble(gc, 1, 1, 1)` for versions of SDL that lack `SDL_GameControllerHasRumble()`. It's so short that the rumble motors won't even have time to spin up.

Gadgetoid:
What's the status on this? I'd love to merge it for the next release.

LordEidi:
> What's the status on this? I'd love to merge it for the next release.

You should be good to go. Just added the changes to the 32blit dev version yesterday.

Although: There is a "welcome rumble" which shortly rumbles the controller when being initialised. You should reduce the duration to 1 if you do not want that to be noticed by the users. I find it quite useful since you know things work because of the welcome greeting. But YMMV.

Gadgetoid:
> I find it quite useful since you know things work because of the welcome greeting.

It does seem like a helpful debugging thing. I'd probably add a command-line switch to enable/disable it in future.

Gadgetoid:
And... merged! Thank you all involved.

PR #730, opened 2021-11-14
Screen mode cleanup and extended API

Daft-Freak:
I thought about it too much and had to write it... cleanup based on an old patch, 565 bits loosely based on the other PR.

The main thing this adds is: `bool set_screen_mode(ScreenMode new_mode, PixelFormat format)`, which enables:
 - Paletted `lores` mode on non-pico (doesn't reconfigure LTDC like `hires`, just changes the input format to the doubling magic)
 - RGB565 modes on non-pico
 - Returns false instead of just ignoring you on non-supported modes/formats (so everything except 565 on pico)
 - The API function this uses could easily be extended to allow things like passing in a framebuffer pointer

This does increase the API minor version, so games built with this won't run on older firmware (but old games will be fine).

[Updated screen mode test](https://github.com/32blit/32blit-sdk/files/7534853/screen-mode.zip). Cycles through all six mode/format combinations. A/B for manual mode change.

(Adding extra formats to the STM32 HAL is reasonably easy if for some reason someone wanted ARGB4444... or 1555... or even 4bit paletted)

### DIFF
--- a/32blit-pico/main.cpp
+++ b/32blit-pico/main.cpp
@@ -96,6 +96,42 @@ static void set_screen_palette(const Pen *colours, int num_cols) {
 
 }
 
+static bool set_screen_mode_format(ScreenMode new_mode, SurfaceTemplate &new_surf_template) {
+  new_surf_template.data = screen_fb;
+
+  switch(new_mode) {
+    case ScreenMode::lores:
+      new_surf_template.bounds = lores_screen.bounds;
+      break;
+    case ScreenMode::hires:
+    case ScreenMode::hires_palette:
+#if defined(DISPLAY_ST7789) && ALLOW_HIRES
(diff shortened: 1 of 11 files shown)


Gadgetoid:
Thanks for putting this together- I have been buried in HUB75 LED matrix stuff for a while but I should get some time for 32blit soon. Looks like things are starting to bubble away here post retail launch.

ARGB4444 parity between 32blit and PicoSystem would be interesting. Another step toward absorbing games built with the PicoSystem SDK :laughing:

Gadgetoid:
Thank you- merging this because it's sat around for far too long and it'd be handy to have available for people to start tinkering with even if I don't immediately tag a release.

Of course it may cause havoc if there's no firmware build to support the new and shiny :grimacing: but I don't want to bottleneck things too much.

PR #731, opened 2021-11-24
Improve feedback for trying to use storage mode with no SD card

Daft-Freak:
Refuse to enter storage mode with "No SD Card" instead of getting stuck at "initialising".

(Seems unmounting on Ubuntu is broken again though)


Gadgetoid:
Thank you!

PR #732, opened 2021-11-25
Start documenting the firmware/launcher/PIC magic

Daft-Freak:
Since the 32blit firmware seems to be where I live now. Not being the only one who understands most of this is probably helpful. :smile: 

Expanded a bit from recent Discord discussion. Probably still missing some details.


Gadgetoid:
Oooh, this is beautiful, thank you. Might encourage some poor soul to try and replicate it on PicoSystem!

PR #737, opened 2021-12-07
ITCM for "user" code

Daft-Freak:
Allows use of ITCM in things that aren't the firmware (which doesn't use any of it).

- Adds the init code to `user_startup`
- Splits ITCM roughly in half (because of the `API`/`ITCMISR` sections the game gets 30k and the firmware gets 4+28+2k)
- Adds some macros for marking "fast"/in-memory functions (similar to the `not_in_flash` stuff from the Pico SDK... also uses that for pico)

Slightly silly minimal test:
```c
void blit_no_inline_fast_code(itcm_print)() {
  debugf("Hello from ITCM!\n");
}
```
(Covers most of the cases that used to fail before the tools stuff)

Something more impressive:
```diff
diff --git a/lua/lvm.c b/lua/lvm.c
index aa3b22b..2a97521 100644
--- a/lua/lvm.c
+++ b/lua/lvm.c
@@ -1118,8 +1118,8 @@ void luaV_finishOp (lua_State *L) {
 #define vmcase(l)	case l:
 #define vmbreak		break
 
-
-void luaV_execute (lua_State *L, CallInfo *ci) {
+#include "engine/fast_code.hpp"
+void blit_fast_code(luaV_execute) (lua_State *L, CallInfo *ci) {
   LClosure *cl;
   TValue *k;
   StkId base;

```


Gadgetoid:
The Lua tweaks built, but I need to look into why on earth they didn't *run*. PEBKAC error, probably!

Either way this is an immensely exciting change that - based on your Lua build - could really make Lua and some other things viable.

Thank you!

PR #739, opened 2021-12-07
Delete things

Daft-Freak:
Just removing some old/broken stuff I saw while looking for other things...

(OpenOCD stuff in tools is also probably outdated...)


Gadgetoid:
Doing the good Lord's work there. Thank you.

PR #740, opened 2021-12-20
Prepend target name to generated metadata.cpp

Daft-Freak:
Avoids a CMake error when you have multiple targets with metadata in the same CMakeLists/subdir.


Gadgetoid:
As always I wonder what edge case you ran into to find this :laughing: 

Thank you!

PR #741, opened 2021-12-26
Fix mipmap boundary check

Blackhawk-TA:
mipmap_index is an unsigned integer so checking it for lesser than 0 has no effect


Gadgetoid:
Thank you!

PR #743, opened 2021-12-27
Use asset packer for voxel map data

Daft-Freak:
Removes a 12MB header file. (Okay, replaces it with a generated 13MB _source_ file...) I only noticed this file when my Pi zero 2 ran out of memory trying to compile the example.

This example is probably a good... example of something that doesn't really belong in the SDK... (It's huge and doesn't use any unique APIs...)


Gadgetoid:
It's a nice demo to have in the releases, but otherwise it doesn't seem to serve a purpose. You're right.

I'm definitely all for stripping the games out of here and replacing them with unit tests- I think I've mentioned that somewhere before.

Releases could very easily grab a selection of projects hosted in their own GitHub repos, build them, and add them into the shipped .zip. I think we should aim to move in this direction. Good coverage of the SDK would be a valuable bonus.

For now, though, this is at least a move in the right direction.

Thank you!

PR #746, opened 2022-01-07
Fix load_tmx flags typo

Daft-Freak:
These were supposed to be different flags. Probably unnoticed as the default is both and I use neither in blit kart.


Gadgetoid:
That's quite the oopsie! Thank you.

PR #747, opened 2022-01-10
Some example changes

Daft-Freak:
Delete racer (there wasn't any code anyway), move `platformer.tmx` (it isn't the map for platformer) and add something resembling the real one.


Gadgetoid:
Thank you! I really need to get away from <OTHER THINGS> so I can release this and catch up!

PR #748, opened 2022-01-10
Firmware tidying

Daft-Freak:
Ended up looking at the firmware code for something and spotted a bunch of things that annoyed me enough to tidy them up a bit...

- Some deleting of unused/duplicated code
- Moves ADC handling out of 32blit.cpp (maybe should've merged the reset of the HAL setup in there too...)
- Moves the last i2c device init to i2c.cpp and renames things for consistency
- moves some of the newer files around a bit
- Only updates the charge LED when charge status changes (there was a TODO about this, also haven't tested this)

There's definitely more stuff that could be tidied up here... (the old execution switching code is probably entirely broken due to being untested for over a year, display SPI code is only used by a function that is commented out)


Gadgetoid:
uh oh
image

this... seems to have exploded into a state of merged/unmerged limbo.

PR #749, opened 2022-01-25
SDL file path fixes

Daft-Freak:
Removes an Emscripten workaround (I fixed it in SDL and we're already using a version with the fix) and makes save paths work everywhere instead of just `open_file`.

Which would be really handy if someone wanted to add a `remove_save` wrapper :smile:


Gadgetoid:
Thank you! Looks like we're long overdue a release.

PR #750, opened 2022-01-26
Add remove_save method to delete existing save data

Blackhawk-TA:
Currently it is not possible to delete existing save files, therefore I added this functionality to the api.

Waiting for https://github.com/32blit/32blit-sdk/pull/749 to be merged.


32blit-sdl/File.cpp
```diff
@@ -182,7 +182,13 @@ bool rename_file(const std::string &old_name, const std::string &new_name) {
 }
 
 bool remove_file(const std::string &path) {
-  return remove((basePath + path).c_str()) == 0;
+  // check if the path is under the save path and do not add basePath if this is the case
```
Gadgetoid, line 185: This should be unnecessary, since the generic "remove_file" probably still wants to be able to delete things from / (firmware/launcher updates for example)

Gadgetoid:
Should be no need to change `remove_file` since that should be generic and not know/care about the save directory.

We have no pretense of filesystem safety here :laughing:

PR #752, opened 2022-02-11
Adjust SDL loop timing a bit

Daft-Freak:
This fixes the fact that SDL only called `tick` every 20ms, so you would always be running behind and get two updates. Also switches `render`/`tick` around as SDL was backwards compared to STM32/Pico.

### DIFF
--- a/32blit-sdl/System.cpp
+++ b/32blit-sdl/System.cpp
@@ -306,9 +306,13 @@ void System::loop() {
   blit::joystick.y = shadow_joystick[1];
   SDL_UnlockMutex(m_input);
 
-  // only render at 50Hz
+  // only render at 50Hz (main loop runs at 100Hz)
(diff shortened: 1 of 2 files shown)


Daft-Freak:
... and then I realise that this will have unstable `render` timing in Emscripten as `loop` gets called at [whatever your refresh rate is]... which probably isn't 100Hz... hmm...


(Which also means it was already running `render` at a different rate to everything else, and was also synced unlike the non-Emscripten SDL code)

Daft-Freak:
Came back to this as it helps the audio buffering in DaftBoy.

"Fixed" Emscripten by #ifdef-ing the render check. It's still the odd one out, running at the display refresh rate unlike SDL on anything else (the same as before this change)...

Gadgetoid:
Thank you!

PR #753, opened 2022-02-12
pico: Fix default screen mode

Daft-Freak:
Someone needs to setup `blit::screen`... Unbreaks everything that uses the "default" mode, which includes a lot of the examples :facepalm: 

(Err, whoops)

### DIFF
--- a/32blit-pico/main.cpp
+++ b/32blit-pico/main.cpp
@@ -359,7 +359,7 @@ int main() {
   init_fs();
   init_usb();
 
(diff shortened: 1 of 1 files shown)


Gadgetoid:
Eek! Thank you. New release needed, I guess!

Daft-Freak:
The worst part was that I'd already fixed the _exact same thing_ in the SDL code in the PR.

There were quite a few combinations to test there though, so only missing one was pretty good.

PR #756, opened 2022-02-17
pico: Delcare screen_fb as uint16_t

### LINKED ISSUE #755: Crash on Picosystem with MinSizeRel only
The following code crashes only on Picosystem and only when built with MinSizeRel:

```C++
#include "32blit.hpp"

class StageScores {
    public:
        StageScores() {
            if (!scores_loaded) {
                scores_loaded = true; // <- uncomment this line to break it
            }
        }
    private:
        static inline bool scores_loaded = false;
};

static StageScores *stage = nullptr;

void init() {
    blit::set_screen_mode(blit::ScreenMode::hires, blit::PixelFormat::RGB565);
    stage = new StageScores();
}

void update(uint32_t /*time*/) {

}

void render(uint32_t  time) {
    blit::screen.pen = blit::Pen((time>>2) & 0xff, 20, 20);
    blit::screen.clear();
}
```

To reproduce:

1. Clone 32blit-boilerplate
2. Replace game.cpp with the above code
3. Build with: `cmake -DCMAKE_BUILD_TYPE=MinSizeRel -DCMAKE_TOOLCHAIN_FILE=../../32blit-sdk/pico.toolchain ..`
4. Flash to Picosystem

If the code is working, the screen should flash red. If the screen is blank then the code has crashed.

As the comment says, writing to the `static inline bool` seems to be the cause of the crash.

I can't reproduce this with the Picosystem SDK or with just the Pico SDK.

Daft-Freak:
Everything that isn't blit::Surface wants that and we assume 2-byte alignment in the blending code. 

Seems to fix  #755, which was hardfaulting in the blending code.

### DIFF
--- a/32blit-pico/main.cpp
+++ b/32blit-pico/main.cpp
@@ -33,18 +33,18 @@ static SurfaceInfo cur_surf_info;
 static const int lores_page_size = (ST7789_WIDTH / 2) * ((ST7789_HEIGHT + 1) / 2) * 2;
 
 #if ALLOW_HIRES
-uint8_t screen_fb[ST7789_WIDTH * ST7789_HEIGHT * 2];
+uint16_t screen_fb[ST7789_WIDTH * ST7789_HEIGHT];
 #else
-uint8_t screen_fb[lores_page_size * 2]; // double-buffered
+uint16_t screen_fb[lores_page_size]; // double-buffered
 #endif
 static bool have_vsync = false;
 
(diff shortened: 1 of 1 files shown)


ali1234:
This fixes it for me. Is declaring the framebuffer uint16 guaranteed to align it or should we take further measures eg [alignas](https://en.cppreference.com/w/cpp/language/alignas)?

Daft-Freak:
I think it should be enough, any attempt to access an unaligned uint16 causes a fault... so I'd hope the compiler is smart enough to not do that.

Though forcing it to be aligned to 4 bytes may be (slightly) more optimal...

Gadgetoid:
Thank you all! This actually reads better, too, albeit the number of uint8 casts seems to have balanced it out :laughing:

PR #757, opened 2022-02-17
Avoid hardcoding the screen size for SDL everywhere

Daft-Freak:
Allows changing the screen size by altering one place (in System.hpp).


Gadgetoid:
Yikes! Makes sense, thank you!

PR #758, opened 2022-02-21
Misc pico docs fixes

Daft-Freak:
Just some small tweaks

- ~half of ITCMRAM is reserved (though none of it was usable before)
- vgaboard has a bit less RAM available
- HID code for vgaboard isn't merged, but does support the joystick

(There's also unmerged SD card support for vgaboard, but it's a huge pile of code...)

I should probably finish some of the "make VGA actually usable" bits.


Gadgetoid:
Thank you!

PR #760, opened 2022-03-07
Fix exit(0) crashing

Daft-Freak:
`exit` is marked as not returning, bad things happen if it does. This worked fine until #661 made it so that `blit_exit` didn't reset when there wasn't an error.

This is the easiest fix of always resetting in `blit_exit`... which isn't as clean an exit as you get from using the menu. While typing this patch I thought of an alternative, which is slightly more magic: https://github.com/Daft-Freak/32blit-beta/commit/645a3a504e8c92328e9811363937cfbdc5ae7758


Gadgetoid:
Ooh, nice, thank you!

The magic fix looks intriguing though if my experience at being terrible at writing drivers for MicroPython is anything to go by, it can't hurt to have a real system reset every now and then.

PR #761, opened 2022-03-10
Fix DIR leak in blit::list_files

Daft-Freak:
`blit::list_files` has been leaking a whole 56 bytes... for the last two years... This fixes it by just putting the `DIR` object on the stack instead.

(I wasn't even doing blit stuff, just came to uh, "borrow" some of the fatfs glue)


Daft-Freak:
Hmm, nothing interesting, at least not yet :smile:

PR #762, opened 2022-03-11
Clean up the Emscripten SDL flags space hacks

Daft-Freak:
... by not using any spaces at all. I didn't know that `-s` without the space was a thing until recently, so now I'm cleaning this up everywhere...

(With new enough Emscripten you don't even need a special case for SDL2 itself as `find_package(SDL2)` just works... doesn't apply to _image/_net though)


Daft-Freak:
Screenshot 2022-10-13 at 17-21-42 32blit Games

I uh, think this may be more than a cleanup for newer Emscripten versions.

(I was about to try and fix those... except they're already fixed by this. The script is just using the last release... )

PR #763, opened 2022-03-17
Adjust SDL DLL copying to handle VS generators

Daft-Freak:
There's an extra subdir for the build type. This also moves the "dev" copy to post-build instead of configure.


Gadgetoid:
Thank you!

PR #765, opened 2022-03-21
Actions tweaks

Daft-Freak:
Copies the bit from the boilerplate that avoids an extra run on releases and tweaks the release action to also work on drafts. (Also in https://github.com/32blit/32blit-boilerplate/pull/24)


Gadgetoid:
Wow the "published" and "created" for releases seem intuitively the wrong way around :laughing: 

Thank you!

PR #767, opened 2022-03-28
Add optional support for filtering in list_files

Daft-Freak:
Mostly to avoid DaftBoy running out of memory while listing the root of my SD card... but seems useful for the launcher too.

### DIFF
--- a/32blit/engine/file.cpp
+++ b/32blit/engine/file.cpp
@@ -29,10 +29,11 @@ namespace blit {
    *
    * \return Vector of files/directories
    */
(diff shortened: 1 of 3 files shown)


Gadgetoid:
Nice! Thank you.

Daft-Freak:
Uh oh, may have made a mistake here
https://github.com/32blit/32blit-sdk/commit/7c1c4491d9b59a42e1472f8ea412c99385b63402#annotation_4237589158

(Dropped the "no extension" check I think...)

PR #768, opened 2022-06-01
pico: scanvideo improvements

Daft-Freak:
Various scanvideo changes (some quite old):

- Moves scanline updates to core1 (more reliable timing)
- Fixes sync
- Partial `hires` support (can only have one mode, defaults to `lores`, `ALLOW_HIRES=1` switches to `hires`)

Last commit is a bit of a hack, but I used it to port DaftBoy. VGA still isn't very useful without the input stuff...


Gadgetoid:
Thank you!

nom nom nom merge all the things!

PR #769, opened 2022-06-01
pico: HID input

Daft-Freak:
(Opening this so I don't keep forgetting about it)

USB HID based input for VGA. Draft because I only tested it on three controllers (one of them doesn't work for some reason). Could use some testing on some "common" gamepads (but not xbox stuff as they don't do HID... *sigh*)

What does work is flashing [this uf2 to a PicoSystem](https://github.com/32blit/32blit-sdk/files/8813671/picosystem-gamepad.zip)

Edit: Patch to hardware-test for testing https://github.com/Daft-Freak/32blit-beta/commit/48650294b5a65d4208b9f50ed4e2a0ccc70a4047

### DIFF
--- a/32blit-pico/CMakeLists.txt
+++ b/32blit-pico/CMakeLists.txt
@@ -101,6 +101,10 @@ elseif(BLIT_DISPLAY_DRIVER STREQUAL "st7789")
     list(APPEND BLIT_BOARD_DEFINITIONS DISPLAY_ST7789) # config defaults use this, also some games are using it for picosystem detection
 endif()
 
+if(BLIT_INPUT_DRIVER STREQUAL "usb_hid")
+    list(APPEND BLIT_BOARD_DEFINITIONS INPUT_USB_HID)
+endif()
+
 if(BLIT_USB_DRIVER STREQUAL "host")
     list(APPEND BLIT_BOARD_DEFINITIONS USB_HOST)
     list(APPEND BLIT_BOARD_LIBRARIES tinyusb_host)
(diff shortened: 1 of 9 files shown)


Daft-Freak:
Maybe when I find some gamepads that behave. :laughing: 

(VGA/DVI board + OTG adapter, done!)

Daft-Freak:
Seems more reliable with the new pico-sdk. Also added keyboard support.

Gadgetoid:
OOoh, that's good news!

Daft-Freak:
Still the same two devices working, but I can now switch them (+ a keyboard) around repeatedly without the pico crashing/resetting. 

Also, hubs and multiplayer fun: https://fosstodon.org/@Daft_Freak/109926331926726469

Daft-Freak:
[Slightly modified hardware-test](https://github.com/32blit/32blit-sdk/files/10888976/hardware-test.zip) (displays raw button info at the bottom)

Daft-Freak:
Considering this doesn't affect PicoSystem at all, is there any reason I _shouldn't_ just merge this?

(I've got some USB/multiplayer bits sort-of blocked by this...)

ahnlak:
I can't think of a particularly strong reason *not* to (although, does that imply that you could in theory plug a keyboard into a PicoSystem?)

Gadgetoid:
32blit support for PicoSystem is "unofficial" so I'm okay with it implying things we don't actually support,

This feature may benefit certain other Pico-based projects in the pipeline :laughing:

Daft-Freak:
If you change the board config I guess, but you'd lose the buttons... (USB role is a build-time config here, also you only get one input "driver")

Daft-Freak:
> 32blit support for PicoSystem is "unofficial" so I'm okay with it implying things we don't actually support,

Hmm, yes _someone_ should stop adding support for random things... :laughing: But uh, what does "official support" even mean now?

Daft-Freak:
Now to wait for a "my device doesn't work"/"the buttons are mapped wrong"!

mikerr:
Just got a pimoroni VGAdemo board, 
and tried this out with -DPICO_BOARD=vgaboard 

Yay, my device DOES work ( Logitech F710) and buttons mapped ok using D mode👌

PR #771, opened 2022-06-02
pico: "Fix" PWM audio clock div

Daft-Freak:
Nothing uses this by default, but I've had the patch lying around for months...


Gadgetoid:
Makes sense, just in case anyone is using this code for reference - or we ship a PicoSystem with less garbage audio :eyes: - thank you!

PR #772, opened 2022-06-02
Pico examples changes

Daft-Freak:
This disables the tilt example (doesn't do anything interesting) and sets `ALLOW_HIRES=0` for platformer. Both of these would fail to build with a 320x240 display.


Gadgetoid:
Thank you!

PR #773, opened 2022-06-11
stm32: work around QSPI failing to read the last 32 bytes

Daft-Freak:
Example code:
```c
*(volatile uint8_t *)0x91FFFFE0; // boom, system gone
```

Locks up _everything_ including debugging...

Guess who spent most of the day debugging this :neutral_face:. (Not that I need those bytes, it was just a really weird crash to hit)

More info: https://www.st.com/resource/en/errata_sheet/es0396-stm32h750xb-and-stm32h753xi-device-limitations-stmicroelectronics.pdf#%5B%7B%22num%22%3A90%2C%22gen%22%3A0%7D%2C%7B%22name%22%3A%22XYZ%22%7D%2C67%2C546%2Cnull%5D


Gadgetoid:
How in the world did you even stumble upon this? :laughing: 

Thank you!

Daft-Freak:
Think I was trying to copy the entire flash over USB... or something like that.

PR #774, opened 2022-06-13
Update USB device library to v2.6.0

Daft-Freak:
I _think_ this is the version after the one currently in use. Most important thing from [the release notes](https://htmlpreview.github.io/?https://github.com/STMicroelectronics/stm32_mw_usb_device/blob/master/Release_Notes.html) is "Fix multiple error with SCSI commands handling" (It was _entirely_ broken).

Some things were already fixed in our second copy of `usbd_msc_scsi.c`, but it wasn't enough and almost any error would make the host think the device was hung.

Had to modify the lib a little as they un-consted some data for some reason. Hopefully this hasn't broken something else...


Gadgetoid:
Looks like the Emscripten fail was some temporary build failure, but I can't for the life of me seem to re-run the job! Could you give it a kick, please?

Daft-Freak:
Kick successful :smile:

Gadgetoid:
Thank you!

PR #775, opened 2022-06-13
Rescan flash after flashing from SD

Daft-Freak:
Found another case of not updating the game list... could result in duplicate games or not flashing new versions.


Gadgetoid:
Thank you.

I feel a release coming on...

Daft-Freak:
Hah, just a few patches here... (I might have more, think I have a branch that deletes a bunch of code somewhere...)

PR #777, opened 2022-08-08
Fix unused variable in launcher

Daft-Freak:
*Immediately adds a warning back after cleaning most of them up* :facepalm: :laughing: 

The reduced warning noise made this easy to spot. Thought this was a bug, but the original user of this just got moved elsewhere.


Gadgetoid:
Thank you!

PR #780, opened 2022-08-27
Flight demo improvements

ThePythonator:
### _**This PR depends on (but could replace) #779.**_

On top of #779, I have added several improvements to the flight demo game:
* Trees are scaled so they get bigger as you get closer (rather than staying the same size, which makes it seem like they are shrinking).
* Water images are now tiled across the whole screen (below the horizon), rather than one square on the left of the screen.
* Joystick and D-pad both turn the player at the same speed (before, the joystick was very slow and the D-pad was very fast).
* Turning speed is clamped so that you can't turn faster by using the joystick and D-pad at the same time.


Gadgetoid:
Looks good to me! Nice sleuthing (and fixing!) poor example has been neglected for so, so very long.

Would you like me to merge #779 so you can rebase this PR?

Daft-Freak:
I have a feeling I also have some changes to this example somewhere, but I _think_ it's completely different stuff to this.

(Definitely approve of making the trees scale, made sure that worked in SBK's mode7 code :smile:)

ThePythonator:
> Would you like me to merge #779 so you can rebase this PR?

It's probably easiest just to merge this and close the other, since _I think_ this has all the changes of #779 already (by accident, because my git knowledge is limited to pull/push/clone/commit).

Daft-Freak:
I think #779 wants to be rebased or squash-merged to remove the self-merge in there... The commits from there will probably disappear here after that one is merged.

(I almost said I'd rebase it myself, forgetting that I don't have write access here, whoops. But that also means I probably don't get to say how things should be merged either :laughing: )

Gadgetoid:
Ah! I'd thought the self merge was here. But yes, you're right. Would be ideal to get those squashed out.

ThePythonator:
I'm not entirely sure how to fix the mess in #779, since when I've tried doing a squash merge before, I've ended up with a ton of seemingly blank commits (probably because I pushed before doing the squash merge). Would rebasing work, despite the fact that I've already pushed it?

ThePythonator:
I did something, not sure if it was the right thing, but there's certainly a lot less commits now!

PR #782, opened 2022-09-05
Fix delayed game switch

Daft-Freak:
The previous check was unreliable, it's still possible we were called from user code when it's disabled (launch from SD will disable user code to flash before calling this).

Replace the check with a bool set around calling tick, which should always be accurate. (Unless someone attempts to launch from `render`...)

This reliably crashed my blithub downloader, but doesn't seem to break the launcher even though it should...


Gadgetoid:
Thank you!

PR #783, opened 2022-09-09
Update to latest SDL for macOS/Windows builds

Daft-Freak:
This started as an attempt to work around build failures on macOS, then turned into "I wonder if I can clean this up by updating"... and then resulted in [android boilerplate for some reason](https://github.com/32blit/32blit-boilerplate/compare/main...Daft-Freak:32blit-boilerplate:android).

First three are fixes and could be split if we just want fixes for now. Then, macOS/Windows (VS+MinGW) SDL versions are updated which allows some simplification as they both support `find_package` now. 

There's some cleanup to make macOS frameworks slightly less special, the VS script doesn't create two copies of the headers/libs and the MinGW script no longer needs to patch the config.

2c89fb106a779b80e258e75108b3433bf379d609 and 96a175fa266e25e1ca613bd2ee1ca79fa443a599 allow [including SDL sources before the SDK](https://github.com/Daft-Freak/32blit-boilerplate/blob/android/android-project/app/jni/CMakeLists.txt), which is handy for platforms that don't have a prebuilt version available.

Oh, and I actually updated the docs this time. (Fixing the SDL download links as well)

(Probably replaces #483)

### DIFF
--- a/.github/workflows/cmake.yml
+++ b/.github/workflows/cmake.yml
@@ -43,7 +43,7 @@ jobs:
           - os: ubuntu-20.04
             name: MinGW
             cache-key: mingw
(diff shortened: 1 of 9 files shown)


Gadgetoid:
Awesome stuff, thank you!

PR #784, opened 2022-09-15
sdl: don't return error when reading 0 bytes

Daft-Freak:
Here's an easy one. Avoids errors at the end of reading files.

### DIFF
--- a/32blit-sdl/File.cpp
+++ b/32blit-sdl/File.cpp
@@ -60,9 +60,10 @@ int32_t read_file(void *fh, uint32_t offset, uint32_t length, char *buffer) {
   auto file = (SDL_RWops *)fh;
 
   if(file && SDL_RWseek(file, offset, RW_SEEK_SET) != -1) {
+    SDL_ClearError();
     size_t bytes_read = SDL_RWread(file, buffer, 1, length);
 
-    if(bytes_read >= 0)
+    if(!SDL_GetError()[0])
       return (int32_t)bytes_read;
   }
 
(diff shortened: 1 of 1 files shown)


ali1234:
As Github said, `size_t` is unsigned. So `>= 0` is always true.

Daft-Freak:
Ugh
```c
    SDL_ClearError();
    size_t bytes_read = SDL_RWread(file, buffer, 1, length);

    if(!SDL_GetError()[0])
      return (int32_t)bytes_read;
```

ali1234:
This doesn't seem right either. The doc says "call [SDL_GetError](https://wiki.libsdl.org/SDL_GetError)() for more information." but the doc for SDL_GetError says "You should not use the results of [SDL_GetError](https://wiki.libsdl.org/SDL_GetError)() to decide if an error has occurred! Sometimes SDL will set an error string even when reporting success. SDL will not clear the error string for successful API calls. You must check return values for failure cases before you can assume the error string applies."

So that means SDL_GetError might report a *different* error. I don't see how it is possible to tell the difference between EOF and error with this API.

Daft-Freak:
Yeah, that was part of the "ugh". I did clear the error before the call, but it's still nasty. Maybe we'll just have to rewrite it to not use rwops (as it's only using files anyway)...

ali1234:
Seems reasonable since the errors are per-thread and nothing else gets called in between.

I asked the question on SDL forums: https://discourse.libsdl.org/t/how-to-tell-whether-an-error-happened-in-sdl-rwread/38828

Gadgetoid:
Thank you both!

Daft-Freak:
Hah, they fixed the "can't tell if EOF or error" problem... in SDL3.

PR #785, opened 2022-10-27
Minor firmware blit flashing improvements

Daft-Freak:
First half: fixes some issues with failed flashes. Second half: use file size from headers instead of the actual file size.

- Fixes a crash when flashing a .blit fails
- Prevents overwriting things stored at the start of flash when running low on space (or trying to flash something unreasonably big)
- Simplifies size calc a little and is consistent with the size used after the file is flashed

A side effect of this is that you can append extra data to the end of a .blit and the flashing code ignores it (at least when loading from the SD card). Previously it would get flashed, but would likely get another .blit flashed on top of it later.


Gadgetoid:
I should probably prep a release at some point to sweep up all these changes!

Thank you, and sorry for the continued lack of attention here- I think burnout is slowly grinding me down :sob:

Issue #776, opened 2022-07-23
`make firmware` fails with Arch Linux default arm-none-eabi-gcc + arm-none-eabi-newlib


pogojotz:
More precisely the linkage fails:
```
[...]
/usr/lib/gcc/arm-none-eabi/12.1.0/../../../../arm-none-eabi/bin/ld: CMakeFiles/firmware.dir/firmware.cpp.obj: in function `std::_Vector_base<unsigned long, std::allocator<unsigned long> >::_M_allocate(unsigned int) [clone .isra.0]':
firmware.cpp:(.text._ZNSt12_Vector_baseImSaImEE11_M_allocateEj.isra.0+0x10): undefined reference to `std::__throw_bad_array_new_length()'
/usr/lib/gcc/arm-none-eabi/12.1.0/../../../../arm-none-eabi/bin/ld: /tmp/ccWLFRW6.ltrans2.ltrans.o: in function `std::_Vector_base<void*, std::allocator<void*> >::_M_allocate(unsigned int) [clone .isra.0]':
<artificial>:(.text._ZNSt12_Vector_baseIPvSaIS0_EE11_M_allocateEj.isra.0+0x10): undefined reference to `std::__throw_bad_array_new_length()'
collect2: error: ld returned 1 exit status
make[3]: *** [firmware/CMakeFiles/firmware.dir/build.make:317: firmware/firmware.elf] Error 1
make[2]: *** [CMakeFiles/Makefile2:2788: firmware/CMakeFiles/firmware.dir/all] Error 2
make[1]: *** [CMakeFiles/Makefile2:2795: firmware/CMakeFiles/firmware.dir/rule] Error 2
make: *** [Makefile:1138: firmware] Error 2
```

A quick search suggests this might be connected to GCC v11+, e.g. [[1]](https://github.com/JuliaLang/julia/issues/40703)

My installed packages:
```
$ pacman -Qs arm-none-eabi 
local/arm-none-eabi-binutils 2.38-1
    A set of programs to assemble and manipulate binary and object files for the ARM EABI (bare-metal) target
local/arm-none-eabi-gcc 12.1.0-1
    The GNU Compiler Collection - cross compiler for ARM EABI (bare-metal) target
local/arm-none-eabi-gdb 12.1-1
    The GNU Debugger for the ARM EABI (bare-metal) target
local/arm-none-eabi-newlib 4.2.0.20211231-1
    A C standard library implementation intended for use on embedded systems (ARM bare metal)
```

[...]

PR #788, opened 2023-01-23
Update stdlibs

Daft-Freak:
Should fix linker errors with GCC > 10. Not extensively tested, but does at least compile with GCC 9-12. (May need to check 8...)


Daft-Freak:
Slight firmware size improvement! 

Also hopefully fixes #776.

Gadgetoid:
All looks good with `gcc-arm-none-eabi-8-2019-q3-update` (8.3.1)

PR #790, opened 2023-02-02
stm32: remove readelf from post build commands

Daft-Freak:
Adds a lot of noise to build output for information that isn't useful most of the time. (`size` providing most of the useful info)

Leaving the .hex output, which is also of questionable usefulness...


Gadgetoid:
Might as well nuke .hex, too? I can't even remember what we used to do with them :laughing:

Daft-Freak:
Seems the pico-sdk stuff is also generating a .hex by default.

Gadgetoid:
Haha, thanks!

"no hexes" :mage_woman:

PR #791, opened 2023-02-03
Update emscripten actions to fix deprecation warnings

Daft-Freak:
Part 2 of fixing action warnings. (There wasn't an update for setup-emsdk when I wrote the other PR)


Gadgetoid:
Nice! Those deprecation warnings are _noisy_!

Thank you.

PR #792, opened 2023-02-08
Increase Emscripten stack size (+options update)

Daft-Freak:
Default stack was decreased in https://github.com/emscripten-core/emscripten/commit/157fcd4650161d9e0261966e2d3b529d62e1babc. 64k results in screen corruption in even the most minimal example, 256k is enough to fix this but increase to 1MB to be safe. (Old default was 5MB)

Also clean up to use the preferred `-sTHING=` syntax and remove the `EXTRA_` from `EXPORTED_RUNTIME_METHODS`.

### DIFF
--- a/32blit-sdl/CMakeLists.txt
+++ b/32blit-sdl/CMakeLists.txt
@@ -178,7 +178,7 @@ function(blit_executable NAME SOURCES)
 	if(EMSCRIPTEN)
 		set_target_properties(${NAME} PROPERTIES
 			SUFFIX ".html"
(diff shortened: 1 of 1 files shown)


Gadgetoid:
Thank you!

I went to grab a zip of build artifacts to try- force of habit from Pico- but there are none :cry: I don't think there's a reason for this? I'll raise a PR to add them when I get a moment.

Built it locally- taking an opportunity to test the guide- and it all looks good!

ahnlak:
You can just add `install(FILES ...` instructions to CMake to get it to add in the extra bits though (I *think* I've done this in Another Project and it seemed to work)

Daft-Freak:
You can do that... and I forgot that we already are:
https://github.com/32blit/32blit-sdk/blob/3c6de84d4dac618246babc86106085a6457d4fea/32blit-sdl/CMakeLists.txt#L187-L189

So uh, ignore that comment.

PR #793, opened 2023-02-10
docs: Make emscripten path warning more obvious.

Gadgetoid:
This bit me when I was testing #792 - hopefully splitting the configure and build steps, and decorating the step with a :warning:  will make it more obvious.

### DIFF
--- a/docs/Emscripten.md
+++ b/docs/Emscripten.md
@@ -4,7 +4,7 @@ Building in Emscripten - asm.js/WebAssembly - works on Linux and WSL.
 
 You will need to install the emscripten compiler. See https://emscripten.org/docs/getting_started/downloads.html for complete instructions, but generally you should just browse to your desired directory and:
 
-```
+```shell
 git clone https://github.com/emscripten-core/emsdk.git
 cd emsdk
 ./emsdk install latest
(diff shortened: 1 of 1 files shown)


Daft-Freak:
Looks more obvious to me.

(I wonder if enough time has passed that we can recommend `cmake -B build` instead of the mkdir/cd dance :thinking:. That's probably for another PR though.)

docs/Emscripten.md
```diff
@@ -14,16 +14,21 @@ source ./emsdk_env.sh
...
-``` shell
+```shell
 mkdir build.em
 cd build.em
 emcmake cmake .. -D32BLIT_DIR="/path/to/32blit/repo"
+```
+
+:warning: Make sure to include the `-D32BLIT_DIR="/path/to/32blit/repo"` parameter to the `cmake` command (even when building the SDK examples). You must specify an absolute path here when building with Emscripten.
+
+Once cmake has finished configuring your project you can build it:
```
Daft-Freak, line 27: "CMake" here I think

ahnlak:
I know the Emscripten docs tell you to use `emmake make` rather than a naked `make`, although functionally it ends up the same (I assume?)

Also, isn't it more 'Emscripten-y' to use `emrun` instead of doing the python web thingie by hand?

Daft-Freak:
Not using `emmake` usually works if you don't depend on environment variables at build time. (CMake's makefiles hardcode pretty much everything).

Using `emrun` would probably be a good idea, at least so stuff doesn't break if security gets more strict again (it sets all the magic headers so that things like threads work). Does have the disadvantage that you have to specify an example though.

PR #794, opened 2023-02-22
Pico board config improvements

Daft-Freak:
Splits the big if/else chain in 32blit-pico's CMakeLists into separate config files and the "drivers" into multiple files instead of a giant pile of #ifdefs.

Also includes some of the configs from the "all-the-boards" branch (the ones that don't need much extra code).


Daft-Freak:
Oh, and removes the pico-extras requirement for devices that don't use it. (PicoSystem most importantly)

Gadgetoid:
Ooooh, veeerryy nice!

I guess this is a huge step toward making this project a little less opinionated about target platforms and a touch more extensible. I think @ali1234 had some thoughts in that direction, but I can't remember where they were mentioned.

Daft-Freak:
At least for anything RP2040-based... 32blit-stm32 remains a bit of a mess. But uh, there's not a huge amount of hardware to target there.

Currently working through the ["everything" branch](https://github.com/Daft-Freak/32blit-beta/commits/pico-all-the-board-configs) attempting to not make things too messy. I think finishing VGA board support might be next...

Gadgetoid:
Le rebase fail? :eyes:

Daft-Freak:
Nope, just forgot to rebase entirely.

(Pressed the resolve button to see what it would do...)

Gadgetoid:
Oh yeah, I'm floored at just how incredibly bad GitHub's integrated conflict resolution is.

Gadgetoid:
Borrowed a little idea from this for Galactic (Blunicorn) Bluetooth Audio. Thanks :laughing:

PR #795, opened 2023-02-22
Remove deprecated compatibility wrapper

Daft-Freak:
The least old thing here is the blit wrappers (nov 2021)


Gadgetoid:
Ah, seizing the opportunity of a minor release to finally act out the deprecations I see? Nice!

Daft-Freak:
The first two have been around since 0.1.x!

Not that I need much of an excuse to delete things.

Gadgetoid:
Eventually we'll collapse back into one file :laughing:

PR #796, opened 2023-02-22
Remember launched item in launcher + font changes

Daft-Freak:
Saves the path of the last file to be launched and scrolls the file/folder lists back there when restarting the launcher. Also used the same font for the folder list as the file list... and a few minor things.

I think we used to have this back before the launcher was a separate thing, and it was annoying me recently. (Trying to test a big list of blits)

### DIFF
--- a/launcher/launcher.cpp
+++ b/launcher/launcher.cpp
@@ -18,9 +18,15 @@
 
 #include "credits.hpp"
 
-Dialog dialog;
-
 using namespace blit;
+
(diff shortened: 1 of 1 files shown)


Gadgetoid:
*charging 32blit intensifies*

Gadgetoid:
Whoop, almost forgot just how easy it was to update stuff on the blit, oh gosh it's so good. So. So. So very good.

PR #798, opened 2023-02-23
CI: Upload install results as artifacts.

Gadgetoid:
I instinctively went for the artifacts to try and test some changes on a PR and found there were none :cry: 

This PR should fix that!

### DIFF
--- a/.github/workflows/Emscripten.yml
+++ b/.github/workflows/Emscripten.yml
@@ -62,12 +62,23 @@ jobs:
         embuilder.py build sdl2 sdl2-image-jpg sdl2-net
         cmake --build . --config $BUILD_TYPE -j 2
 
+    - name: Prepare Site
+      run: |
+        mkdir -p site/examples
+        cp ../build/examples/*/*.{js,wasm} site/examples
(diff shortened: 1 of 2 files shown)


Daft-Freak:
Ah, using install instead of package to avoid the "double zip" problem.

Gadgetoid:
Hahaha 🤣 yeah that particular nonsense irritates me far more than it should. Enough that I’d waste way too much time working around it.

Daft-Freak:
Just noticed a problem here: MinGW/macOS don't do releases, so don't have a "release-suffix" set... which results in them generating a combined artifact.

... and as those are the largest two builds, it's quite big...

Screenshot 2023-02-27 at 13-47-21 Merge pull request #800 from ali1234_launcher_message · 32blit_32blit-sdk@3f96f95
(the top one)

Daft-Freak:
Okay, the size is mostly macOS... also takes nearly four minutes to upload its huge artifact. (I guess it's a combination of the fat x86+arm binaries and every bundle including the SDL framework)

PR #799, opened 2023-02-25
Only build picosystem hardware test for picosystem

Daft-Freak:
Running it on non-picosystem things can result in some brokenness due to the GPIO stuff. (Make me think I've broken VGA more than once...)


Gadgetoid:
Ooh that makes a lot of sense :laughing: thank you!

PR #800, opened 2023-02-26
Fix the message displayed when no launcher is installed

ali1234:
"32blit flash launcher.blit" has never been a valid command. It would have once been "32blit flash save --file launcher.blit" but that is now deprecated in favour of "32blit install launcher.blit".


Daft-Freak:
I think there was a hack in the tools to make that command work at some point... Anyway, `install` is definitely the right thing now.

Gadgetoid:
Ooof, I've seen that screen so many times and never clicked :grimacing:  thanks!

Issue #568, opened 2021-01-25
Buffer audio instead of firing callback at 22,000Hz


Gadgetoid:

Daft-Freak:
Checking if the DAC should be enabled every sample probably isn't great either... https://github.com/pimoroni/32blit-beta/blob/master/32blit-stm32/Src/sound.cpp#L22

Gadgetoid:
I did some sleuthing for the old audio code that used DMA and a buffer, and recalled that I never did manage to get it working. The system would just lock up solid and I had to abandon all hope in favour of... this travesty. So... yep. Long overdue a fix.

zenodante:
I will work on it, after I receive the hardware. Current my h750-dk boards don't have a DAC audio setup.  \(Q_Q)/

Daft-Freak:
I ended up figuring this out (eventually): https://github.com/Daft-Freak/32blit-beta/commit/9b8d4a268d107b3d35ba43957482833acba02414 (Very lightly tested)

Mostly because I suspected that this was why audio was weird in my new DOOM port...

PR #801, opened 2023-02-26
Avoid DMA channel/PIO SM conflicts from scanvideo/audio

Daft-Freak:
This is a bit of a hack due to pico-extras not using `*_claim_unused_*` (scanvideo using hardcoded numbers). Avoids more hardcoding/hacks in the SD card and DVI code.

### DIFF
--- a/32blit-pico/audio_i2s.cpp
+++ b/32blit-pico/audio_i2s.cpp
@@ -1,10 +1,14 @@
 #include "audio.hpp"
 #include "config.h"
 
+#include "hardware/dma.h"
+#include "hardware/pio.h"
 #include "pico/audio_i2s.h"
 #define AUDIO_SAMPLE_FREQ 44100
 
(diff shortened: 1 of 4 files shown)


Daft-Freak:
This might get less nasty, depending on what happens with the pico-extras PR I just opened...

Gadgetoid:
IMO Pico Extras looks... neglected? I'm tempted to just vendor the i2s driver and fix it downstream until we get something that's not in i2s purgatory.

Not that we should rush to reinvent the wheel, but I'm pretty sure I've had use cases where all the buffer pool handling is just unnecessary bloat. This might even be one of those cases- IIRC the STM32 32blit port just uses a single buffer with a half complete callback.

Gadgetoid:
> The STM32 port... doesn't even have a buffer... (#568)

ಠ_ಠ

I had forgotten this little factoid :grimacing:

PR #806, opened 2023-03-01
Update CubeMX project clocks to match final hardware

ali1234:
Final hardware uses a 12MHz crystal on HSE and the PLL dividers are found in main.c.


ali1234:
Just noticed that there is another one of these in 32blit-hardware repo.

Daft-Freak:
This one is my fault: https://github.com/32blit/32blit-sdk/pull/558, wouldn't be surprised if there's more code that's out-of-sync with the .ioc though.

ali1234:
I checked all the pin assignments and they look okay.

I did a pull against the hardware repo. Maybe we should get rid of the copy here?

Gadgetoid:
Might be sensible to drop the copy here, though now you've corrected it (Thanks, STM32Cube is freakin' awful) it wont ever change again, so it's not like we'll have to keep them in sync.

PR #807, opened 2023-03-03
Sync some blit_metadata changes to 32blit-pico

Daft-Freak:
Port two commits that were missed in the pico port. This does highlight that the first half of `blit_metadata` is identical between the three copies of it...


Daft-Freak:
... and one for `blit_executable`

Gadgetoid:
The second you refactor three copies of something into one function they'll need to diverge again :laughing:  it's a trap!

PR #809, opened 2023-03-05
Firmware: reject erase commands beyond the end of external flash

ali1234:
This prevents a crash when QSPI tries to read past the end of flash as a result of a bad serial command. 

There's also a patch waiting for the tools to stop it from doing that.


Daft-Freak:
Ouch, yeah out-of-bounds accesses can get quite nasty.

PR #810, opened 2023-03-05
Pico build size improvements

Daft-Freak:
Adds some flags to the BlitEngine library as the pico sdk does not set things globally. Also switches to nano stdlibs, mostly to avoid exceptions related code getting pulled in through libstdc++.

Saves ~24k of flash for the logo example and some RAM too.

The enabled policy requires CMake 3.13, but the pico-sdk requires that itself.

### DIFF
--- a/32blit-pico/CMakeLists.txt
+++ b/32blit-pico/CMakeLists.txt
@@ -141,6 +141,7 @@ endfunction()
 function(blit_executable_int_flash NAME SOURCES)
     add_executable(${NAME} ${SOURCES} ${ARGN})
     target_link_libraries(${NAME} BlitHalPico BlitEngine)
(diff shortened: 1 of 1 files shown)


Gadgetoid:
Interesting, same libstdc++ issue I was seeing with my MicroPython builds?

Daft-Freak:
Similar, but not as huge as the verbose stuff.

The standard linker script ends up putting anything unwind related in RAM... and any c++ compiled with exceptions support (including the non-nano libstdc++) pulls all of it in.

Gadgetoid:
Wow, .uf2 size differences say it all. Nice work, thanks!

Daft-Freak:
Mostly 12bc110 I think.

This was all discovered by accident while digging into low-level pico stuff for... other reasons.

PR #812, opened 2023-03-08
stm32: "fix" the Reset_Handler warning

Daft-Freak:
Point it to `do_init` (the closest thing to a game's entry point)

### DIFF
--- a/32blit-stm32/startup_user.s
+++ b/32blit-stm32/startup_user.s
@@ -54,6 +54,8 @@
   .section  .text.do_init
   .weak  do_init
   .type  do_init, %function
+  .global Reset_Handler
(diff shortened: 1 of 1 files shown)


Gadgetoid:
Haha, this is the change we've needed all this time :laughing:

Gadgetoid:
Good grief, 27m!? I wonder if we can get a close-enough test of macOS compatibility by just building in Linux with whatever whacky compiler they use. Clang? It's getting a bit silly lately.

Gadgetoid:
Good lord, you're not wrong. ~10min just to compress and upload.

ali1234:
Is that before or after the MacOS minute multiplier?

PR #813, opened 2023-03-08
blit_executable(_int_flash) improvements

Daft-Freak:
Basically this replaces `blit_executable_int_flash(name crc.cpp)` with `blit_executable(name INTERNAL_FLASH src.cpp)` but also:

- makes it work for external projects
- allows overriding the default with `-DBLIT_EXECUTABLE_INTERNAL_FLASH=`
- removes the sdl/pico stubs

(Somewhat inspired by `add_executable(WIN32/MACOSX_BUNDLE)` ... and needing a similar thing for pico standalone/loader builds)

### DIFF
--- a/32blit-config.cmake
+++ b/32blit-config.cmake
@@ -61,11 +61,18 @@ if (NOT DEFINED BLIT_ONCE)
 	# common helpers
 	function(blit_executable_args)
 		set(SOURCES)
+		set(INTERNAL_FLASH FALSE)
+
 		foreach(arg IN LISTS ARGN)
-			list(APPEND SOURCES ${arg})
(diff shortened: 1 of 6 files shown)


Gadgetoid:
oooh! Tidy! I guess there are precious few projects using this at all so it's a pretty safe merge?

Daft-Freak:
If any, you'd have to manually pull in 32blit-stm32's CMakeLists to get `_int_flash` to work in an external project without this...

(Hmm, if anyone was doing that, it would probably still work as the function is still there...)

Gadgetoid:
Boop beep I am ... not a bot :laughing: 

Thank you!

PR #803, opened 2023-02-27
Split examples

Daft-Freak:
#744, but moving _everything_. More discussion over there. WIP.

PR #814, opened 2023-03-08
ci: enable ccache for macOS

Daft-Freak:
Should help a bit with build times, also consistent with the other builds

### DIFF
--- a/.github/workflows/cmake.yml
+++ b/.github/workflows/cmake.yml
@@ -49,7 +49,7 @@ jobs:
           - os: macos-latest
             name: macOS
             cache-key: macos
-            cmake-args: -DCMAKE_OSX_ARCHITECTURES=x86_64\;arm64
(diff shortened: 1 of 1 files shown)


Daft-Freak:
Added a patch from #803 to fix the macOS and MinGW builds sharing an artifact.

Doesn't do anything about the macOS artifacts containing 108 copies of SDL though. (3 for each example due to artifacts not doing symlinks, > 3MB each)

Daft-Freak:
It should be 3.2MB (and is in release packages), but the artifact upload follows symlinks and makes it 3x bigger...

(Which is a pretty good size considering that the windows .dll is 2.5MB and doesn't contain two architectures)

Don't know if we can get `fixup_bundle` to put dependencies somewhere that isn't in the bundle. We could do some hack and not run it for artifacts, but then the output is even harder to run. (not signed, so you already have to mess with security settings)

Gadgetoid:
Ugh I forgot about macOS security settings. Hacks are probably generally a bad idea :laughing:

Daft-Freak:
Hah, this is a weird case where the "double zip" thing would actually help...

PR #815, opened 2023-03-09
firmware: fix launch early-out errors

Daft-Freak:
The reloc header check returned ~0 instead of false... and leaked the file.

Also if file opening the file failed we'd leave the flash unmapped, so move the unmap after both.


Daft-Freak:
Uh, I'm just going to merge this bug fix...

PR #818, opened 2023-05-17
Docs: Update references to examples.

Gadgetoid:
Counterpart to #803 focusing on documentation updates.

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,7 +42,7 @@ The `32blit-stm32` directory contains the STM32 HAL for 32blit, compatible with 
 
 ## Examples / Projects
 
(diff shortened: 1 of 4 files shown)


Daft-Freak:
Does this replace my attempt in the other PR? (so I should drop those to avoid conflicts)

Gadgetoid:
Ugh of course I'd forgotten about those. I should probably go over yours and make sure any changes I've missed are reflected here before you consign them to infinity.

I was wondering where we'd got with the .sln file and your changes confirm that.

But yeah TLDR they don't replace them, just supplement them, I need to rebase.

Daft-Freak:
TBH I didn't put much effort into those. Mostly "build that not this" I think.

PR #819, opened 2023-05-22
pico: SPI-mode SD card support

Daft-Freak:
This was the first thing in my branch after the latest rebase. Only includes the SPI driver, not the native SD driver (that still has issues).


Daft-Freak:
Extra commit to prepare for PicoVision. (...or anything else that uses different pins for the SD card)

PR #821, opened 2023-05-22
st7789 config and parallel support

Daft-Freak:
Mostly because I said you could configure the pins from the config, but that patch wasn't merged yet... Also includes the parallel support and the tufty config and input code so that has a user.

I think I have a refactor of this code into a generic DBI driver _somewhere_ though it may be multiple somewheres... and in other projects :thinking:


Gadgetoid:
Replaced by #827?

Daft-Freak:
The first commit was, the rest is... less important.

Gadgetoid:
Hmm, conflicted on some pio headers...

Daft-Freak:
Addition of SD support probably... My priorities are elsewhere (if anywhere), but I do actually have my tufty now...

Daft-Freak:
Eh, rebased anyway. (Not tested but nothing should've changed to break it...)

Daft-Freak:
Turns out this (and anything else without a `config.h`) was broken... now it isn't (and has a `config.h` anyway)


Also I just ran pico3d on it :smile:

Daft-Freak:
Hmm, now I have other changes that conflict with this, but only because they delete a file that this adds :thinking: 

(https://github.com/Daft-Freak/32blit-beta/compare/pico-st7789-config-parallel...Daft-Freak:32blit-beta:pico-gpio-input)

PR #822, opened 2023-05-22
Extend .blit header with api/device info

Daft-Freak:
Some patches from the pico launcher/api stuff. Adds a device id and the api version to the .blit file header, taking advantage of there being an unused constant `00 00 00 90` at the end for backwards compatibility (first byte is repurposed as device id, non-0 means the api info is there).

I have no idea where the name "STM32Blit" came from though. Should probably be something like STH32H7 (existing .blits are built for cortex-m7 and assume STM32 RAM layout, though there isn't a loader for anything that isn't a 32blit...)

### DIFF
--- a/32blit-stm32/executable.cmake
+++ b/32blit-stm32/executable.cmake
@@ -1,7 +1,7 @@
 set(MCU_LINKER_SCRIPT "${CMAKE_CURRENT_LIST_DIR}/${MCU_LINKER_SCRIPT}")
 set(HAL_DIR ${CMAKE_CURRENT_LIST_DIR})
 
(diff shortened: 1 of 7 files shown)


Gadgetoid:
> I have no idea where the name "STM32Blit" came from though.

I have no earthly clue, it's not even a nice round number of bytes :laughing:

Daft-Freak:
Oh, it does allow checking files for compatibility _before_ launching them as well. (WIP: https://github.com/Daft-Freak/32blit-beta/commit/99fb0985eec241757c32943b9af2489004286be9)

Daft-Freak:
Okay, now this works:
Launcher0
(Those are pico-blits)

PR #825, opened 2023-06-29
Delete old stdlibs files when redownloading

Daft-Freak:
If we were re-downloading the stdlibs because the hash changed, we wouldn't extract it later because the dir already existed.

I only noticed this because my build script was failing to link everything after a system upgrade. (That is _after_ working around the python annoyingness with the tools...)


Daft-Freak:
Well, that saved me having to remember to check if the actions had finished yet.

PR #826, opened 2023-07-01
SDL screen mode/size improvements

Daft-Freak:
Syncs SDL screen mode handling with the other ports and allows overriding the screen size on the command line. (Can finally test picosystem things in SDL without hacking the sources)

Also fixes a bug with surfaces taller than 255. (I tried 240x320 for fun).

(I definitely did this because of the picosystem thing and not because of blitsim.)


Daft-Freak:
Oops, included #824.

My preferred solution is to merge that though, bit hard to build things here without it.

PR #827, opened 2023-07-14
More pico config

Daft-Freak:
Adds pin config for ST7789 displays and moves some of the definitions to a board-specific config.h instead of specifying all of them on the command line.

This is before I add _more_ defines for SD card config... (Which isn't here, as SD card support is in another PR)

I've only tested this on PicoSystem, as I don't currently have the display packs or a monitor for my VGA board.


Daft-Freak:
Okay, tested the vga board hires change on pico dv (it does the same thing)... after spending way too much time getting that to work again.

PR #828, opened 2023-07-17
Add Surface::get_pixel and remove special case from blit code

Daft-Freak:
After repeating my "blit that is actually a fill" optimisation for the third time, I worked out a way to avoid it entirely and improve TileMap perf a bit. Also adds a useful `get_pixel` method to `Surface` (returns the pixel converted to a `Pen`).

I haven't benchmarked the change to `stretch_blit`, but I'd expect a similar effect there.

(Making the blit code smaller is also a good thing for the firmware)


Gadgetoid:
Wow, I can't even wrap my brain around what the edge case was doing and why, but it was... janky 😬 

Thanks!

PR #829, opened 2023-07-18
PicoVision

Daft-Freak:
Opening this to keep track of things now that the dependencies have been merged. Most things are at least working at this point. The biggest limitation is that `screen.data` is null, so anything that directly reads/writes screen data is going to be broken. (The recently added `get_pixel` helps with reading a little)

Potential breakages:
 - ~~28ac10d5938dd1a5f0ba2c36980d6512fa37741d likely breaks VGA board (needs a limit on samples generated each update)~~
 - ~~7553ae468b5af45eeb616176846de3298909d87c~~
 - ~~Screen mode changes (c275eb2ce8fb242a4d00ab0827f12edef1fbdebd, 322adce8cef94fc9a3cf4890a6b29a06974b832d) affect everything and _definitely_ require an api version bump~~
 - ~~Branch is a bit broken in the middle (wrote the code on top of a bunch of WIP screen changes)~~

PicoVision-specific unfinished-ness:
- ~~Currently claims the screen is RGB565, which is wrong but you can't directly access the data anyway~~
- Small blits/blends are very unoptimised
- Large blits with no transparency are also suboptimal (always reads old data)
- ~~pimoroni-pico usage is uh, questionable (using one and a half drivers, see 32blit-pico/CMakeLists.txt changes)~~
- ~~Clock override (34ac6b7a0c1990b4c7bae50e2c4ee6232a4ad0af)~~

I also have plans to allow more screen modes (and 640x480 without recompiling). That shouldn't involve huge changes though... and I need to add some way to write a bunch of raw data to I can port DaftBoy (if only blits from 16-bit formats were supported. Not sure I want to add a format that's only supported as a blit source on one device...)


32blit/engine/engine.cpp
```diff
@@ -14,9 +14,10 @@ namespace blit {
   void (*render)(uint32_t time)                     = nullptr;
 
   void set_screen_mode(ScreenMode new_mode) {
-    auto &new_screen = api.set_screen_mode(new_mode);
-    screen = Surface(new_screen.data, new_screen.format, new_screen.bounds);
-    screen.palette = new_screen.palette;
+    if(new_mode == ScreenMode::hires_palette)
+      set_screen_mode(ScreenMode::hires, PixelFormat::P);
+    else
+      set_screen_mode(new_mode, (PixelFormat)-1);
```
Daft-Freak, line 20: I should probably define a PixelFormat::Default or something instead of the magic -1...

Daft-Freak:
Now with less breakage to other things and actually compiles before the end! (Now I just need to fix audio and maybe make it not depend on core1... for uh, _reasons_)

Daft-Freak:
Just need to merge the screen and audio PRs and rebase then this should be safe.

Gadgetoid:
I think Mike has rained on your parade a little by adding vertical scrolling and changing the number of available scroll whatsits, but I guess in principle you can just target whatever prior version of the pico-stick driver since you don't use fancy stuff anyway?

Daft-Freak:
The unfancy-ness should mean it's not too hard to fix hopefully. Though this does make me wonder if I should make it automatically grab the picovision repo to simplify setup :thinking: 

Need to do a rebase/tidy tomorrow (probably) anyway...

Gadgetoid:
Ugh that reminds me I need to get Mike to fix permissions on the pico-stick repo so releases can upload! Or maybe I can...

Gadgetoid:
Just in case it's useful: I have added a new release for pico-stick, which includes both normal and widescreen firmware headers and elf files: https://github.com/MichaelBell/pico-stick/releases/tag/v0.1.0

Daft-Freak:
Might decide to copy the relevant drivers instead of the dependency... but probably not right now. :thinking: 

Anyway this should have no risk of affecting anything else now, so un-drafted. Also, claiming a more accurate format... it uh, may be named differently to the official code though...

Daft-Freak:
Now with more blend/blit fixes/improvements for https://github.com/Daft-Freak/Pico3D/tree/picovision (Also need the "special blit" patch for DaftBoy)

Gadgetoid:
Are there .blit builds for PicoVision anywhere, or was that too borken?

Gadgetoid:
Patch for Geometry to support the 8bitdo Zero 2 (ie: re-enabled and tweaked joystick support that totally doesn't allow 32blit users to press the dpad and joystick at the same time noooo)-

```patch
diff --git a/examples/geometry/geometry.cpp b/examples/geometry/geometry.cpp
index 970dc8e..7178619 100644
--- a/examples/geometry/geometry.cpp
+++ b/examples/geometry/geometry.cpp
@@ -415,12 +415,12 @@ void update(uint32_t time) {
 
     if (pressed(Button::DPAD_LEFT))  { player1.rotational_velocity += rotation_speed; }
     if (pressed(Button::DPAD_RIGHT)) { player1.rotational_velocity -= rotation_speed; }
-    //player1.rotational_velocity -= joystick.x * pi / 720;
+    player1.rotational_velocity -= joystick.x * pi / 1440;
 
     if(player1.energy >= MOVEMENT_COST) {
         if (pressed(Button::DPAD_UP))    { movement.y -= 0.03f; }
         if (pressed(Button::DPAD_DOWN))  { movement.y += 0.03f; }
-        //movement.y += joystick.y / 10.0f;
+        movement.y += joystick.y / 40.0f;
 
         if(pressed(Button::DPAD_UP) || pressed(Button::DPAD_DOWN)) {
             player1.energy -= MOVEMENT_COST;
```

Daft-Freak:
> Are there .blit builds for PicoVision anywhere, or was that too borken?

Should still be over here: https://github.com/Daft-Freak/32blit-beta/tree/picovision-and-everything-else (if I didn't break it). It's stuck behind an uh, somewhat large API refactor...



> Patch for Geometry to support the 8bitdo Zero 2 (ie: re-enabled and tweaked joystick support that totally doesn't allow 32blit users to press the dpad and joystick at the same time noooo)-

... or press Up + Select for 5 seconds (Left + Select to undo, super-well-documented feature :laughing:)

Gadgetoid:
> ... or press Up + Select for 5 seconds (Left + Select to undo, super-well-documented feature 😆)

Wait is this a feature of the *controller* or the build? :laughing:

PR #830, opened 2023-07-19
Optimise stretch_blit_vspan

Daft-Freak:
Copy the code for stretch_blit and remove the x logic.

For raycaster (one of the two users):
- -2 ms on 32blit
- -6 ms on PicoSystem
- -27 ms (ish) on PicoVision (drawing vertically is still bad for perf)

A lot of the improvements can also be gained by replacing the contents of `stretch_blit_vspan` with `stretch_blit(src, {uv, Size{1, sc}}, {p, Size{1, dc}});`. Probably because I already optimised that a lot.

### DIFF
--- a/32blit/graphics/surface.cpp
+++ b/32blit/graphics/surface.cpp
@@ -542,24 +542,34 @@ namespace blit {
    * \param dc
    */
   void Surface::stretch_blit_vspan(Surface *src, Point uv, uint16_t sc, Point p, int16_t dc) {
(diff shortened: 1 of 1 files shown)


Daft-Freak:
Oh, PicoSystem number may be slightly off as I'm running from the PicoVision branch with it's hacked down clock. It's definitely still an improvement though.

Gadgetoid:
Nice, thank you!

PR #832, opened 2023-07-20
pico: 8BitDo Zero 2 HID mapping

Daft-Freak:
+ extra code to handle not having a dpad and mapping the stick to the dpad. Tested on the picovision-and-everything-else branch.

(The other controller is probably going to involve a tinyusb issue...)

### DIFF
--- a/32blit-pico/input_usb_hid.cpp
+++ b/32blit-pico/input_usb_hid.cpp
@@ -30,6 +30,7 @@ struct GamepadMapping {
 static const GamepadMapping gamepad_mappings[]{
   {0x15320705,  0,  1,  3,  4, NO, NO, NO, NO, 16, 15, 13}, // Razer Raiju Mobile
   {0x20D6A711,  2,  1,  3,  0, NO, NO, NO, NO,  8, 12, 10}, // PowerA wired Switch pro controller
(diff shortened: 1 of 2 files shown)


Gadgetoid:
Haha, I hope you had fun with those :grimacing: 

Only suitable controllers we had, really :(  need something custom that does Bluetooth LE HID over GATT and some sensible USB HID.

Daft-Freak:
Could probably hack something together with a pico w and existing controller parts... :thinking: 

Also, forgot that I had this as a fallback: https://github.com/Daft-Freak/picosystem-gamepad

Gadgetoid:
Fair point a Pico W could actually be both a USB HID controller and a HID-over-GATT-profile controller.

I wonder if there are off-the-shell shell parts we could just make a PCB for :thinking:

Daft-Freak:
Turns out we don't need the last commit, just hold up+select for 5 seconds. (It's in the FAQs that I definitely didn't just read.)

But also:

>  ### Can I use this controller via the USB connection?
>   No, you can not. The USB connection is for firmware upgrade and power charging only.

Pretty sure that I just did. :shrug:

Daft-Freak:
Also rebased the PicoVision PR on top of this, should be able to grab the artefacts from there.

(Current status of the other controller: https://github.com/hathach/tinyusb/issues/2163)

PR #834, opened 2023-08-29
pico: include config in led.cpp

Daft-Freak:
Otherwise the LED is never going to get enabled (broken by #827 as this file still compiles without its #defines, it just does nothing)


Gadgetoid:
Me, merge a broken PR? noooooo :laughing: 

Thank you!

PR #835, opened 2023-08-30
Add VERBATIM to CMake custom commands

Daft-Freak:
Someone thought using a build dir with () in it was a good idea, resulting in a lot of
```
/bin/sh: -c: line 1: syntax error near unexpected token `('
```

Adding this magic flag ~~allows them to continue making poor choices~~ fixes that problem.

Also the CMake docs say that this is a good idea:

     Use of VERBATIM is recommended as it enables correct behavior.


Gadgetoid:
Urge to look up why this isn't the default :laughing:   sheesh!

Thank you!

PR #836, opened 2023-09-04
Some screen mode refactoring

Daft-Freak:
Non-picovision screen-related changes from that branch (in an attempt to clean it up to not break other things).

Most break-y change is always using `set_screen_mode_format` at the API level to simplify some stuff. (Could probably delete the old one from -sdl/-pico after this).

Tested on SDL, 32blit, PicoSystem and VGA board... I think.

(Still not sure about `(PixelFormat)-1`, but also didn't want to think about `Surface(..., PixelFormat::Default)` being a thing)

### DIFF
--- a/32blit-pico/config.h
+++ b/32blit-pico/config.h
@@ -8,6 +8,10 @@
 #define ALLOW_HIRES 1
 #endif
 
+#ifndef DEFAULT_SCREEN_FORMAT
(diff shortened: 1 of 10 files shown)


Gadgetoid:
Presto mergo testo ... uh no 😢 

Thank you!

PR #837, opened 2023-09-26
Pico audio improvements

Daft-Freak:
This should be all the relevant patches from my branches so that PicoVision support doesn't break VGA board (and also not require core1 to be enabled).


Gadgetoid:
Thank you! 🎉 

(Yes I have been a little distracted by Pi 5 mayhem)

Daft-Freak:
Hah, I'm getting distracted now just _thinking_ about it...

PR #841, opened 2024-01-05
pico: add "neopixel" led support (WS2812)

Cpasjuste:
This PR add "neopixel" rgb led support to the pico platforms by defining "LED_WS2812_PIN".

Brightness is handled/simulated with the "pen LED" alpha channel.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,7 +23,7 @@ bld/
 build/
 build.stm32/
 build.mingw/
-cmake-build-debug/
+cmake-build-*/
 [Bb]in/
 [Oo]bj/
 [Ll]og/
--- a/32blit-pico/led.cpp
+++ b/32blit-pico/led.cpp
@@ -13,6 +13,24 @@
 #if defined(LED_R_PIN) && defined(LED_G_PIN) && defined(LED_B_PIN)
 static const int led_pins[]{LED_R_PIN, LED_G_PIN, LED_B_PIN};
 #define HAVE_LED
+#elif defined(LED_WS2812_PIN)
+#include "hardware/pio.h"
+#include "ws2812.pio.h"
+
+static uint32_t last_color = 0;
+static int pio_sm = -1;
+
+static void put_pixel(uint8_t r, uint8_t g, uint8_t b, uint8_t a) {
+  if(pio_sm < 0) return;
+  uint32_t color =
+          (((uint32_t)((r * a) >> 8) << 8)
+          | ((uint32_t)((g * a) >> 8)  << 16)
+          | (uint32_t)((b * a) >> 8)) << 8u;
+  if(color != last_color) {
+    pio_sm_put_blocking(pio0, 0, color);
+    last_color = color;
+  }
+}
 #endif
 
 void init_led() {
@@ -31,6 +49,15 @@ void init_led() {
   bi_decl(bi_1pin_with_name(led_pins[0], "Red LED"));
   bi_decl(bi_1pin_with_name(led_pins[1], "Green LED"));
   bi_decl(bi_1pin_with_name(led_pins[2], "Blue LED"));
+#elif defined(LED_WS2812_PIN)
+  PIO pio = pio0;
+  uint pio_offset = pio_add_program(pio, &ws2812_program);
+  pio_sm = pio_claim_unused_sm(pio, true);
+  if(pio_sm > -1) {
+    ws2812_program_init(pio, pio_sm, pio_offset, LED_WS2812_PIN, 800000, true);
+  } else {
+    printf("LED_WS2812: could not find a free pio sm\r\n");
+  }
 #endif
 }
 
@@ -45,5 +72,7 @@ void update_led() {
   pwm_set_gpio_level(led_pins[1], value);
   value = (uint16_t)(std::pow((float)(api.LED.b) / 255.0f, gamma) * 65535.0f + 0.5f);
   pwm_set_gpio_level(led_pins[2], value);
+#elif defined(LED_WS2812_PIN)
+  put_pixel(api.LED.r, api.LED.g, api.LED.b, api.LED.a);
 #endif
 }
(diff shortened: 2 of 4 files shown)


32blit-pico/led.cpp
```diff
@@ -31,6 +49,16 @@ void init_led() {
   bi_decl(bi_1pin_with_name(led_pins[0], "Red LED"));
   bi_decl(bi_1pin_with_name(led_pins[1], "Green LED"));
   bi_decl(bi_1pin_with_name(led_pins[2], "Blue LED"));
+#elif defined(LED_WS2812_PIN)
+  PIO pio = pio0;
+  uint pio_offset = pio_add_program(pio, &ws2812_program);
+  pio_sm = pio_claim_unused_sm(pio, true);
```
Daft-Freak, line 55: `required` is `true` here, so the error check is redundant. (It'll panic before it gets there)

32blit-pico/led.cpp
```diff
@@ -13,6 +13,24 @@
...
+
+static uint32_t last_color = 0;
+static int pio_sm = -1;
+
+static void put_pixel(uint8_t r, uint8_t g, uint8_t b, uint8_t a) {
+  if(pio_sm < 0) return;
+  uint32_t color =
+          (((uint32_t)((r * a) >> 8) << 8)
+          | ((uint32_t)((g * a) >> 8)  << 16)
+          | (uint32_t)((b * a) >> 8)) << 8u;
```
Daft-Freak, line 28: Not sure about having the alpha channel do something here when it doesn't for "regular" RGB LEDs...

32blit-pico/led.cpp
Daft-Freak: Also, probably want a `bi_decl` for the pin.

32blit-pico/ws2812.pio
```diff
@@ -0,0 +1,85 @@
...
+    int cycles_per_bit = ws2812_T1 + ws2812_T2 + ws2812_T3;
+    float div = clock_get_hz(clk_sys) / (freq * cycles_per_bit);
+    sm_config_set_clkdiv(&c, div);
+
+    pio_sm_init(pio, sm, offset, &c);
+    pio_sm_set_enabled(pio, sm, true);
+}
+%}
+
+.program ws2812_parallel
```
Daft-Freak, line 50: I'm guessing this file is copied from the example, but the parallel program is unused so might as well delete it.

.gitignore
```diff
@@ -23,7 +23,7 @@ bld/
 build/
 build.stm32/
 build.mingw/
-cmake-build-debug/
+cmake-build-*/
```
Daft-Freak, line 26: Not a comment on the actual change, but I've just noticed that the previous three lines are duplicated with line 340/341... (except that one has a wildcard)

Daft-Freak (commented):
Think this was the last "driver" that only supported one thing. Beat me to it :laughing: 

I assume you have a user for this somewhere?

32blit-pico/led.cpp
Cpasjuste: > `required` is `true` here, so the error check is redundant. (It'll panic before it gets there)

My bad ! :)

Cpasjuste: > Also, probably want a `bi_decl` for the pin.

I didn't know about bi_decl, and a very little of picotool. This is very nice !
```
File cmake-build-debug-pico/examples/saves/saves.uf2:

Program Information
 name:            saves
 version:         v1.0.0
 web site:        https://github.com/32blit/32blit-sdk
 description:     Save data example.
 features:        UART stdin / stdout
 binary start:    0x10000000
 binary end:      0x10018460
 embedded drive:  0x10600000-0x10800000 (2048K): 32Blit

Fixed Pin Information
 0:   UART0 TX, X Button
 1:   UART0 RX, Y Button
 2:   DOWN Button
 3:   LEFT Button
 7:   UP Button
 8:   Display D/C
 9:   Display CS
 10:  Display SCK
 11:  Display TX
 12:  Display Backlight
 13:  Display Reset
 16:  NeoPixel RGB LED (WS2812)
 24:  B Button
 25:  A Button
 29:  RIGHT Button

Build Information
 sdk version:       1.5.1
 pico_board:        adafruit_feather_rp2040
 boot2_name:        boot2_generic_03h
 build date:        Jan  5 2024
 build attributes:  Debug
```

Cpasjuste: Fixed in d2d5f7f28b8c6d0a3afe1a032bbb06d97f1f02d4 i hope.

32blit-pico/led.cpp
Cpasjuste: This is really working (tested), colors are "dimmed" i would say, sending 0 as alpha will completely turn off the led.

32blit-pico/ws2812.pio
Cpasjuste: Yes you're right, and i can say i didn't even look at it before submitting the PR  :x
This is fixed in 51984714665b03408279b8d540b8d872593209ea

Cpasjuste:
> Think this was the last "driver" that only supported one thing. Beat me to it 😆
> 
> I assume you have a user for this somewhere?

Yes! For the "long" storry, i'm developing my own rp2040 handheld device (since a while now since my time is very limited...) initially to run some gb/nes/sms emulators i "ported". Then i wanted to develop a framework for it, which i did and it works very fine (linux/pico). Then i discovered this project.... which would have saved me lot of work :)

But well, all this stuff is fun after all, so it's cool. Here is a short video of the device running 32blit (which use a feather rp2040, with the rgb led) :  https://files.mydedibox.fr/files/dev/microboy/microboy-000.mp4

32blit-pico/led.cpp
Daft-Freak: My point was more that it's inconsistent to do that for one type of LED but not the other (and also the other port that has an LED).

But doing it everywhere would be a bigger change... and entirely independent from adding a new driver...

PR #843, opened 2024-01-24
Add can_launch API

Daft-Freak:
This is a few patches shared by the "SDL launcher" and "pico launcher" branches. Moves checking if a file can be launched over to the "firmware" side. Should allow better compatibility checks when combined with #822 (and less "Oops!").

Also makes it a little easier to implement alternative launchers. (Which _may_ be what made me dig these up...)

### DIFF
--- a/32blit/engine/api_private.hpp
+++ b/32blit/engine/api_private.hpp
@@ -53,6 +53,13 @@ namespace blit {
     Pen *palette = nullptr;
   };
 
+  enum class CanLaunchResult {
+    Success = 0,
+    UnknownType,       /// no known handler for this file
+    InvalidFile,      /// file is not valid/doesn't exist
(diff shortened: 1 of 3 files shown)


Gadgetoid:
Alternate launchers? 👀

Daft-Freak:
Was trying to make something that works in lores on a PicoSystem:

https://github.com/32blit/32blit-sdk/assets/3074891/3cf7f0af-8231-4334-ab23-06bc1168b6ee

... though I haven't actually implemented the launching part... (too busy doing the scrolling)

Gadgetoid:
👀

Daft-Freak:
Figured I'd ~~forgot about~~ left it long enough. Might even get back to doing small launcher-y stuff at some point. :shrug:

PR #844, opened 2024-02-22
pico: generic st7789 rotation config

Daft-Freak:
The values are not exactly the same, but the scan order was previously somewhat inconsistent.

Wrote this patch a few weeks ago then forgot about it...


Gadgetoid:
Ooh, this makes sense! Nice, thanks!

Daft-Freak:
That sound like a "you should merge this" to me. :laughing:

(Should've mentioned that I tested this on three displays with different default rotations)

PR #846, opened 2024-05-15
Fix actions

Daft-Freak:
Both a bit workaround-y, but I don't have Windows or macOS handy right now.


Gadgetoid:
Ah the ceaseless and inevitable march of bitrot.

Playing whack-a-mole with actions and CI issues on Python projects at the moment 🤣

I don’t see any failed actions runs, though, did this soft break? 🤔

Daft-Freak:
Failed when I pushed my fork (https://github.com/Daft-Freak/32blit-beta/actions/runs/9101170780/job/25018338322, https://github.com/Daft-Freak/32blit-beta/actions/runs/9101388783/job/25018458593).

Was going to just switch everything to `pipx install 32blit` since I know that works, but it's not the only thing being `pip install`-ed...

Now I need to figure out why the tools are broken...

Daft-Freak:
Gonna merge this so things aren't broken and sort out python-y stuff later. Maybe.

PR #847, opened 2024-05-16
Fix pico metadata title

Daft-Freak:
The pico SDK sets a default program name if PICO_PROGRAM_NAME is not set. As we're specifying all the metadata using bi_decl directly, disable the fallback.


Daft-Freak:
Oh right, the actions are broken.

Issue #484, opened 2020-12-23
blit::random() cannot be used in global initializers (except on device)


a2:
If I want to store a random number that's globally initialized, I can't do that with `blit::random()` until `init()` is called. This works on a 32blit device though, which I think is because `api` is initialized differently there.

```cpp
const uint32_t _test = blit::random();
```

This can be worked around by removing the `const` modifier and assigning in `init()`. However, it becomes more complicated when `blit::random()` is used in other parts of the code:

```cpp
class Flower {
  public:
    Flower() : height(blit::random()) {};
    const uint32_t height;
};

Flower flower;
```

Right now my interim solution is to use a pointer like so, and initializing `Flower` later in `init()`:

```cpp
Flower *flower;
```

Daft-Freak:
Yeah, I've hit this one before, not sure there's actually a way to fix this...

The only reason this works on device is because the API is initialised before the game even exists. (API is setup by the firmware. which then starts loading the game, for SDL the constructors have already happened before we can do anything.)

Gadgetoid:
I’m not totally averse to quietly “ifdef”ing this issue under the rug. I don’t think there’s any other way it could be fixed?

The whole concept of the API is a little contrived for SDL, where there’s no real separation between the “HAL” and game. It feels like it should not be intrinsic to the “engine” but constitutes something that’s really neither “engine” nor “HAL”. Maybe these a fundamental architectural issue behind this bug?

Daft-Freak:
Hmm, an idea: add a API_FUNC macro
```c++
//SDL
#define API_FUNC(name) name
//STM32
#define API_FUNC(name) (*name)
```
Then all the API functions could be implemented for SDL as
```c++
uint32_t  API::random()
{
...
}
```

Going even further, they could all be regular function declarations, implemented by either 32blit-sdl32, or a new 32blit-stm32-user library (that just forwards trough the API section). Though that would end up with some stuff having multiple levels of wrapper...


(I think these were function pointers even before the API stuff, so it's always had this issue... I just got accidentally fixed on device by the user/firmware split)

Daft-Freak:
Hmm, first idea is not so bad https://github.com/pimoroni/32blit-beta/compare/master...Daft-Freak:sdl-no-ptrs (needs some indent/namespace tidying)

I had an attempt at going further which... did not go so well.

Gadgetoid:
[...]

PR #851, opened 2024-07-01
API refactor

Daft-Freak:
Split the API struct into a "const" and "data" part. This is a result of me not wanting to reserve 2K of RAM on pico. After this, all the function pointers can be constants (may improve #484) and the data part only needs 64 bytes.

For compatibility, 32blit-stm32 is still using the old layout by inserting padding and overlaying the two structs on top of each other.

These are (I think) the only remaining non-pico changes on my big pico branch.

### DIFF
--- a/32blit-pico/input/gpio.cpp
+++ b/32blit-pico/input/gpio.cpp
@@ -88,5 +88,5 @@ void update_input() {
       new_buttons |= 1 << i;
   }
 
-  blit::api.buttons = new_buttons;
+  blit::api_data.buttons = new_buttons;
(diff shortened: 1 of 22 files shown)


Gadgetoid:
In hindsight this makes an awful lot of sense.

Daft-Freak:
Fixed warnings from these patches being older than `can_launch` and needed to rebase for recent build fixes.

I'll have to add the STM32 nastiness to the list of things that can be cleaned up if there's an API break... (Also, I should actually write that list)

PR #859, opened 2024-09-15
Actions updates

Daft-Freak:
The usual version bumps for deprecation warnings, plus the ubuntu version as 20.04 is EOL next year (may also solve a problem I'm having)...

Need to make sure there are no surprises from the compiler version bump (though I'm locally using 14.1 so I hope not). Also might not want to use latest ubuntu to ensure wider compatibility :thinking: 

(22.04 is an arm gcc bump from 9 -> 10, 24.04 is -> 13)

### DIFF
--- a/.github/workflows/Emscripten.yml
+++ b/.github/workflows/Emscripten.yml
@@ -29,8 +29,8 @@ jobs:
 
     - name: Install deps
       run: |
-        sudo apt update && sudo apt install doxygen graphviz python3-setuptools
-        pip3 install 32blit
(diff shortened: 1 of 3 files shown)


Gadgetoid:
> ubuntu version as 20.04 is EOL next year

💀😭

I guess I should make sure I’m switched over to sideloading ARM GCC on all of my Pico builds!

Daft-Freak:
Went with smaller OS bump, also made sure releases weren't broken.

PR #861, opened 2024-09-26
Big tilemap/map rework

Daft-Freak:
TileMaps being slow and a bit hard to use finally annoyed me enough... Requires https://github.com/32blit/32blit-tools/pull/105 for layer names.

Adds a new `TiledMap`, which covers most of the functionality of `TileMap` and `Map`. (`TileMap` is kept as a deprecated alias of what is now `TransformedTileLayer` for some level of compatibility)

New features:
- Can load the 16-bit tile output from the tools
- Also improved handling of multiple layers (don't need to `load_tmx` each layer, layer names from Tiled)
- Easier (and much faster) support of just scrolling a map 
- Removes the power of two bounds and 16 column tileset restrictions unless you specifically request the transformed variant (`TiledMap(..., LAYER_TRANSFORMS)`)
- Doesn't leak memory (at least, if the map allocated it)
- Supports loading maps from files as well as embedded assets

Most basic usage is:
```c++
sprites = Surface::load(asset_tileset);
map = new TiledMap(asset_tilemap, sprites);
...
map->set_scroll_position(scroll_position);
...
map->draw();
```

But also:
```c++
// from a file
map = new TiledMap("tilemap.blah", sprites);
// remove COPY_TILES/COPY_TRANSFORMS flags (read-only, less ram)
map = new TiledMap(asset_tilemap, sprites, 0);
// old TileMap behaviour, more restrictions, scanline transforms supported (+ anything other than a translate in general)
map = new TiledMap(asset_tilemap, sprites, COPY_TILES | COPY_TRANSFORMS | LAYER_TRANSFORMS);
```

For the affected examples:
- flight is now significantly faster (22 -> 12 ms)
- platformer is a little faster (38 -> 36ms)
- raycaster and tilemap-test are about the same

Anything that used `Map` probably also has a lower memory usage during init due to less copying `vectors` around.

Has a circular dependency on the [examples branch that ports everything](https://github.com/32blit/32blit-examples/tree/one-map-to-tile-them-all). (Deleted the `Map` code, can't find any uses outside of the three examples using it)


Gadgetoid:
Niiice! I should drag my brain kicking and screaming back into the world of the productive and try to port Rocks & Diamonds.

PR #863, opened 2024-10-08
sdl: make include directories PUBLIC in fallback path

Daft-Freak:
Because the library link is PUBLIC, when we have a new SDL2 with the targets defined, the include dirs also end up public. Change this in the fallback path for consistency.

This change is a result of me needing to call an SDL function in something as a workaround, which built fine locally... but blew up the actions build.


Daft-Freak:
Oh, since I updated the actions to 22.04, the fallback code isn't getting tested. Hmm.

PR #866, opened 2024-10-18
Remove JPEG API

### LINKED ISSUE #860: Remove decode_jpeg_* APIs?
I've been thinking about this a bit:
- As far as I can tell, `decode_jpeg_buffer` has two users (the example and my video player demo), `decode_jpeg_file` has none
- They're a big chunk of code in the STM32 firmware (~6k text, ~5k bss)
- Not even implemented in `32blit-pico`
- sdl2-image keeps causing dependency issues
- It would probably be more useful to have something that wraps the same decoder everywhere on the user side of the API :thinking:

Daft-Freak:
Listed some reasons for this in #860, but also:
- lots of deleting
- hacked together a [JPEGDEC wrapper](https://github.com/Daft-Freak/32blit-imgdec) yesterday and it isn't that much slower (11 -> 12 ms for the image in the example, and that includes having to do a R/B swap)

Fixes #860


Daft-Freak:
If I patch JPEGDEC a bit I can get it _faster_.

Gadgetoid:
An out-of-SDK JPEGDEC port seems like a reasonable alternative if that's what you're hinting at?

PR #869, opened 2024-10-24
Don't force 4-byte alignment on api structs

Daft-Freak:
There aren't any 64-bit members in these structs so this should have no affect on 32-bit platforms and the packing is causing problems with misaligned pointers on (mac) arm64.

Should fix this error:
`ld: building fixups: pointer not aligned at __ZL14blit_api_const+0x4 from 32blit-sdl/libBlitHalSDL.a[10](System.cpp.o)`

### DIFF
--- a/32blit/engine/api_private.hpp
+++ b/32blit/engine/api_private.hpp
@@ -149,7 +149,7 @@ namespace blit {
   };
 
   struct APIData {
-    uint16_t pad[2];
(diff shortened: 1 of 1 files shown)


Daft-Freak:
Doesn't break the firmware and I can't reproduce the error anywhere other than the original action that failed so... only one way to find out!

PR #873, opened 2024-11-04
pico: support switch pro controller

Daft-Freak:
Finishing off some old patches. Unsure about this though as it's a big pile of code, but it's one of the big gamepads...

### DIFF
--- a/32blit-pico/input/usb_hid.cpp
+++ b/32blit-pico/input/usb_hid.cpp
@@ -28,6 +28,7 @@ struct GamepadMapping {
 #define NO 0xFF
 
 static const GamepadMapping gamepad_mappings[]{
+  {0x057E2009,  3,  2,  1,  0, 17, 16, 19, 18,  8, 12, 11}, // Switch Pro Controller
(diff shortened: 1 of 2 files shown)


Gadgetoid:
I'd be inclined to believe that more PS5 controllers have sold, but 90% of them don't work anymore 😆 

But yeah, big is an understatement- wouldn't surprise me if this, and compatibles, weren't *most* of the market. Does this work with Pro controller-emulating third party controllers to? 8bitdo have a whole bunch among others.

Daft-Freak:
It's a good question, the third party one I had worked more like a regular HID controller (third one in the list).

Don't have any playstation controllers other than some generic "PS3" ones, but they seem to have the same "try to autodetect host and confuse tinyusb" issues as the retroflag ones... (Though i think the controller example in the tinyusb repo is for DS4?)

ali1234:
Most clone pro controllers don't even work with the Linux kernel driver, including the 8bitdo ones, in either USB or bluetooth mode.

PR #876, opened 2024-11-09
sdl: fix exiting while waiting for a render to be presented

Daft-Freak:
Did not expect to be debugging threading issues today. Fixes a 500ms hang + occacional crash at exit.


Daft-Freak:
Real change: https://github.com/32blit/32blit-sdk/pull/876/files?w=1

(I blame .editorconfig)

PR #877, opened 2024-11-26
Pico loader

Daft-Freak:
I finally untangled this!

Building with `BLIT_EXECUTABLE_PICO_BLIT=1` creates .blit files instead of .uf2s, these are _mostly_ portable between devices. (The framebuffer size is still set a compile time, so if you have the board set to PicoSystem (the default) the resulting .blit won't run on a device with a > 240x240 display. PicoVision is unaffected by this as its framebuffer is in the PSRAMs)

Installing multiple blits requires manually building them for different offsets (`-DBLIT_EXECUTABLE_PICO_BLIT_OFFSET_KB=x`), the launcher is built at a lower offset be default. On RP2350-based devices address translation is used to support running blits from any offset. Not much hardware to try that on though...

Some of the CDC API used by the tools is supported and can be used to flash blits onto a PicoSystem. This requires manually specifying the port as the tools don't recognise the VID/PID. (Which should probably be changed to something other than 0xCAFE first).

There's currently no menu but the "hold HOME to exit" feature is there. (If the device _has_ a HOME button... Sorry PicoSystem users.)

Still a lot of missing bits and duplicated logic with the -stm32 port, I have some plans to clean that up a bit. Need to see what it looks like complete first though...

(May edit if I remember more details... Been working on this for > 1.5 years...)

EDIT:
- flashing has no progress indicator
- CDC flashing doesn't deduplicate
- out-of-tree blit builds still build the loader
- CDC API only has `PROG`, `SAVE`, `__LS`, `LNCH` and `ERSE` (no `INFO`, `_RST`, `SWIT`)

### DIFF
--- a/.github/workflows/cmake.yml
+++ b/.github/workflows/cmake.yml
@@ -48,7 +48,7 @@ jobs:
             cmake-args: -D32BLIT_DIR=$GITHUB_WORKSPACE -DPICO_SDK_PATH=$GITHUB_WORKSPACE/pico-sdk -DPICO_BOARD=pimoroni_picosystem -DBLIT_EXECUTABLE_PICO_BLIT=1 -DCMAKE_C_COMPILER_LAUNCHER=ccache -DCMAKE_CXX_COMPILER_LAUNCHER=ccache
             apt-packages: ccache gcc-arm-none-eabi libnewlib-arm-none-eabi libstdc++-arm-none-eabi-newlib python3-setuptools
 
-          - os: ubuntu-20.04
+          - os: ubuntu-22.04
             pico-sdk: true
             name: PicoVision
             cache-key: picovision


Daft-Freak:
> 😱

A lot of that is going to be "distraction time" (or was this the distraction?)... But yeah. https://fosstodon.org/@Daft_Freak/109984773682008272

Daft-Freak:
I wonder if we can get a USB PID for "32blit SDK on RP2xxx" :thinking: (Considering that the exposed device is always the same... and the pythons have one).

I'm getting a little tired of doing `32blit flash --port /dev/serial/by-id/usb-TinyUSB_Device_6C89D8A6DE0598AF-if00 list`

ali1234:
See https://github.com/raspberrypi/usb-pid

They recommend using iManufacturer and iProduct (which you can set to whatever you want) for devices that have a CDC serial interface.

Daft-Freak:
Mmm, but this is a composite CDC+MSC device. We're currently defaulting to the very placeholder 0xcafe from the tinyusb examples.

Also it would seem that using the manufacturer/product strings doesn't work on windows. (I only get a manufacturer of "Microsoft" and no product from pyserial)

Gadgetoid:
> I wonder if we can get a USB PID for "32blit SDK on RP2xxx"

I can certainly ask RPi for an allocation!

PR #878, opened 2024-11-26
Specify a max version in cmake_minimum_required to remove some warnings

Daft-Freak:
I'm getting a lot of `Compatibility with CMake < 3.10 will be removed from a future version of CMake.` warnings whenever I configure the examples. Specifying a maximum compatible version makes those go away. (CMake 3.27 removes `FindPythonInterp`, the replacement for that is `FindPython/FindPython3` but those require 3.12).

I also needed to fix how we compile asm files in 32blit-stm32 to be compatible with CMake 3.20, where setting the source `LANGUAGE` to CXX is interpreted as "compile as C++" not "use C++ compiler".

(We should maybe consider bumping the minimum version to clean some stuff up... I'm not sure how far back we're trying to support, but even Ubuntu 20.04 has 3.16 and Debian oldstable has 3.18)

### DIFF
--- a/32blit-stm32/CMakeLists.txt
+++ b/32blit-stm32/CMakeLists.txt
@@ -1,3 +1,5 @@
+enable_language(ASM)
+
 add_subdirectory(../32blit-shared ../32blit-shared)
 
 add_library(BlitHalSTM32 OBJECT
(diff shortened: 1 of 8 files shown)


Daft-Freak:
Another point for "what CMake version should we require" is that the pico SDK requires 3.13. (vs our 3.9)

Daft-Freak:
Gonna merge this and fixup the examples so I can be free of the wall of warnings.

Still leaves the question of how old a CMake version should we support... (and how much version checking I can remove)

PR #889, opened 2025-01-31
Bump required CMake version and clean up a little

Daft-Freak:
Bit of a continuation of #878, the first patch makes it so we aren't using a "removed" feature.


Gadgetoid:
On the off chance it crops up here: I had a heck of a time recently on macOS trying to get `find_package` to find the Python in an activated virtualenv. It's very stubborn about picking the latest version, or picking fully qualified binaries rather than just "python."

I had to use `set(Python_FIND_STRATEGY "LOCATION")`.

Issue #681, opened 2021-06-06
dfu-util: Invalid DFU suffix signature


ali1234:
When flashing the latest firmware on linux I get this warning message:

```
dfu-util: Invalid DFU suffix signature
dfu-util: A valid DFU suffix will be required in a future dfu-util release!!!
```

I am not sure what this means but we should look into it to avoid it breaking at some point in the future.

I suspect it means there is some invalid header on the firmware.dfu.

Gadgetoid:
The suffix is this portion here: https://github.com/32blit/32blit-sdk/blob/dd9d68df7246eb220abc71168d7e5ceee4cf3de6/tools/dfu#L117-L122

Which, according to this datasheet: http://rc.fdr.hu/UM0391.pdf

● The bcdDevice field gives the firmware version contained in the file, or 0xFFFF if ignored.
● The idProduct and idVendor field give the Product ID and Vendor ID respectively of the device that the file is intended for, or 0xFFFF if the field is ignored.
● The bcdDFU field, fixed to 0x011A, gives the DFU specification number. This value differs from that specified in standard DFU rev1.1.
● The ucSignature field contains a fixed string of three unsigned characters (44h, 46h, 55h). In the file they appear in reverse order, allowing valid DFU files to be recognized.
● The bLength field, currently fixed to 16, gives the length of the DFU Suffix itself in bytes.
● The dwCRC (Cyclic Redundancy Check) field is the CRC calculated over the whole file except for the dwCRC data itself

And to meet the conditions for the CRC:

https://github.com/32blit/32blit-sdk/blob/dd9d68df7246eb220abc71168d7e5ceee4cf3de6/tools/dfu#L125-L130

This all *looks* right, insofar as there isn't much to get wrong.

Maybe - contrary to the comments at the top of the DFU tool - the suffix should be included in the file length: https://github.com/32blit/32blit-sdk/blob/dd9d68df7246eb220abc71168d7e5ceee4cf3de6/tools/dfu#L55

Gadgetoid:
P.S. What's your DFU Util version and flash command?

On Ubuntu I see a different flavour of wrong:

```
dfu-util -D firmware.dfu -d 0x483:0xdf11 -a 0
dfu-util 0.9

Copyright 2005-2009 Weston Schmidt, Harald Welte and OpenMoko Inc.
Copyright 2010-2016 Tormod Volden and Stefan Schmidt
This program is Free Software and has ABSOLUTELY NO WARRANTY
[...]

PR #892, opened 2025-02-05
docs: update dfu-util command

Daft-Freak:
To avoid #681. Specifying the vid:pid requires https://github.com/32blit/32blit-tools/pull/106 to not hit the other error in that issue.


Daft-Freak:
Uh, GitHib messed up merging this... the commit is there but it didn't close the PR or run the actions...

PR #897, opened 2025-04-12
pico: Presto + QwST Pad

Daft-Freak:
Involves adding a new DPI driver to replace scanvideo, also allowing the vgaboard to have `hires` on a Pico 2.

### DIFF
--- a/32blit-pico/CMakeLists.txt
+++ b/32blit-pico/CMakeLists.txt
@@ -43,7 +43,7 @@ target_sources(BlitHalPico INTERFACE
 )
 
 target_link_libraries(BlitHalPico INTERFACE
-    hardware_dma hardware_pio hardware_pwm hardware_spi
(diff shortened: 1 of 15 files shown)


Daft-Freak:
Because of there being no board header in the SDK, and there not being a generic RP2350B config(?), building this is a little weird... `-DPICO_BOARD=pimoroni_pico_plus2_w_rp2350 -DPICO_ADDON=pimoroni_presto`

Daft-Freak:
Bonus screen mode from repeat being generic: `set_screen_mode(ScreenMode::hires, {160, 160});`

Gadgetoid:
> Because of there being no board header in the SDK

We should probably fix that! Though keeping them downstream is very convenient 😆  (for me)

Daft-Freak:
We do have a custom board header dir set up here, I just didn't look very hard to see if there was one anywhere... :thinking:

PR #912, opened 2026-03-12
pico: additional warning fix in blit_launch

Daft-Freak:
Picked the wrong revision of a commit...


Daft-Freak:
... and another fix because I didn't test the "blit supports launch path, but there isn't one" case.

Daft-Freak:
Now that I've rebased I'm sure that there aren't any other little bits that got missed...

PR #914, opened 2026-07-25
Fix Windows SDL2 + CMake 4 brokenness

Daft-Freak:
```
Compatibility with CMake < 3.5 has been removed from CMake.
```
Strikes again...

Tried to keep this to the minimum updating to fix things (and not break other things, which updating SDL2_image on macOS would do...)

(I was trying to test the upload-artifact action [finally supporting not zipping a file](https://github.com/actions/upload-artifact#upload-an-individual-file-unzipped), but got windows breakage instead...)

### DIFF
--- a/.github/workflows/Emscripten.yml
+++ b/.github/workflows/Emscripten.yml
@@ -19,10 +19,10 @@ jobs:
     runs-on: ubuntu-22.04
 
     steps:
-    - uses: actions/checkout@v4
+    - uses: actions/checkout@v7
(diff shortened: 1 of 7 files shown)


Gadgetoid:
ohno

but also omg non-zipped artifacts!!!!!!

Daft-Freak:
It even works: https://github.com/Daft-Freak/DaftBoy32/actions/runs/30165928229#artifacts

